PR #192, opened 2015-02-23
Allow "transfer-source" to be specified so slave zones within the views ...

bradleykite:
...can match the correct view on the master.


jcameron:
Thanks for this patch!

PR #199, opened 2015-04-01
Add httpOnly to cookies on login page. Solves #4565.

tandrup:
Add httpOnly to pam_login and session_login to avoid security scans reporting false positives on cookies without httpOnly.

Should solve http://sourceforge.net/p/webadmin/bugs/4565/


jcameron:
Thanks!

PR #216, opened 2015-05-30
Important addition to make icons work with Authentic Theme

qooob:
Jamie, ok! It's very complicated to make things work in all possible situations. At the moment it doesn't work as in 90% of cases the certain rules are used. Those are that container for `&generate_icon` should be `<div class="row icons-row></div>`. In case it's done in funky way like in this module, I add extra class `inline-row` to make things work. For example, latest update 13.03 does great with style, that is used in Webmin Servers Index module. You can test it. Try using right mouse clicks for selects.

I strongly recommend, to try to recall all funky ways of using `&generate_icon` like in this module and add  `<div class="row icons-row></div>` wrapper.

It will not affect neither your nor Joe's themes.

Thanks!


jcameron:
Thanks! I will include this in the next Webmin release..

I have also checked in a similar fix for the proftpd module.

PR #224, opened 2015-07-02
Update sarg-lib.pl

claudius23:
execute sarg with configured sarg.conf file


jcameron:
Thanks for the patch!

PR #229, opened 2015-07-08
add child-src

kkprince1998:
frame-src is now depreciated, added child-src for compatibility for new browsers.


jcameron:
Thanks!

PR #230, opened 2015-07-08
XSS and cloudflare

KaidenP:
I use cloudflare for my server. Cloudflare is a rather advanced CDN that greatly improves the performance and bandwidth usage of my server, as well as speeding up page load times for the user. Also I get a free valid SSL certificate with them, so thats a bonus.

Anyhow, one of the many performance-optimizing scripts cloudflare uses is rocketloader, which modifies all other script tags so the browser just ignores them. Then the rocketloader downloads and/or executes them asynchronously.

The XSS stops all of cloudflare's script from executing, thus stopping rocketloader from running, and then all other scripts will not execute, which breaks the ui. The ui still looks good, just the menu at the side doesn't do anything when you click the submenus, and probably other things don't work too, but I cannot easily test them with the menu not working. but after making a small change (the one in this commit) it is fixed.

### DIFF
--- a/web-lib-funcs.pl
+++ b/web-lib-funcs.pl
@@ -766,7 +766,7 @@ if (!$gconfig{'no_frame_options'}) {
 	print "X-Frame-Options: SAMEORIGIN\n";
 	}
 if (!$gconfig{'no_content_security_policy'}) {
-	print "Content-Security-Policy: script-src 'self' 'unsafe-inline'; frame-src 'self'; child-src 'self'\n";
+	print "Content-Security-Policy: script-src 'self' 'unsafe-inline' 'unsafe-eval' https://ajax.cloudflare.com/; frame-src 'self'; child-src 'self'\n";
 	}
 if (defined($_[0])) {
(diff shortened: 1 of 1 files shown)


jcameron:
I don't think we can include a change that is specific to cloudflare. Have you instead considered making this configurable (by including $config{'something'} in the header), or disabling the content security checks entirely on your system? (By adding no_content_security_policy=1 to /etc/webmin/config)

kkprince1998:
> (By adding no_content_security_policy=1 to /etc/webmin/config)

 Awesome! I seen that it looked like it could be configured to be disabled, but I had no clue how. Google only brought me to the webmin config module. In the end, I was just manually adding it. I think someone should create more documentation on this, i was looking everywhere. Of course there may be docs i do not know of.

jcameron:
It's a hidden option, as it very rarely makes sense to disable this, and opens your system up to some nasty clickjacking attacks.

kkprince1998:
But if it runs under https, shouldn't it be fine?

jcameron:
No - https is no protection against clickjacking.

PR #241, opened 2015-07-30
`mode 1` endlessly better looking, right?

qooob:


jcameron:
Maybe .. if you have only a few commands, it makes sense to show them right on the first page.

qooob:
To show what on the first page? List of commands?

qooob:
I, by the way, added (in v. 14.01) automatic syntax highlight detection, now editing files in _Webmin_ with `Custom Commands` is absolute pleasure! Please try! ;)

jcameron:
I mean the command buttons and their inputs.

qooob:
Jamie, I would like to improve where I can and this is why I continue this discussion, I didn't understand to what you are trying to say at all. Makes no sense to me. What do you mean by `command buttons and their inputs.`? Sorry.

jcameron:
What I mean is the command parameters, like the "User to check" box in this screenshot ..
screen shot 2015-08-01 at 3 13 16 pm

qooob:
Alright, Jamie! I will take a close look!

PR #247, opened 2015-08-26
add module description in header

claudius23:
add module description to every page to make cloned modules distinguishable

### DIFF
--- a/sarg/edit_log.cgi
+++ b/sarg/edit_log.cgi
@@ -4,7 +4,7 @@
 require './sarg-lib.pl';
 
 $conf = &get_config();
-&ui_print_header(undef, $text{'log_title'}, "");
+&ui_print_header($text{'log_title'}, $module_info{'desc'}, "");
 print &ui_form_start("save_log.cgi", "post");
 print &ui_table_start($text{'log_header'}, "width=100%", 4);
 $config_prefix = "log_";
--- a/sarg/edit_report.cgi
+++ b/sarg/edit_report.cgi
@@ -4,7 +4,7 @@
 require './sarg-lib.pl';
 
 $conf = &get_config();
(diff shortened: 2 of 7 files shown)


jcameron:
This would make the UI quite different to other Webmin modules - perhaps instead this change should only happen in index.cgi ?

claudius23:
Maybe swap positions of $text{'log_title'} and $module_info{'desc'} like:
&ui_print_header( $module_info{'desc'}, $text{'log_title'}, ""); 

We are using 5 cloned sarg modules and without module description it gets really confusing, especially when using multiple tabs to compare configs or reports.
I don't think changing only index.cgi is sufficient.

Since  $module_info{'desc'} is set by "Webmin Configuration" -> "Module Titles", it would be really nice for consistency to have this description not only in the menu, but additional on every page.

qooob:
Hey, guys! From my point of view (if I understood you correctly) it would better to add module's description into `title` or `data-title` tag on where the module title is. Later I will be able to add tooltip for that purpose and when you hover on Title the Description will pop-up?

claudius23:
I hope some screenshots help
screenshot051
screenshot052
screenshot053
screenshot054

For me it is most important to see which module I'm currently looking at, since the title and url are not changing, I included the module description in page title.

qooob:
I see. You want this only for this module or global?

claudius23:
I would prefer global, but for the moment I only had time to implement this for sarg and squid. (Changes to squid are not pushed to github yet)

jcameron:
If you want to know which is the current module, this may be better implemented at the theme level. For example, in the new authentic theme the currently selected module is highlighted on the left side.

qooob:
+1

PR #248, opened 2015-08-26
execute sarg with all logfiles as parameter

claudius23:
Sarg needs a list of logfiles to generate reports.
Executing sarg multiple times generates only reports from last log file.


jcameron:
Thanks - but can you actually use @all in a string like that? Will it  be expanded with spaces between the filenames?

claudius23:
yes @all is expanded to a list of logfiles separated by spaces

jcameron:
Thanks!

PR #249, opened 2015-08-26
add missing sarg reports for topusers and downloads

claudius23:


jcameron:
Thanks for the patch!

PR #250, opened 2015-08-28
Add Edit word instead of OK

qooob:
To be the same over all modules it rather should be EDIT and not OK. Why not adding EDIT to the lang and use it? ;)


jcameron:
That would make sense - can you also include the lang file change in this patch?

qooob:
I think, I just should have edited `OK->Edit` at the very first time! :)

jcameron:
Thanks!

PR #251, opened 2015-08-28
Get rid of weird align=right

qooob:
It's better to remove align=right from this and other files, like `edit_lserv.cgi`,`edit_lserv.cgi`,`edit_aserv.cgi`,`edit_vserv.cgi` as it makes the module look different in a bad way compare to ALL others.


qooob:
I would guess `edit_lserv.cgi,edit_lserv.cgi,edit_aserv.cgi,edit_vserv.cgi` have the same. Please double check and then you could safely remove.. cause I didn't myself..sorry..

jcameron:
Sure, I'll fix those too.

qooob:
Thank you, Jamie!

PR #255, opened 2015-09-10
Fixed not calling switch_from_folder_user() for some types of maildir

gerco:
My spam was no longer being cleared up automatically and after some investigation and debugging it turned out that it doesn't work correctly when you use qmail maildirs. The first user on a domain will work just fine, any others will run into permission errors.

This turned out to be caused by not resetting the switch_to_folder_user() on certain types of mail directory by calling switch_from_folder_user() when needed. I replaced some return statements with an assignment to $rv. There may be other changed needed in the module, but with these changes it worked correctly for me.

The error messages looked like this:

    Error
    -----
    Failed to open /home/<user>/domains/<domain>/homes/<emailuser>/Maildir/.spam/cur : Permission denied
    -----

### DIFF
--- a/mailboxes/folders-lib.pl
+++ b/mailboxes/folders-lib.pl
@@ -1719,7 +1719,7 @@ if ($f->{'type'} == 0) {
 	}
 elsif ($f->{'type'} == 1) {
 	# A qmail maildir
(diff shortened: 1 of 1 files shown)


jcameron:
Thanks for this fix!

PR #271, opened 2015-12-29
LSI MegaRAID support

BindAr:
added support for LSI MegaRAID using MegaCLI

TODO: support multiple adapters, currently only adapter 0 is used


jcameron:
Thanks for the patch!

PR #289, opened 2016-03-09
Handle pipe commands

nielosz:
When using pipe commands in config like `@|"find /etc/bacula/conf.d -name '*.conf' -type f -exec echo @{} \;"`, this change executes the command and reads all files


jcameron:
Thanks for the patch!

PR #306, opened 2016-04-12
Add check for VirtIO to possible interface types

casparsmit:
A VirtIO interface is now shown as "Ethernet" because it is a "ethX"
The speed will never be shown as there is no "speed" for a VirtIO interface because it is a virtual interface. This way makes a distinction between Ethernet and VirtIO interfaces.


jcameron:
I'm not actually sure if this is a good idea, as other code in Webmin expects the interface type to be "Ethernet" for ethN interfaces.

casparsmit:
Ok, no problem. Skip this one then.

Kind regards,
Caspar Smit

2016-04-13 4:26 GMT+02:00 Jamie Cameron <notifications@github.com>:

PR #307, opened 2016-04-12
Add common spanning tree tunables for bridges

casparsmit:


jcameron:
Thanks for the patch!

PR #308, opened 2016-04-13
Anything is an object, right?

qooob:
Jamie, it's nicer to be specific.

Please separate tables and rows from being objects in case my changes will cause problems.

Ajaxifiyng MySQL module, just wanted to be best.


qooob:
However, it should be separated from table indexes to make it very clear.

PR #316, opened 2016-04-29
9000+ is way to little for `$pagelimit`

qooob:


qooob:
Also added `-init_datatables()` for Authentic.

PR #317, opened 2016-04-29
get_paths(); doesn't return proper results

qooob:
For wheel users (non-root but admins) `get_paths();` has a bug and doesn't work correctly.

It causes some other issues in user mode. Probably, if you on the `inthebox` login as `blog` user and go to File Manager, click copy - Perl Error happens. (under all themes)

### DIFF
--- a/filemin/bookmark.cgi
+++ b/filemin/bookmark.cgi
@@ -5,7 +5,7 @@ use lib './lib';
 
 &ReadParse();
 
-get_paths();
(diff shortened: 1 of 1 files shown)


qooob:
Could you also please not run for Authentic `init_datatables()` - it's not needed;

```Perl
if ($current_theme eq 'bootstrap') {
        init_datatables();
}
```

PR #320, opened 2016-05-04
Support for extracting: .7z, .bz2, .xz, .gz, .rar

qooob:
Commands are checked before they run so no need for last else part which outputted an error message. The rest of the code I will add to -lib file.

P.S. Jamie, I tried removing `use File::MimeInfo;` in this file as on current branch and it returned an error!? Please double check.

P.S.S. Just noticed - `&redirect("index.cgi?path=$path");` must be put to the very end only once.

### DIFF
--- a/filemin/extract.cgi
+++ b/filemin/extract.cgi
@@ -9,16 +9,29 @@ get_paths();
 
 $archive_type = mimetype($cwd.'/'.$in{'file'});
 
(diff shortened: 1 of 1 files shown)


jcameron:
It looks like the Filemin module was including a copy of File::MimeInfo  - but not being called anywhere.

PR #333, opened 2016-05-08
To compliy with upcomming Authentic release 18.00

qooob:
Respecting language forms is nice.


jcameron:
I assume this won't break systems running the older theme version?

qooob:
No, it will just print same stuff.

qooob:
Double checked - all is fine.

qooob:
Ahh. Older Authentic Themes... Hmm.. Let me think..

No, it will just print nice on the first load and then print old/incorrect stuff after first XHR.

qooob:
Actually, it will not print old stuff. It will print properly right away..

PR #337, opened 2016-05-10
Folder is more Windows like, Directory is Linux.

qooob:
To comply with Webmin and upcomming Authentic 18.00+ release.


jcameron:
Please make this change to the `lang/en` file instead - the `en.UTF-8` file is auto-generated from it.

PR #352, opened 2016-05-18
It has so many beautiful modules, calling it `Others` not very cool. …

qooob:
…I think `Tools` suites perfectly.

### DIFF
--- a/lang/en
+++ b/lang/en
@@ -23,7 +23,7 @@ category_net=Networking
 category_syslet=Syslets
 category_cluster=Cluster
 category_info=Information
-category_=Others
+category_=Tools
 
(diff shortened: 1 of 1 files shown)


jcameron:
While I agree that "Others" isn't the best name, so many users are used to it now that renaming it would be confusing.

qooob:
Ehh, ok. I understand. Let's do it on Webmin 2.0.

singhdd93:
I think it should be named Other/Tools in next release and then shifted to Tools in a major update

qooob:
+1

qooob:
I still think it should be called tools.

Confusion will go away after using it once.

PR #357, opened 2016-05-25
That seems incorrect for non numerical values

qooob:
In case it's used you can't use other values there, like 'true/false'. I think it should be removed, right?


jcameron:
Actually, the current code is intentional - it allows use of undef or "" for the No option.

qooob:
Then you can't use strings? (Are you on the plane? :) ) If I comment it on the theme's ui function will it work correctly? If not how to make it universal?

jcameron:
That's right, strings aren't supported. If you want more fine-grained control, the ui_radio function is a better choice - ui_yesno_radio is a convenience function for boolean options.

qooob:
Ok.

PR #362, opened 2016-05-29
strict and warnings compliant. Tighten up scope, fix a few undefined

swelljoe:
variable issues.

There are a few bits I'm unsure of; they're marked with XXX and/or FIXME. I tested the BIND module pretty thoroughly, and everything that usually works was working; I couldn't immediately get DNSSEC to work, but it looks like that code is a bit spotty, anyway. I've filed tickets about a couple of spots where I think these changes revealed bugs in the original code (things that I don't see how they could possibly work, as-is).

Please do check over this before merging, but I think it's looking pretty good. This is also Perl Critic friendly (I have a dream of being able to run Perl Critic across all of Webmin someday), except for the "return undef;" bits which are all through this...I wasn't sure what you'd want to return other than undef, and if that would impact other parts of the code, so I didn't touch those.

Let me know if you see anything you don't like, or would rather see done another way. I'm gonna try to work through a module or two each week until it's all done.

### DIFF
--- a/bind8/acl_security.pl
+++ b/bind8/acl_security.pl
@@ -1,11 +1,15 @@
+use strict;
+use warnings;
 
 require 'bind8-lib.pl';
+# Globals from bind8-lib.pl
+our (%config, %text, %in);
 
 # acl_security_form(&options)
 # Output HTML for editing security options for the bind8 module
 sub acl_security_form
 {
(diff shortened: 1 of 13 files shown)


swelljoe:
Ok, I think I merged all the things. This can probably be merged into webmin/webmin now.

jcameron:
Thanks for the fixes!

PR #383, opened 2016-06-03
Finish up strict/warn for bind8 module

swelljoe:
If you have automated tests for this module, it'd be worth running them. There's a lot of surface area I don't know how to exercise from the UI, though I've tried. (I think it has fewer bugs than it had before, due to finding a lot of undefined variables, even if I introduced a new one or two or three.)


jcameron:
I have tests for Virtualmin (which mainly exercise the module API) that I will run

PR #391, opened 2016-06-18
at module strict/warn conversion

swelljoe:


at/delete_jobs.cgi
```diff
@@ -1,16 +1,19 @@
...
+my @d = split(/\0/, $in{'d'});
 @d || &error($text{'deletes_enone'});
 
 # Delete each one
-@jobs = &list_atjobs();
-foreach $d (@d) {
-	($job) = grep { $_->{'id'} eq $d } @jobs;
+my @jobs = &list_atjobs();
+foreach my $d (@d) {
+	my ($job) = grep { $_->{'id'} eq $d } @jobs;
```
qooob: What is the point of `()` in this case in `my ($job)`, wouldn't just `my $job` be enough?

jcameron: No, I think that would cause $job to be assigned the number of matches (typically 1 or 0)

swelljoe: That's one of the weird Perl quirks that probably shouldn't be used much, as it is so subtle and not particularly intuitive. Especially when there are so many places where list context is used for other purposes...like `my ($one, $two, @therest) = @_`

Ilia, in short, the parentheses force list context (one or more items) on the receiving variable ($job), which would ordinarily have a scalar (one item) context. Perl functions that operate on lists behave differently when the receiver is scalar vs. list; in most cases, as Jamie notes, it gives the number of items in the list. In other cases, it'll give the first or the last item. In some cases this isn't what you want. So, you can tell Perl, "I want the list", which will in this case actually be the whole item it found.

Webmin has some really hairy nested data structures (from a time before Perl had good class support), so it has a lot of interesting uses of context like this. It can be tricky to figure out what's going on. It twists up my brain all the time. (Jamie has a bigger brain than most folks, and can hold all this stuff in his head in a way that I rarely can.)

jcameron: BTW, if there's a better way of doing this (extracting one element from a list), I'd be happy to switch to it.

qooob: :) This is tricky with parenthesis really. Thanks, Joe for explaining.

Jamie, I don't know any more concise way of doing it. All other would be using _foreach_ and probably slower.

PR #407, opened 2016-07-26
Clone rules are added after the original Rule

maanas:
In the existing state the clone rules are added in the bottom of the firewall wall rule list. Generally the clone rules are for similar interface, port, etc and most of the time after adding they need to be arranged. The change enable the clone rule to be placed after the original rule. This will allow a better comparison and arrangement of the firewall rules.


jcameron:
Thanks!

PR #415, opened 2016-08-19
Copy jobs do not have destination tape

ymulleneers:
copy jobs do not have any destination tape (here called volume) (=> ? on the 7th or latest field)
example (spaces removed) : Full Backup 11 19-Aug-16 17:50 ykfdc1-Copyjob


jcameron:
Thanks!

PR #432, opened 2016-09-30
firewall6 modules changes

cornec81:
updated firewall6 module. 
- some bugfixes.
- added pager - useful when there are many firewall rules.


jcameron:
Thanks for the patch! You could include the pager feature in the firewall/ module as well.

PR #437, opened 2016-10-09
increase specificity of regex to avoid problems with new output from certbot/letsencrypt client

deserted:
Some changes in the output of certbot/letsencrypt mean that the general regex is matching lines like `/etc/letsencrypt/keys/0037_key-certbot.pem` - this is resulting in the directory for finding cert.pem, fullchain.pem and privkey.pem is incorrect - they are looking up in `/etc/letsencrypt/keys/` instead of `/etc/letsencrypt/live/{domain}/`.

being this only applies when `$letsencrypt_cmd` has been set, it should be safe to push it to `/etc/letsencrypt/(live|archive)` as a hard folder match, to avoid this issue.

### DIFF
--- a/webmin/letsencrypt-lib.pl
+++ b/webmin/letsencrypt-lib.pl
@@ -81,7 +81,7 @@ if ($letsencrypt_cmd) {
 		return (0, "<pre>".&html_escape($out || "No output from $letsencrypt_cmd")."</pre>");
 		}
 	my ($full, $cert, $key, $chain);
-	if ($out =~ /(\/etc[a-zA-Z0-9\.\_\-\/\r\n ]*\.pem)/) {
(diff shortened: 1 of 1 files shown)


leotm:
Symbolic link fine?

leotm:
Symbolic link fine for the live folder?
Permission denied for hard links.

jcameron:
@leotm - are you saying that you're getting permission denied errors because Let's Encrypt is creating a hard-linked certificate file?

leotm:
Solved it, had my `/etc/nginx/sites-enabled/default` server block configured wrong, giving the failed output after running `letsencrypt` (so it wasn't creating the `/etc/letsencrypt/live/` folder).

Thought I needed to link `/etc/letsencrypt/keys/` to `/etc/letsencrypt/live/`, for some reason, my b.

ndstate:
Hello, I am running into the issue SSLCertificateFile: file '/etc/letsencrypt/live/.../fullchain.pem' does not exist or is empty when I run apachectl -S. Is there a fix? Thank you.

PR #446, opened 2016-10-22
Fix bind version comparison in bind8-lib

augustr:
I'm using bind 9.10 (>=9.4) and I don't get the "DNSSEC response validation enabled" option in webmin due to a bug in the version comparison of bind versions in bind8-lib.pl.


augustr:
The old version of the code treats the string "9.10" as the number 9.1, which is then regarded less than 9.4...

jcameron:
Thanks for the fix!

PR #462, opened 2016-11-28
Output optimisaztions. Part 2

qooob:
Templates must be printed here.


qooob:
No, 

I missed the fact that they contain `form` inside.

Moving `print_template("unauthenticated/templates/dialogs.html");` back...

PR #464, opened 2016-11-29
SeLinux/attributes listing and control - Security strengthening

qooob:
Jamie, here is the fix.

About `html_escape` for printing to user. Well, this is usually happens on error, when we use "`$file_name` - error message". Is this really necessary to do?

Example, would that be better:

`push @errors, "html_escape($file) - $text{'attr_label_error_proc'}: $?";`

than just:

`push @errors, "$file - $text{'attr_label_error_proc'}: $?";`


jcameron:
Yes, calling html_escape on the filename would also be a good idea.

qooob:
Sure, I will tomorrow morning.

PR #476, opened 2017-01-10
Enable editor for .XSL files

qooob:
https://github.com/qooob/authentic-theme/issues/629#issuecomment-271280127


qooob:
Jamie it needs to be added.

PR #488, opened 2017-02-01
Don't blank the password field on the login page

mnoorenberghe:
This interferes with password managers and the general consensus from the security community is that password managers improve security by encouraging stronger passwords and reducing password re-use.

There was no useful GIT blame for this line as it dates back to the initial import. I assume it's leftover from the old days and thinking has changed since then.


jcameron:
Thanks - I have no idea why I added that code.

PR #499, opened 2017-02-23
Enable/Disable for Winbind & File/Printer Share buttons

jjinno:
The following adds an administrative ability to:
- turn on or off the "winbind options" button
- if create/read/write ALL disabled for File sharing, then dont display the button
- if create/read/write ALL disabled for Printer sharing, then dont display the button

There is additional cleanup of some code inside of samba/index.cgi to improve readability, and minor fix to maintain more I18N consistency (see SWAT-title reference)

This DOES NOT fix the fact that "winbind" variables are used for the feature "Bind to Domain" and "conf_bind" variables are used for the feature "Winbind Options"... which is seemingly backward... but at least is consistent... so was left like that for now.


jcameron:
Thanks for the patch!

PR #502, opened 2017-03-01
enable/disable cluster-useradmin features & fix visual bugs

jjinno:
The idea was to allow an administrator to disable the user/group add/search mechanisms, thereby only showing the "refresh" and "sync" buttons... but I have expanded the config options to cover all 6 of these features, AND enable/disable for the link used to access edit_host.cgi. 

In the process, I discovered that none of the icon image paths were resolving, so fixed that. Additionally, noticed some table-formatting issues in the "edit hosts" page (fixed).

Finally, since there is one module-global option for choosing table versus icon mode, I ported the table-style view to edit_group & edit_user... this way, if table-view is desired, it is held consistent throughout the module.

Default values for everything should maintain existing behavior.


jcameron:
Thanks!

PR #519, opened 2017-03-18
Let a function to accept non numeric value

qooob:
In case the way it is now, there is no use in custom values for radio buttons. For example, it will never let it to set it `false` or `true` (as a string), thus always passing an integer to the `ui_radio()`.


jcameron:
Thanks! This looks reasonable..

PR #525, opened 2017-03-27
makemodulerpm.pl updates

swelljoe:
This does the following: 

  - Allows it to run without root privs without generating errors.
  - Colorizes usage output (falls back to no colors on terminals that don't support it (tested on Windows CMD, PuTTY, git bash, and of course Linux, but the compatibility chart for ANSI colors covers everything UNIX-y for the past 15 years at *least*).
  - Is strict and warnings happy.

Some other things I'd like to change, but not today:

  - No need to untaint input when it doesn't run as root, so that taint code can go away.
  - /tmp/makemodulerpm is racy, if I have multiple build jobs going at once on the same vm or container (I think it'll spin a container up for each build rather than sharing one, but not sure about that...still a racy thing...somebody might do it some day.)
  - Move it out to its own repo and package. No reason for it to be in Webmin proper, I don't think. It could get proper docs and other stuff on its own repository. Could be packaged as a standalone RPM/deb, making it easier to pull into a build environment as a dependency, etc. I can just grab it as a lone script for now.
  - There's a bunch of unnecessary stuff going on, like scripted deps checking in pre-install, that should just be a "Requires: webmin" or "Requires: usermin". But, it's harmless and does mean a non-RPM install could get an RPM installed package (which is asking for trouble, but oh well).

It needs more testing, but will get thoroughly tested over the next few days as I automate builds of modules.


swelljoe:
Oh, yeah, I should note your user needs to have an rpmbuild directory setup in $HOME. It no longer tries to use a system redhat dir in /usr/src, or whatever (that isn't a thing, anymore)

```bash
mkdir -p $HOME/rpmbuild/{RPMS,SRPMS,SOURCES,BUILD,SPECS}
mkdir $HOME/rpmbuild/RPMS/noarch
```

The script insists on a noarch directory existing in RPMS, but that's not actually necessary, I don't think. I'm pretty sure rpmbuild will create the arch directories when it needs them. I'll test it and remove that check, if that recollection turns out to be true.

PR #526, opened 2017-03-28
Excluding bind mounts properly

qooob:
Jamie, hi.

I finally had a time to take a look. The fix is simple and straight forward. It works perfectly on my production system. All "bad" mounts are excluded and not listed anymore.


jcameron:
Thanks for the fix!

PR #527, opened 2017-03-31
Added v flag for tar extraction

razor2k:
Added the v (verbose) flag when really extracting the backup tar.
If this flag is not set tar will not print the list of files, that are extracted. This leads to an empty output which later leads to an empty @$files when trying to split the output from the untar command.
This then leads to restore.cgi falsely reporting "0 files have been restored".
Tested on Ubuntu 14.04.5


jcameron:
Thanks for the patch!

PR #528, opened 2017-04-06
fix: memory leak  - remove circular reference in bind8 lib

Demayl:
Spotted when restoring from backup.
So memory pressure when restoring is significantly reduced.

### DIFF
--- a/bind8/bind8-lib.pl
+++ b/bind8/bind8-lib.pl
@@ -284,7 +284,7 @@ if ($str{'name'} eq 'inet') {
 				my $substr = &parse_struct(
 						$_[0], $_[1], \$i, $j++, $_[4]);
 				if ($substr) {
-					$substr->{'parent'} = \%str;
+					$substr->{'parent'} = {%str};
 					push(@{$str{'members'}->{$t}}, $substr);
 					}
 				}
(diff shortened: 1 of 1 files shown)


jcameron:
I'm not so sure about this patch - it would mean that navigating down the named.conf parse tree and back up it wouldn't return to the same object.

Demayl:
In worst case it must be rewritten. I tested it fast and it doesn't break anything, but something can count on this bug to work. This circular reference must be fixed, because leaked memory is huge ( around 12MB per domain restore ).

jcameron:
When you say "domain restore", are you talking about Virtualmin backups and restores?

Demayl:
Yes

jcameron:
Ok, what confuses me is that this patch doesn't seem related to any code that is invoked as part of the backup or restore process - if there is a memory leak, it should happen any time DNS records are listed or edited by Virtualmin.

Demayl:
Ok, you can try it too. It's a problem on long running operations ( like the restore ). Ops like list/edit are living short time.

jcameron:
I wasn't able to re-produce this - how large is the named.conf file on your system?

Demayl:
1663 records, 258K
To reproduce this you must restore 2+ domains with DNS opt enabled

jcameron:
So you have 1663 domains hosted by Virtualmin with DNS enabled? That's .... quite a lot.

But I'm still having trouble re-producing this issue, or seeing where the memory leak could be coming from. Even though the parsed named.conf structure has a loop,  this on it's own shouldn't cause a leak unless it is serialized - which doesn't happen anywhere in the backup / restore code.

swelljoe:
Closing this, as no follow up in 9 years. I guess if it's still an issue, re-open.

PR #530, opened 2017-04-14
href also needs to be escaped

qooob:
It's nessesary in case of a fancy directory name to prevent HTML breakage/injection


jcameron:
This is incorrect - for URL params, you need to use `urlize` instead of `html_escape`

qooob:
My bad, sorry.

PR #543, opened 2017-05-02
Make editor respect file's initial encoding: Viewer

qooob:
This patch will make sure that user, will be able to READ files with different encodings properly and also, will be able to SAVE them without changing the initial encoding.

This awesomeness requires dependencies. However, RHEL has it installed by default but not Debian. New Webmin release should make sure to require/pull dependent packages:

For RHEL: `perl-Encode-Detect`
For Debian: `libencode-detect-perl`

Example of how editor will display detected encoding and display encodning in the editor's header/tabs:
screenshot from 2017-05-02 12-32-37

<br>

-----

> P.S. We should no let user to set non-UTF8 locale anymore. All languages should be displayed unambiguously, like Google does, (Русский, svenska, Deutsche, English (United States) .. ).

-----

<br>

### DIFF
--- a/filemin/edit_file.cgi
+++ b/filemin/edit_file.cgi
@@ -5,39 +5,49 @@ require './filemin-lib.pl';
 
 get_paths();
 
-$data = &read_file_contents($cwd.'/'.$in{file});
+my $data = &read_file_contents( $cwd . '/' . $in{file} );
 
-&ui_print_header(undef, $text{'edit_file'}, "");
+use Encode::Detect::Detector;
+my $encoding_name = Encode::Detect::Detector::detect($data);
+if ( $encoding_name && lc($encoding_name) ne "utf-8" ) {
+    use Encode qw( encode decode );
+    $data = Encode::encode( 'utf-8', Encode::decode( $encoding_name, $data ) );
+}
+
+&ui_print_header( undef, $text{'edit_file'}, "" );
 $head = "<link rel='stylesheet' type='text/css' href='unauthenticated/css/style.css' />";
 
-if ($current_theme ne 'authentic-theme') {
-    $head.= "<script type='text/javascript' src='unauthenticated/jquery/jquery.min.js'></script>";
-    $head.= "<script type='text/javascript' src='unauthenticated/jquery/jquery-ui.min.js'></script>";
-    $head.= "<link rel='stylesheet' type='text/css' href='unauthenticated/jquery/jquery-ui.min.css' />";
+if ( $current_theme ne 'authentic-theme' ) {
+    $head .= "<script type='text/javascript' src='unauthenticated/jquery/jquery.min.js'></script>";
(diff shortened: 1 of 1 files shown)


qooob:
Ref: [virtualmin.com@52020#comment-775771](https://www.virtualmin.com/node/52020#comment-775771)

qooob:
https://github.com/qooob/authentic-theme/commit/83acce8f0a54a59305b5c52437870d923d348646

jcameron:
So it looks like this change adds a dependency on a new perl module, and does auto-detection of the edited file encoding. But Webmin has had several file editors (ie. in the Custom Commands module) for a while now and I didn't run into the issue of the encoding being changed before ... so I'm not sure why this special handling is needed?

qooob:
To reproduce the bug, use old release, (1.831 + with shipped theme). Go to `../webmin/lang/`, find non UTF-8 file, like `sv` and open it. You will see things broken if you have UTF-8 set as language in Webmin configuration.

In old themes, your /lang files will look broken, if you set non-unicode charset in settings and try to see the files.

The only and best thing is to use unicode and detect things correctly letting Perl convert data properly. It worked very stable on any file I tried.

The point is, to read right first, then to save file in the same encoding. Third option that I will add is a select that will let users select/change encoding for the edited file.

This is why folks need non-unicode. When we can view/save data without issues, all should be happy and I can't think of other reasons to keep old, non unicode stuff, can you?

jcameron:
I'll also check in a fix to deal with the case where this Perl module is missing.

chris001:
Hi @jcameron , is it possible you forgot to check in a fix to deal with the case where this Perl module is missing and install it or alert the admin if install fail, and this is why the [bug](https://github.com/webmin/usermin/issues/80) has struck in Usermin Mailbox message list?

jcameron:
Ah ... so the causes is that the authentic theme isn't handling the case where the `Encode::Detect::Detector` perl module is missing?

iliajie:
Oh, nice! Actually, on contrary! Having it installed breaks things!

chris001:
On this system (Deb 11), that perl module is missing, and the Mailbox List Subject emoji `UTF-8` is fail!

iliajie:
Chris, check this patch - https://github.com/webmin/authentic-theme/commit/5a97f6caebc163ef1ee57d7c4703881ab786e81c

chris001:
Ilia, This patch is success!  All emoji in Subject in Mailbox Message List now display correctly!
I correct my previous comment, in fact, the library is indeed installed, and bug was happening anyway (previous to this patch), is it possible when you called the library it was returning a value that was unexpected by the code?

iliajie:
It was double encoding it. The problem with `Encode::Detect::Detector` is that it requires pretty long string to be accurate at guessing.

When I further rework the mail for Usermin I will take a closer look at the current implementation of `Encode::Detect::Detector` on the theme side.

PR #551, opened 2017-05-07
filter out chains not managed by firewall and add warning if fail2ban detected

gnadelwartz:
I permanantly ban hosts and networks often found by fail2ban plus a huge list of networks loaded from different sources. the list contains nearly 5000 entrys:

```
 rrr ~/ #./fail2ban-stat.sh
So 7. Mai 21:05:47 CEST 2017 start fail2ban-stat.sh 22785
Fail2Ban stats: 265 attacks, 11 ignored matches and 165 banned hosts this Month
Currently banned: temporary 5 hosts, permanent 345 hosts and 4623 networks...

```
I wanted to filter out my BAN-forever chain from firewall because of 2 reassons:

1. its a quite long list to display in the browser
2. processing the list takes very long and I have to klick away 3-5 browser warnings about a long running script

So I implemented a regexfilter to skip chains in get_iptables_saved() and a config option filter_chain to enter the regex with a default of "forever" , its a simple oneliner :-)
`next if ($config{'filter_chain'} && s/$config{'filter_chain'}//);`

pls consider merging or implement a similar solution

### DIFF
--- a/firewall/config
+++ b/firewall/config
@@ -4,3 +4,4 @@ view_condition=1
 cluster_mode=1
 comment_mod=0
 force_init=0
+filter_chain=forever
--- a/firewall/config
+++ b/firewall/config
@@ -4,4 +4,3 @@ view_condition=1
 cluster_mode=1
 comment_mod=0
 force_init=0
-filter_chain=forever
--- a/firewall/config.info
+++ b/firewall/config.info
@@ -11,4 +11,4 @@ after_apply_cmd=Command to run after applying configuration,3,None
 line1=System configuration,11
 save_file=IPtables save file to edit,3,Use operating system or Webmin default
 direct=Directly edit firewall rules instead of save file?,1,1-Yes,0-No
-filter_chain=Regex to filter out chains not managed by firewall, e.g. fail2ban chains,0
+filter_chain=Comma sepeated list of regexes to filter out chains not managed by firewall,0
--- a/firewall/config.info
+++ b/firewall/config.info
@@ -11,3 +11,4 @@ after_apply_cmd=Command to run after applying configuration,3,None
 line1=System configuration,11
 save_file=IPtables save file to edit,3,Use operating system or Webmin default
 direct=Directly edit firewall rules instead of save file?,1,1-Yes,0-No
+filter_chain=Regex to filter out chains not managed by firewall, e.g. fail2ban chains,0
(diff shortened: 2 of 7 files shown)


gnadelwartz:
yes, it can be a bit dangerus to use the user input as is, in case someone type ".*" as regex ,every line is dicarded.

I tryed to fix this by palaceing "(-A|-j) [^ ]*" in front of the entered regex to fiaxate ist to chain but it did not work as expected.

may be we should use the user input as an list of comma seperated list of fixed strings containig the names of chains to ignore.

any ideas?

gnadelwartz:
OK, I have a solution now :-)

comma seperated list of regex matched against chain names only
should be saver than the oneliner.

If you think its OK I'll impement it.

```
open(FILE, "iptables-save 2>/dev/null |");

LINE:
while(<FILE>) {
        local $read_comment;
        s/\r|\n//g;
        # regex to filter out chains not manage by firewall, i.e. fail2ban
             foreach $filter (split(',', '.*-forever, ,,')) {
                 if (/^.?:(\S+)\s+.*/) {
                         next LINE if($1 =~ /^$filter$/);
                    }
                 # -A|-I chain ... -j chain
                 if (/^.?-(A|I)\s+(\S+).*\s+-j\s+(.*)/) {
                         next LINE if($2 =~ /^$filter$/);
                         next LINE if($3 =~ /^$filter$/);
                    }
                }

         print "$_\n"
}

```

jcameron:
Maybe it's better to do the filtering on `index.cgi` , so that Webmin still heads the rules but just doesn't show them?

gnadelwartz:
This was also my first guess to do, but I'm shure not only index.cgi read and handle chains, there may many scripts read tables and we may have to filter them everywhere.

So the question is: is it better to filter them out for all scripts in filewall-lib or only to exclude from displaying in index.cgi and let all other scripts now about them?

filtering in firewall-lib is much faster and has less memory usage. e.g. the time difference in my setup is 30s for read in and transfer html to client (message transfer data from .../webmin) and 2 Minutes to process them in the browser vs. read, transfer, displaying the filtered list takes less than 2 seconds after request.

gnadelwartz:
after some investigation to index.cgi and other scripts, I decided to implement filtering in get_iptables_saved().

dealing directly with tables and chains structures seems to error prone, at least for me...

jcameron:
Thanks for the patch!

gnadelwartz:
there is is bug in filter chains code, see #555, fix will be uploaded tomorrow

PR #554, opened 2017-05-09
add warning about fail2ban firewalll rules detected

gnadelwartz:
while thinking about to filter chains I investigated how to index.cgi handle tables and chains.
if fail2ban is in use admin should be careful about mangling rules by hand, so I implemented a warning for fail2ban smilar to shorewall and firewalld.

OOPS, created a new branch and was thinging only my new chages will be ditributed, sorry.
pls have a look on my changes to index.cgi and the lang files only!

### DIFF
--- a/firewall/config
+++ b/firewall/config
@@ -4,3 +4,4 @@ view_condition=1
 cluster_mode=1
 comment_mod=0
 force_init=0
+filter_chain=forever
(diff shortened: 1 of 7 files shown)


gnadelwartz:
grafik

gnadelwartz:
can be closed, I pulled the warning changes into my master, to avoid merging old filter method.

gnadelwartz:
there is a bug in my code wich I didn't found while testing because I was using "direct" mode.
A fix is already in my local installation and will be checked in tomorrow.

In short: works only with direct mode, when "using save to file" mode filtered chains will be silently disacarded.

Issue #555, opened 2017-05-09
bug in filter_chains code #551


gnadelwartz:
there is a bug in my code #551 wich I didn't found while testing because I was using "direct" mode.
A fix is already in my local installation and will be checked in tomorrow.

In short: works only with direct mode, when "using save to file" mode filtered chains will be silently disacarded.

jcameron:
OK, please send me the fix when you have it.

gnadelwartz:
OK, fix is in pull request #557

gnadelwartz:
UI changes should be also finished, pls review  #557 and #558

PR #557, opened 2017-05-10
hotfix filter chain #555

gnadelwartz:
fixes bug #555 in filter_chain code.


gnadelwartz:
should work now as expected

PR #561, opened 2017-05-14
tarnslation adjustements like for usermin

gnadelwartz:
If I'm bored again next week I may do the same as for usermin for webmin, but may be I'll wait until you fixed this mix of languages to UTF-8 everywhere :-)


jcameron:
Thanks!

PR #562, opened 2017-05-14
script for updating lang files from repository

gnadelwartz:
When translating usermin on with github editor, which is great for UTF-8, I cant review my changes in webmin/usermin UI. With autentic-theme it was not so worse, because after qooob accepted a commint you can update theme and review the UI.

But with webmin/usermin I had to make the changes also local or have to wait until next release :-( So I stole qooops script and modified it to download the latest LANG files from webmin/usermin repository or a forked repository, e.g. mine. It supports both webmin and usermin.

This helped me a lot to review changes I made directly in the UI and fix them if somethins was wrong, and there was much to fix :-),  without it I can't adjust so much translations for usermin today.

Maybe you are interested in it...


jcameron:
This is cool! However, another option would be to checkout the Webmin source entirely, and run from the checked-out dir rather than an RPM. We provide a `local-setup.sh` script to do this.

gnadelwartz:
regarding updateting from gitgub source, here I have some questions:

- in usermin you used linkfiles often. how does the link files work when downloading with git? are they resolved or created as links.?

- in usermin I missed filemin and because I'm not shure if other modules are also missing, I implemented the workaround of downloading webmin lang files first, then overwriting with usermin files. This may not needed  or sould be more selective.

- I was also thinking about complete chekcout but the description on the website how to install from tar.gz with setup.sh seems to "interactive" I'll have a look at local-setup,sh

gnadelwartz:
some of my questions i can answer myself now

- links are converted to softlinks wenn pulled from usermin repo, but they point all to ../webadmin instead of ..webmin here I need a better workaround for my script.

- missed filemin is bcause its also a link

these links in repo usermin are the reason why we cant checkout usermin an run setup.sh directly.
an other resason why I only update latest lang files is that cgi scripts may broken temporary in repo.

for the softlinks in usermin repo I have an idea how to solve better for updating lang files ...

gnadelwartz:
I setup usermin from scratch and updated langfiles from my repo ... it worked including dereferencing repo symlinks to get files from webmin dir :-)))

jcameron:
Yeah, the symlinks have some requirements about the install directory. Have a look at http://www.webmin.com/git.html for details.

gnadelwartz:
ok, i got the idea why symlinks points to webadmin ☺. but for the final rpm/tgz you replace them by the referenced files/dirs, correct?

gnadelwartz:
after some investigation on how the tarball is created, i got an idea how to update webmin to latest repo changes for installitions NOT in /usr/local/webadmin

gnadelwartz:
hi again,

I have a working script to update webmin/usermin from repo using you scripts makedist.pl and setup.sh after git clone to dereference symlinks and install only needed files, not whole rpeo.

one problem remains: alls files has the x bit set after git clone and also after setup.sh.
may be this is not a problem, but after installing from rpm only *.pl *.pm and *.cgi has the x bit set, which is more corect.

gnadelwartz:
in makedist.pl you use absolute path to /usr/local/webmin .../usermin to go there and make charset translations. This may also not a problem, but I'm not shure if something is missed if I checkout to a tempdir to do the update.

I know makedist.pl is there to create a taball for installation, not to prepare files fro an update, but it provides the some nice features like derefencing symlinks, copying only needed files into tarballs and does some additional processing.

this seems much better than a simple `cp -r -L` to copy everything from temp dir into /usr/libexec/webmin or /usr/libexec/usermin

jcameron:
Generally users won't need to run `makedist.pl` , as it's for packaging Webmin up for public distributuion.

gnadelwartz:
I know, but it makes my life easyer 😁 . If you suggest I shouldn't use it I can change the script.

jcameron:
I'd recommend not calling makedist.pl , as it might get updated in future to behave in a different way.

gnadelwartz:
for now I like to stay with it to not duplicate functionallity, but Im prepared if it changes, is this OK for you?

jcameron:
Sure, works for me..

PR #563, opened 2017-05-16
`en.UTF8` -> `en` for the upcoming 1.841 to satisfy compiler

qooob:


gnadelwartz:
hi,
may be I'm behind your knowledge, but do we have a compiler for webmin scripts/lang files?

qooob:
Jamie does it. It's about to change in the near future. For now, just edit non _utf8_ file to make sure it's added to the package properly.

qooob:
👍

PR #565, opened 2017-05-17
This one working for me and displaying info properly

qooob:
```BASH
root@debug-freebsd:~ # sysctl hw.model hw.ncpu
hw.model: Intel Core Processor (Skylake)
hw.ncpu: 8

```

However, it might not catch properly the information where output is like this:

`Intel(R) Xeon(R) CPU E5-2630 0 @ 2.30GHz, 24 cores`

### DIFF
--- a/proc/freebsd-lib.pl
+++ b/proc/freebsd-lib.pl
@@ -145,7 +145,7 @@ else {
 	return ( );
 	}
 $out = &backquote_command("sysctl hw.model hw.ncpu");
-if ($out =~ /hw.model:\s+(\S+)\s+(\S.*\S)\s+\@\s+(\S+)/) {
+if ($out =~ /hw.model:\s+(\S+)\s+(\S.*\S)\s+(\S+)/) {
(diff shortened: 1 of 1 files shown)


jcameron:
@iliaross - which FreeBSD version are you running there? All of mine have an @ in the CPU name.

qooob:
Yes, correct if it's Xeon processors. On my desktop in virtual machine I have:

`Intel Core Processor (Skylake)` and it doesn't work.

On the other hand, my served has:

`Intel(R) Xeon(R) CPU E5-2630 0 @ 2.30GHz, 24 cores` and it works fine.

jcameron:
Are those both the same FreeBSD version?

qooob:
No.  Only (Skylake). You can see what that outputs on the very top.

Your regex should include to catch a string without @.

At current state it's incomplete but mine on the commit incomplete as well. I just wanted to show how and why it works on my end and what's missing.

jcameron:
Thanks - I've checked in a new change that handles both cases.

PR #568, opened 2017-05-20
hofix detecion (random) iso-8859-x charset

gnadelwartz:
when editing a file containg mixed iso-8859-x charsets eg module.info :-), Encode::Detect::Detector::detect returns the first guessed iso-charset. My hotfix is to use iso-8859-1.

may be it is an option to convert over to `Encode::Guess` , its also guessing, but gives an error if multiple charsets are possible.

### DIFF
--- a/filemin/edit_file.cgi
+++ b/filemin/edit_file.cgi
@@ -13,6 +13,8 @@ my $encoding_name;
 eval "use Encode::Detect::Detector;";
 if (!$@) {
 	$encoding_name = Encode::Detect::Detector::detect($data);
+	# fix random iso-8869-* detection to be always iso-8859-1
+	if ( $encoding_name ~= /^iso-8859/ ) $encoding_name="iso-8859-1";
 	}
 if ( lc( get_charset() ) eq "utf-8" && ( $encoding_name && lc($encoding_name) ne "utf-8" ) ) {
     use Encode qw( encode decode );
--- a/filemin/edit_file.cgi
+++ b/filemin/edit_file.cgi
@@ -10,11 +10,18 @@ my $file = &simplify_path($cwd.'/'.$in{'file'});
 my $data = &read_file_contents($file);
 
 my $encoding_name;
+use Encode::Guess; 
 eval "use Encode::Detect::Detector;";
 if (!$@) {
 	$encoding_name = Encode::Detect::Detector::detect($data);
-	# fix random iso-8869-* detection to be always iso-8859-1
-	if ( $encoding_name ~= /^iso-8859/ ) $encoding_name="iso-8859-1";
+        if ( $encoding_name =~ /iso-8859/i ) {
+          # try a second guess
+  	  my $dec = guess_encoding($data, qw/iso-8859-1 iso-8859-2 iso-8859-3 iso-8859-3 iso-8859-4 iso-8859-5 iso-8859-6 iso-8859-7 iso-8859-8 iso-8859-9/);
+  	  if (!ref $dec) {
+            # can be more than one iso-encoding, fallback to iso-8859-1
+            $encoding_name = "iso-8859-1";
+  	    }
+          }
 	}
 if ( lc( get_charset() ) eq "utf-8" && ( $encoding_name && lc($encoding_name) ne "utf-8" ) ) {
     use Encode qw( encode decode );


qooob:
What if it's really `iso-8859-7` (Greek)? Then it will always fall to `iso-8859-1` incorrectly?

gnadelwartz:
as you mentioned already its always a guess onyl what charset is used and a hotfix :-).

detecting iso-8859-x is a mess, as decribed in Encoding::Guess doku.
http://search.cpan.org/~dankogai/Encode-2.89/lib/Encode/Guess.pm#CAVEATS

May be we should implement a setting for "preferend iso-8859 endoding".

qooob:
Kay, at the moment if encoding is not detected, the theme will let you pick encoding and once done so, it will fall back to your previous choice.

In case you have mixed set of characters in a file and it's detected incorrectly, just change encoding on top of the editor. It's that simple. 

I don't think that your fix is going to be right for all.

gnadelwartz:
it can be that simple, but it is not working, as I explained in my issue https://github.com/qooob/authentic-theme/issues/785

If i change encoding on top of the editor, an empty editor is diplayed.
If I do copy file contents to clipboard, then change the encoding as suggestet and paste in it looks correct in the editor, but after saving. But when saving with the changed encoding special charactes are broken :-(

An alternativ soltuion may be to reopen the editor with charset enforded to the one selected in editor.

qooob:
Then it means that this encoding doesn't support such chars?

Cab you please attach a file I will try it on my computer?

gnadelwartz:
I think it is because filremin does conversion to UTF-8 first, then edit and convert back on saving.
you can see the result here:

I alway change only the german lines.

1. change: stay with detected charset and insert german -> german ü is wrong
2, change: copy, change charset, paste -> german is OK, but all others are broken

https://github.com/gnadelwarty/usermin/commits/master

qooob:
You could use https://gist.github.com to share a file.

qooob:
Can you please do so? I don't see the value fail on the commit?

gnadelwartz:
If I paste this in gisthub it also goes vi cut and paste and may change charactes also, so I made a raw zip file for downloading an in gisthub https://gist.github.com/gnadelwartz/31b0ecf08a2e7ef8d06f9029db1b8213

[files.zip](https://github.com/webmin/webmin/files/1016149/files.zip)

qooob:
Kay, in case I open `module.info` all seems correct. What makes you want to change encoding to `iso-8859-1`? It will not work with _el_ and _gr_ languages.

qooob:
I get expected results. Just keep file in `iso-8859-7` or `cp1253`?

gnadelwartz:
I want to edit in inso-8559-1 to enter german characters correctly, If stay with iso-8859-7 and enter german chracters they are not correct after saving. Eg in webmin, because file is inpreteted as iso-8859-1 or more correct as CP1252.

IMHO the workflow in edit_file is:
`open file -> detect charset -> convert to UF-8 -> edit (enter eg. german ü) -> convert back from UTF-8 to charset (german ü is gone) -write file 
`
an option to open file with a specific charset sems nice to me or as mentioned if you select other charset in editor reload the file and allow edit original content in different charset.

Sorry, but I have not better way to explain it :-)

qooob:
Why not using Unicode? I suppose, you can't have different char-sets in one file, unless it's `UTF-8`.

World is moving to Unicode. That is the whole point!

gnadelwartz:
thats whats webmin has to do :-)

qooob:
We'll do!

gnadelwartz:
detection of  Encode::Detect::Detector::detect seems to be really random, it detects  an de.UTF-8 file as japanese:

grafik

gnadelwartz:
what about a second guess if its iso-8859-x and fallback to iso-8859-1 if it can be more than one of them?

qooob:
It could happen for two reasons:

1. There is too little data in the file to get encoding correct;
2. Encoding is not detected at all and it feel back to the previously was chosen by you.

What you should do is to change the encoding of the file. In case server returns an empty encoding, it will use previously chosen encoding by the user. 

In practice a person usually works with only one type of encoding. It's meant that encoding is auto-detected. When it doesn't happen, previous user chose is forced on all other files.

Give it a try.

gnadelwartz:
I tryed it:

1. change encccoding in editor and close it
2. open the file agian, comtent is diplay for a second, then cange to empty
3 file min shows file is not empty

grafik

grafik

this remains also after logout/login for the file :-(, I'll try if flushing cache helps

gnadelwartz:
after fulshung browser cache and restart noting changes empty edit window ... in chrome its ok, so flushing cache does not help. Cookies?

qooob:
No, it has nothing to do with cache. What `data-encoding` attribute's value on form tag of opened editor window?

gnadelwartz:
utf-8 is shown in select, form is:

`<form class="ui_form" role="form" action="save_file.cgi" method="post" data-encoding="">`

qooob:
Like I said. This case scenario means it's not detected automatically. Now choose decided encoding one more time, close the editor's window and open it again.

gnadelwartz:
I tryed every charset in the dropdown, but noting diplayed after reopen file. for other files its working.

I check also in Chrome. on first open its dipalayed as utf8, if I switch to iso-8859-1 and reopen, nothing is displayed. after switching back to utf8 and reopen, also nothing is diplayed.

qooob:
??

Are you sure that you didn't break `edit_file.cgi` file? If you revert to the default version of it, still the same?

gnadelwartz:
OK, to be shure everything is like it should I reinstalled webmin from rpm, restart webmin and logged out and in, nothing changed:

```
rrr:/usr/libexec/webmin # cat version
1.840
rrr:/usr/libexec/webmin # ll version filemin/edit_file.cgi
-rwxr-xr-x 1 root root 2514  7. Mai 00:47 filemin/edit_file.cgi
-rw-r--r-- 1 root root    6  6. Mai 21:08 version

```
now updated theme also:
```
Pulling in latest changes for Authentic Theme (https://github.com/qooob/authentic-theme)...
Updating to Authentic Theme 18.49-2, done.
```
after changing a files encoding its no more displayed in editor, even it switch back to original encoding.
happens in webmin and usermin

may be proxy setup?

gnadelwartz:
edited edit_file.cgi and forced file encoding: 

`$encoding_name = "iso-8859-1";`

now its diplayed again ...

qooob:
Are you talking about the file (module.info) that you sent me before today?

qooob:
When I click on `change encoding select`, it really changes it. In case server response is empty on detected encoding, the encoding select first will fall to _UTF8_. In case I change it to something else, like _KOIR-8_, then next time I open undetected file it will be my last choice, which is _KOIR-8_ in this example.

It just works. I spent a lot of time testing it. It was fine. I'm trying it now and it works as well on the files you sent me.

I'm still not very sure what result are you expecting?

gnadelwartz:
It with every file. If I klick here and change the encoding
grafik
the file content is no more displayd

qooob:
What do you have on Network tab happening in the console?

By the way, you don't have latest theme `18.49-2` installed!

qooob:
Install the latest and tell me how it feels?

gnadelwartz:
I have:  Authentic Theme 18.49-2

also a nice effect: its working again without any change in xxxmin!

not really without any change, but I noticed no charset was detected when the strange effect appears, even it should, so I refresh all installed perl modules from CPAN. this resulted in detecting encodings again, even if they are not the ones I like.

in webmin now I can also change charsets without loosing editor content, in usermin it is like before content vanishes, but is back after open again. mayby I have to wait until tomorow and see what happens :-)

gnadelwartz:
since it seems not a webmin issue I close the pull request

gnadelwartz:
in usermin was not the latest theme, no switching lang is working as expected, content is no more vanishing. Thanks!

qooob:
Reinstall Webmin/Usermin from repos. Then install the latest theme using script. Restart services. Clear browser's cache.

I'll work then.

gnadelwartz:
I played around with you theme update, did not notice much changes, but it looks like german is displayed in more places. Did you change something, like fallback from de.utf-8 to de instead of en?

gnadelwartz:
already done :-)

qooob:
I didn't change anything bin this regard. There was a lot of change to editor. You should see menu icon next to new save button. If so, the functionality I was talking about for encodings will work. Before that, it was just the select that worked only in display mode.

gnadelwartz:
I didn't notice because I update your wonderful theme every day 😁

btw:. editor is now working like a charm, many thanks for your work!

qooob:
Danke.

PR #572, opened 2017-05-23
majordomo change index view from buttons to list (part 1)

gnadelwartz:
This is the first patch in a series to give majordomo module a modern look and to provide more info about lists at a glance.

For now I converted only `index.cgi ` from simply showing Buttons, to a table with an link to edit and an list overwiew. Next step is to fill table also with real content: List Info, Oner Name/Mail and number of users.

Even majordomo is on the list of to be removed modules, majordomo has a big fan base and there are patches to run with recent perl versions.

At latest: I'm a BIG fan of Majordomo and run 20 lists or so with it.


gnadelwartz:
grafik

gnadelwartz:
grafik

jcameron:
Thanks!

PR #573, opened 2017-05-24
correct soltion for missing col when chain is filtered

gnadelwartz:
after dealing more with tables, I can also provide the correct soltion for my ugly fix for problem of missing colum when a Chain is filtered out.

additonally a  `<br>` is inserted if a `<b>` tag is detected in a col, so i changed `describe_rule` in `firewall-lib.pl` to use `<strong>`


jcameron:
Thanks again!

PR #575, opened 2017-05-26
majordomo: finish index, delete_list, edit_members (Part 3)

gnadelwartz:


gnadelwartz:
Status: everything is working, I can set up new lists including list archive, subscribe, unsubscribe, delete list.
detection of postfix, sendmail is working, also manual path to aliases file.
only problem on my system is the owner of files in /var/lib/majordomo/lists are not correct for my setup
create_list set them 28:mail but I need mail:mail, I'll investige tomorrow on this

jcameron:
This is looking good ... let me know when it's ready to merge

gnadelwartz:
you can merge now. wrong owner was my setup fault. I had the wrong owner on `/var/lib/majordomo/lists` after changing to `mail:mail` also add_list set the correct owner.

I found an additional bug: it was not possibel to change mail settings because there was some `foreign_call('sendmail'...` left, witch has to be converted to `foreign_call($aliases_module...`

jcameron:
THanks!

swelljoe:
Hey, this is kind of amazing. I had assumed *no one* was still using Majordomo. Even majordomo2 seems to have been abandoned nearly 20 years ago. Is it still being maintained somewhere not easily found with a web search?

I had been campaigning to not bring this module (among a few others) forward to Webmin 2.0 when we fork it in a few weeks, as we're trying to reduce the number of modules we have to forward port to using a new UI and new API. I was assuming that nobody had used it in years. Knowing somebody not only is using it but willing to work on it just blew my mind.

gnadelwartz:
not really maintained but a community privididung fixes, i.e. patch for perl version > 5.12
its a typical linux/unix tool: does work and everything is in simple flat files 😁

PR #580, opened 2017-05-29
Links2button real changes - just for couriosity

gnadelwartz:
just for couriosity, same PR showing only the REAL changes after majordomo changes are merged.
as you can see it stqys as simple like before, only changed from ui-link to ui-input and everything floating in one cell. only the scrit at the end is something that souldbe not there, but I liked to showcase how it can be done in the theme 😋

can be dissmised without further notice

### DIFF
--- a/usermin/list_configs.cgi
+++ b/usermin/list_configs.cgi
@@ -33,4 +33,4 @@ print   "<script>",
                 " button.innerHTML=button.innerHTML.replace(/^/,'<i class=\"fa fa-fw fa-cog\"></i>&nbsp;');});",
 	"</script>",
 	"<style>i.fa.fa-fw.fa-cog { padding: 0 !important; color: lightgrey;}",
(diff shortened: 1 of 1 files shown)


jcameron:
Nice, but I think I'd prefer sticking with ui_link

gnadelwartz:
thanks!

gnadelwartz:
totally unrelaed question:
I use vsftp. In module section you are listed as maintianer of a vsftpmodule for virtualmin, is it only for virtualmin really? all other vsftp modules seems orphaned, site no more existing.

swelljoe:
It is not a general purpose UI for vsftpd, it just provides Virtualmin hooks so that it can turn off and on the bits a virtual domain in Virtualmin would need to provide FTP features. 

So, yeah, it really is just for Virtualmin and would not be useful (or even do anything) without Virtualmin.

gnadelwartz:
OK the I will pull one of the orphaned versions out from archive and see if they are working

PR #581, opened 2017-05-30
convert majordomo final part

gnadelwartz:
everything I wanted to do is now finished.

new in this part:
- allow to save list-owner mail in `list.owner` file in listdir instaed in aliases file
- provide majordomo download link patched for perl version > 5.12
- most important options are visibile before editing
- archive also placed in `/var/lib/majordomo/archive` and can stay after list removal


jcameron:
Thanks again!

PR #583, opened 2017-05-31
Get rid of about half of validation errors

swelljoe:
This PR fixes most of the really easy to fix validation errors related to literals in tags (e.g. "width=100%"), by quoting literals. To fix the rest it'll take modification of the modules where they occur. I may try to automate that task or may wait until 2.0 is branched, as a lot of HTML will be stripped out anyway.


gnadelwartz:
util now I learned and used always double quotes for tag parameters, does HTML 4 also allow single quotes?

swelljoe:
Sure. They are inter-changeable. In Perl, however, double quotes turn on variable interpolation...so, if we want to include Perl variables in the resulting string, we can wrap it in double quotes, and use single quotes for the HTML portion of the string.

We should probably standardize on one common format for this (there are places where we've got escaped double-quotes within double quoted strings), but it doesn't make a difference to the browser or the W3C validator (which is what this patch was intended to make happy; Ilia ran into a bug with forms not working in his new loaded for Authentic, and I suspected it was due to invalid HTML...and it was, but it can be really hard to find when there's 50+ errors in the HTML!).

gnadelwartz:
I propose to fix tag value quotes at least in module libs *-lib.pm, I have a PR doing this for majordomo-lib.pm ...

PR #584, opened 2017-05-31
Use escaped ampersand

swelljoe:
Get rid of a couple more validation problems in default ui-lib and theme.


ui-lib.pl
```diff
@@ -1442,7 +1442,7 @@ =head2 ui_buttons_start
 sub ui_buttons_start
 {
 return &theme_ui_buttons_start(@_) if (defined(&theme_ui_buttons_start));
-return "<table width='100%' class='ui_buttons_table'>\n";
+return "<table width='100%' class='ui_buttons_table'>\n<tr><td>";
```
jcameron: Is this compatible with the HTML generated by ui_buttons_row , which starts with <tr> ?

web-lib-funcs.pl
```diff
@@ -7258,8 +7258,8 @@ sub help_search_link
 if (&foreign_available("man") && !$tconfig{'nosearch'}) {
 	my $for = &urlize(shift(@_));
 	return "<a href='$gconfig{'webprefix'}/man/search.cgi?".
-	       join("&", map { "section=$_" } @_)."&".
-	       "for=$for&exact=1&check=".&get_module_name()."'>".
+	       join("&amp;", map { "section=$_" } @_)."&amp;".
+	       "for=$for&amp;exact=1&amp;check=".&get_module_name()."'>".
```
jcameron: These meed to be & though, as they are used in a URL and so don't need escaping

swelljoe: I thought so, too, but not according to the W3C, and I'm guessing they know.

http://htmlhelp.com/tools/validator/problems.html#amp

They're pretty emphatic that & alone is invalid and &amp; is valid.

ui-lib.pl
swelljoe: I believe so.

ui_buttons_row needs to also have a full table/tr/td wrapping the form. It is invalid to have a table/form/tr/td, and it breaks event-handling on the page when loaded via JavaScript (this is the bug Ilia was seeing in Authentic, and we'd see if here, too if we were to load the page via XHR or fetch or whatever JS loader).

I think I'd like to get rid of ui_buttons start. It's just an extraneous table wrapping the whole thing...it's a unneeded extra layer of table nesting. But, tons of stuff calls it, and I suspect making it a no-op would cause formatting problems. But, it's probably worth testing.

So, in short: A table can be contained inside of a tr/td, but a form cannot be contained in a table (without a tr/td outside of it). As it stands, we either need this to be a fully formed table wrapper with tr/td or we need it to not stick a table tag in there at all.

qooob:
I ran having multiple issues due to that in the past, with other app. It happened that it's easy to mix things around.

>Note that replacing & with &amp; is only done when writing the URL in HTML, where "&" is a special character (along with "<" and ">"). When writing the same URL in a plain text email message or in the location bar of your browser, you would use "&" and not "&amp;". With HTML, the browser translates "&amp;" to "&" so the Web server would only see "&" and not "&amp;" in the query string of the request.

qooob:
> ..ui_buttons_row needs to also have ...

Joe, are you saying it's not going through UI lib and thus can't be easily fixed? I think it is. We can completely remove tables and use some other block elements.

swelljoe:
No, I was talking about ui-lib function ui_buttons_row.

If there are places where ui_buttons_start is used without ui_buttons_row or ui_buttons_row is used without ui_buttons_start, there may need to be some changes, but I actually think this change makes either case more robust. (Because each is self-contained and doesn't assume another function will provide the necessary tags in the middle.)

I think, but haven't tested, we could just kill ui_buttons_start/ui_buttons_end. I think they're extraneous now...maybe always were, and we just didn't notice.

jcameron:
If there are modules that don't  follow the expected order of ui_buttons_start / ui_buttons_row (1+times) / ui_buttons_end , they should be fixed.

qooob:
It was rather broken HTML, now it's fixed. It looks great and most importantly it works as designed.

jcameron:
Ok, so is this patch still needed?

swelljoe:
Jamie are you asking if this PR with the HTML validation fixes is needed?

If so, then yes. We need this. It could be overridden in the theme to make it right in one theme, but why? This patch just fixes broken HTML. I've been encouraging Ilia to pass things like this up to me so I can fix them upstream; broken HTML will affect any theme ever. No reason for us to spit broken HTML that breaks event listeners.

If this change caused any layout bugs, I'll be happy to fix them. But, from what I can tell, the browser was *already* silently inserting this extra table to make our HTML make sense. Browsers are very forgiving, especially with these old doctypes.

jcameron:
Yeah, I was wondering about the original PR.

Are you sure that using &amp; is correct in a URL that is being used in a link? I'm pretty sure that just & works fine.

swelljoe:
Yes, I'm certain of it. The W3C leaves no room for interpretation there:

"To avoid problems with both validators and browsers, always use &amp; in place of & when writing URLs in HTML"

http://htmlhelp.com/tools/validator/problems.html#amp (this is a W3C site).

swelljoe:
Oh, actually, that's not the W3C site, but I found it there, too...gimme a sec...

swelljoe:
Hmmm...actually again...it may be OK to leave it unescaped in HTML5 (but not in older doctypes).

swelljoe:
Since we're moving to HTML5 soon, let me research it a bit more. The other changes are needed, but if HTML5 allows unescaped amp, I'll roll that one back.

qooob:
We have to escape `<`, or `&` when followed by anything that could begin a character reference.

> Character references, other elements, and comments, but the text must not contain the character `< (U+003C)` or an ambiguous ampersand. [*Ref](http://w3c.github.io/html/single-page.html#syntax-text)

qooob:
> .. if HTML5 allows unescaped amp ..

It doesn't.

jcameron:
I sure hope & is allowed in links, as there are literally thousands of occurrances of it in Virtualmin.

qooob:
That is strange, I just created a [file ](https://rostovtsev.ru/pub/src/authentic/18.50/HTML5-amp.html) and run it using [W3C Validator](https://validator.w3.org/nu/?doc=https%3A%2F%2Frostovtsev.ru%2Fpub%2Fsrc%2Fauthentic%2F18.50%2FHTML5-amp.html) and without escaped `&` it was valid.

Anyway, then it's even easier for us. No need to go through all of the code plunking it out.

swelljoe:
HTML5 allows unescaped amp, if it is unambiguous. I mentioned that in the pull request that rolled this change back. It is *not* allowed in HTML 4.01 strict or transitional, but it it *is* allowed in HTML5 as long as it is unambiguous. If the browser can mistake it for an entity, it is an error in HTML5, as well...but, hopefully that won't be the case in any of our code. (note the inclusion of the word "ambiguous" in your quote from the spec above...that's important. It literally means *only* ambiguous use of ampersands is prohibited.

I don't like that we're not validating with the current doctype, but it'll solve itself when we fork 2.0 and I switch Webmin's doctype to HTML5. Webmin has never spit valid HTML, but this is not one of the problems we have to fix, and it is not one that will cause errors in events on XHR loaded pages.

PR #601, opened 2017-06-20
Firewall/Firewall6: analyse and unify shared code, may result in one iptables firewall module

gnadelwartz:
firewall and firewall6 are nearly the same the main difference is the version of iptables used.
I was guessing most of the code and most of language strings are the same and we can do some unification, which may lead to merging them in future.

Why? I don't like maintaining the same things twice and while firewall get new features and fixes, firewall6 does not.

After a look to the help and lang files, I noticed there 99% the same and I deceided to start with them.
in help files the only difference was: `iptables` vs `iptables6`. in lang files the difference was only one string added and two renamed to contain a `6`, but with the same content.

this RP does: 

- change wording in firewall help files from `iptables` to `ip(6)tables`
- add the missing and renamed strings to firewall lang files
- remove all firewall6 help/lang files and replaced them with symlinks to the firewall files. 

Hope you like it ...


gnadelwartz:
oups, I can do it much more easier by replacing lang/help the directories ...

gnadelwartz:
after some investigation, code share in firewall(6)-lib is also >95 % ... I'am working on an unified version firewall-lib.pl which can be included from firewall-lib4 and firewall6-lib

the main difference is the used save-file and the itptables commands

gnadelwartz:
tested so far:

Messages and Help seems to be correct in firewall6 after unifiying lang and help files:
grafik

new, minimal `firewall4-lib.pl` and unified `firewall-lib.pl` works for IPv4. Its not in this PR but I'm really suprised that it runs for IPv4 without any error or problem until now!

jcameron:
Yeah, the firewall6 module was forked and edited by a contributor at some point in the past. In an ideal world there would be just one module, but given a choice between not having IPv6 support and having it imperfectly implemented, I chose the latter..

gnadelwartz:
ok, then lets start to share as much as possible between the two modules and see how far  it goes

gnadelwartz:
It looks like he doesn't change any code, only replaced `iptables` with `ip6tables` so I merged also the *lib.pl's. Additionally I implemented a pseudo tabbed interface to switch between IPv4 and IPv6 to give an impression how it a merged module may look..
grafik
grafik

If it goes forward this easy we may have one iptables module  end of next week ...

gnadelwartz:
next step is done, unified index cgi and move firewall6-lib from firewall6 to firewall and replaced by symlinks. from a first diff all other .cgi are more ore less identical in both dirs.

jcameron:
This is nice .. let me know when it's done, and I will merge into the webmin repo.

gnadelwartz:
after fixing my `update-from-repo.sh` script its easier for me to test the changes made, e.g. if symlinks are resolved correct.

gnadelwartz:
for now I'm finished, pls merge PR.

tested so far:

- update from repo works (symlinks, compliling UTF-8)
- firewall6 and firewall use same *.cgi scripts
- protocol is correctly detected by location (firewall6) or parameter `version=inet6` (firewall)
- create, move, delete rules and chains is working for both protocols
- save to correct save_file is working

I will stop unifying for now and test more.
what I can't test is cluster setup.

bugfixes and merging the modules as final step will be a seperate RP

gnadelwartz:
grafik
grafik
grafik

gnadelwartz:
what is `open-ports.pl` for? Its not called from any script within `/firewall`?
Somehing similar for `setup.cgi` here I guess its call from webmin at some point but not from within `/firewall`

gnadelwartz:
Its all im my final PR!

PR #602, opened 2017-06-22
certmgr: Updated view/manage to work with openssl > 0.9.8

Jaybus2:
Newer versions of openssl generate PKCS#8 format PEM files, rather than PKCS#1, and use SHA1 as default fingerprint, and etc. other differences that caused View/Manage page to misinterpret openssl -text output.

Also updated certmgr-lib.pl to use UI API functions, rather than direct output of HTML, to hopefully make it more theme friendly.

Additionally, this commit adds:
  * view of CRL files
  * subjectAltName (if any) now displayed in cert/csr view


jcameron:
Thanks!

PR #604, opened 2017-06-29
Log if no_pam and pam_only both are set

amishxda:
Some old config has no_pam set to 1. Now if administrator enables pam_only too then both are conflicting. Which makes webmin to exit with PAM error. But administrator can not figure out why? This logs additional line so that administrator know the reason.


jcameron:
Thanks!

PR #606, opened 2017-07-01
finished merging of firewall module

gnadelwartz:
Now firewall 6 is only a smylink to firewall.

as a consequence we have now to identical modules afer upgrading.
only difference is firewall6 defaults to inet6 while firewall defautls to inet4 as protocol version.

Why dont remove firewall6 for now? because two problems remain:

1. in cluster setup not all clusters may updated at the same time, so keeping firewall6 does not break them, but I think this is not really a problem.
2. before removing firewall6 completly we need to merge the the config of both modules in pre_install

what is your suggesion:

- don't bother and remove firewall6 on update
- try to merge config (where I need your help)

Kay

contains all changes from https://github.com/webmin/webmin/pull/601


swelljoe:
I hope/think we're forking a Webmin 2.0 branch in October (20 year anniversary of Webmin). So...maybe that's the time to just rip it out, and not worry about it breaking anything, since breaking backward compatibility is OK in a major version bump. That's just three months away...so not all that long to wait.

gnadelwartz:
sounds good to me ...

jcameron:
On the other hand, if this is ready to go I can merge it in for the next major Webmin release.

Is the functionality of the existing modules preserved? Or does this effectively remove the firewall6 module and instead make the firewall module capable of IPv4 and IPv6 ?

gnadelwartz:
we was talking about finally removing firewall6
for now we have firewall capable using v4 and v6 and  "clone" firewall6 which default's to v6.

pls merge

gnadelwartz:
synced with with current webmin/master, pls merge

jcameron:
Thanks!

PR #609, opened 2017-07-03
Some fixes for perl 5.26

amishxda:
New perl 5.26 has few changes. Like:
1. A left brace '{' inside regex is not allowed. It should be escaped like this '\\{'
2. Current directory '.' is no more included in @ INC

These patches fix warnings and errors you get when running setup.sh. (when using new perl 5.26.0)

However, there may be more such files which require fixing?

### DIFF
--- a/system-status/enable-collection.pl
+++ b/system-status/enable-collection.pl
@@ -1,6 +1,7 @@
 #!/usr/local/bin/perl
 # Command-line script to enable status collection
 
+BEGIN { push(@INC, '.'); };
 use strict;
 use warnings;
 our (%config);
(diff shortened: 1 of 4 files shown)


system-status/enable-collection.pl
```diff
@@ -1,6 +1,7 @@
 #!/usr/local/bin/perl
 # Command-line script to enable status collection
 
+BEGIN { push(@INC, '.'); };
```
jcameron: Instead of this, change the "require" line below to use ./system-status-lib.pl

amishxda:
I used BEGIN method because I saw similar thing done in other .pl files in webmin.

none-the-less... changes requested are made.

jcameron:
Thanks!

PR #610, opened 2017-07-06
Let's Encrypt: FIX CN parsing to work with OpenSSL 1.1

manusfreedom:
https://github.com/diafygi/acme-tiny/commit/9537453586cd5124d5e4e46d78f9ed909180835d

CN used to be without whitespaces around the `=` but OpenSSL 1.1 introduced
whitespaces:
1.0.1: subject=/CN=example.com
1.1.0: subject=CN = example.com

This commit makes them optional.


jcameron:
Thanks!

Issue #619, opened 2017-08-17
fdisk/help files duplicate with different file name case


robertoschwald:
The fdisk/help files axist twice with different name case. This causes serious trouble on case-insensitive file systems, like UFS on OSX and irritates git.

Example:

A.ca.html
a.ca.html

Please clean up.

robertoschwald:
only works if you are on a case sensitive file system, but not on a case insensitive one as I've written. Why do you have those files twice?

chris001:
@robertoschwald  You're right.  On windows which is a case-insensitive file system, those 2 files overwrite one another.
The only reason for having them I can imagine, would be because the code refers to both upper case version and lower case version.  The fix would be to fix the code, to refer only to the lower case version. Then you could remove the upper case version from the repo.

swelljoe:
They aren't the same file. They actually have different contents and refer to different options in the UI. But, they don't need to be named this way, so it is fixable.

Looks like it's just A/a and K/k. I'll poke at it to see if I can fix it easily.

swelljoe:
Hahah...this is actually pretty clever. The files are named for the hdparm option they describe, and it's pulled out automagically from an array. Changing it is do-able, and I'm checking in a fix now and will do a pull request for Jamie to approve when he gets a moment to look it over.

swelljoe:
Fixed in this PR https://github.com/webmin/webmin/pull/620

It's likely to go into the next Webmin release, and should show up in the mainline git repo soonish.

PR #620, opened 2017-08-18
Case insensitivity fixes in fdisk help files #619

swelljoe:
This seems to work for me and should resolve case insensitivity issues on those other operating systems that can't tell the difference between a and A. ;-)


jcameron:
Thanks for the fix!

Issue #622, opened 2017-08-21
init module believes upstart exists on Ubuntu 16.04, but it's systemd


swelljoe:
Ubuntu 16.04 (and later) have systemd. But, Webmin's init module believes that because initctl command exists, it is an upstart system. initctl still exists in 16.04 but it's not an upstart system, by default...it's systemd.

It's probably better to check for systemctl, and then fall back to checking for initctl. initctl on systemd versions of Ubuntu is a shell script that just contains "exit 0", so it does nothing.

I'm not sure how the init module actually functions on these systems, since it seems like it ought to not work at all since it's trying to configure upstart. But, it does seem to kinda work. I only noticed it was broken because of an error when trying to restart a service that didn't exist ("Error: Upstart service named cannot be created unless a command is given").

PR #623, opened 2017-08-21
Check for systemd before upstart #622

swelljoe:
I think this will resolve the problem I reported with misdetection of systemd vs upstart in #622.


jcameron:
Thanks for the fix!

Issue #615, opened 2017-07-27
Webmin 1.850 behind Apache Reverse Proxy


thealliancenetwork:
I have Apache setup with HTTPS and reverse proxying to Webmin which has SSL switched off.

I followed https://doxfer.webmin.com/Webmin/Running_Webmin_Under_Apache to do this.

Webmin seems to try and return 302's in some circumstances to http://myfqdn:10000 when it should be https://mfqdn:443 it's generally actions like Login or Create directory or Chown etc.

Example from login page

```
POST https://MYFQDN/webmin/session_login.cgi HTTP/1.1
Accept: text/html, application/xhtml+xml, */*
Referer: https://MYFQDN/webmin/
Accept-Language: en-GB
User-Agent: Mozilla/5.0 (Windows NT 6.3; WOW64; Trident/7.0; rv:11.0) like Gecko
Content-Type: application/x-www-form-urlencoded
Accept-Encoding: gzip, deflate
Host: MYFQDN
Content-Length: 46
Connection: Keep-Alive
Cache-Control: no-cache
Cookie: redirect=1; testing=1

user=REMOVED



HTTP/1.1 302 Moved Temporarily
Date: Thu, 27 Jul 2017 11:49:13 GMT
Server: MiniServ/1.850
Location: http:// MYFQDN:10000/
Set-Cookie: sid=REMOVED; path=/; expires="Thu, 31-Dec-2037 00:00:00"; httpOnly
Keep-Alive: timeout=5, max=100
Connection: Keep-Alive
Transfer-Encoding: chunked

0


```

Apache Config
```
ProxyPass /webmin/ http://localhost:10000/
ProxyPassReverse /webmin/ http://localhost:10000/

ProxyPassReverseCookieDomain /webmin/ http://localhost:10000/
ProxyPassReverseCookiePath /webmin/ http://localhost:10000/
```

Webmin Config
```
webprefix=/webmin
webprefixnoredir=1
referer=MYFQDN
```

Miniserv Config
```
ssl=0
ssl_redirect=0
```

I think it's down to line 4074 - 4105 in the miniserv.pl file

I have managed to work around this by adding in the below.
ProxyPassReverse /webmin/ http://MYFQDN:10000/

qooob:
You don't have to disable SSL, it's possible to make it work by adding above _ProxyPass_:

```Config
SSLProxyEngine On
SSLProxyVerify None 
SSLProxyCheckPeerCN Off
SSLProxyCheckPeerName Off
[...]

PR #627, opened 2017-09-14
Fix login redirect from FQDM/webmin

ionbasa:
Login behind reverse proxy with SSL enabled results in a redirect from FQDM/webmin to localhost:1000 due to a an issue in the miniserv conf file. This change fixes that. Logins are now redirect properly.

### DIFF
--- a/miniserv.pl
+++ b/miniserv.pl
@@ -4096,7 +4096,7 @@ if ($ok && (!$expired ||
 			&write_data("Set-Cookie: $sidname=$sid; path=$cpath".
 				    "$sec\r\n");
 			}
-		&write_data("Location: $prot://$hostport$in{'page'}\r\n");
+		&write_data("Location: $prot://$host$config{'cookiepath'}$in{'page'}\r\n");
 		&write_keep_alive(0);
 		&write_data("\r\n");
 		&log_request($loghost, $authuser, $reqline, 302, 0);


ionbasa:
See: https://github.com/webmin/webmin/issues/615

jcameron:
Are you sure this isn't going to break redirects in the normal non-proxy case?

ionbasa:
Good point! It does seem to break redirects in non-roxy cases. I tried playing around with the following tonight and it doesn't seem to work as expected:

```
                if ($config{'cookiepath'} != undef) {
                        &write_data("Location: $prot://$host$config{'cookiepath'}$in{'page'}\r\n");
                        }
                else {
                        &write_data("Location: $prot://$hostport$in{'page'}\r\n");
                        }
```
That configuration will work for normal non-proxy, but seems to fail when proxied, am I missing something?

jcameron:
I really don't think this is the correct fix for everyone.

How is your proxy configured exactly?

ionbasa:
@jcameron It probably isn't the correct fix for everyone, but an fix for people who are having issues with reverse proxies. Did you read through https://github.com/webmin/webmin/issues/615 ? Is there My reverse proxy is though Apache 2, setup as the following: 

```
        ServerName *******.me
	ServerAdmin hostmaster@********.me
	ProxyPreserveHost On

	ErrorLog ${APACHE_LOG_DIR}/**********-error.log
	CustomLog ${APACHE_LOG_DIR}/***********-access.log combined
	RewriteEngine on

	ProxyPreserveHost On
	SSLProxyEngine On
	SSLProxyVerify none
	SSLProxyCheckPeerCN off
	SSLProxyCheckPeerName off
	SSLProxyCheckPeerExpire off
             
	RewriteRule ^/webmin$ /webmin/ [R]
	RewriteRule ^/docker$ /docker/ [R]
    
	ProxyPass /webmin https://localhost:10000
	ProxyPass /docker/ http://localhost:9000/
	ProxyPassReverse /webmin https://localhost:10000
	ProxyPassReverse /docker/ http://localhost:9000/
	ProxyPassReverseCookieDomain /webmin https://localhost:10000
	ProxyPassReverseCookiePath /webmin https://localhost:10000
	ProxyPassReverseCookieDomain /docker/ http://localhost:9000
	ProxyPassReverseCookiePath /docker/ http://localhost:9000
    
    #Enable Phpmyadmin only in this VHost
    Include /etc/phpmyadmin/apache.conf

	SSLCertificateFile /etc/letsencrypt/live/********.me/fullchain.pem
	SSLCertificateKeyFile /etc/letsencrypt/live/**********.me/privkey.pem
	Include /etc/letsencrypt/options-ssl-apache.conf
```

I have redacted some information with asterisks (*), I encourage you to read through the issue I linked. It doesn't seem to be just me who's having this issue. 

If there's another way to fix this, then I'd like to hear your thoughts. The change I PR was the only way to prevent the issue from happening on my end.

qooob:
My be related?

Trying to set to _Off_ option `ProxyPreserveHost` solves an issue?

https://github.com/qooob/authentic-theme/issues/807#issuecomment-330511928

ionbasa:
Can definitely say it is related! With the original unmodified miniserv.pl, having `ProxyPreserveHost `off solves all the issues I have been having. Having it enabled breaks the login. 

Should probably add that to the documentation. Secondly, why would enabling it break Webmin logins? Some of my back end reverse proxy apps, such as [Mattermost](https://docs.mattermost.com/install/config-apache2.html), require it to be enabled.

This makes it slightly more inconvenient now that I need seperate configs in Apache only for Webmin to have the `ProxyPreserveHost ` directive disabled.

PR #628, opened 2017-09-14
Support devuan

kwmiebach:
Adding 4 lines to os_list.txt. This makes the webmin.deb file installable on the devuan os (https://devuan.org/) and also enables to choose the correct version in a setup.sh install.

Tested on the current version "devuan 1.0.0 codename jessie" (1st line). All 4 lines were assembled in the #devuan channel on freenode.


swelljoe:
The "all known Devuan versions" patch won't work, I don't think. Webmin treats recent versions of Debian as a systemd distro, while Devuan will forever be intiscripts, as I understand it. So, I think you'd need to always set it to debian 7, or add new configs just for Devuan, but Jamie probably has a better feel for how that would play out.

jcameron:
The init mode detection happens automatically - so even if Devuan uses init rather than systemd, Webmin will work fine.

PR #637, opened 2017-09-26
restart networking service on save

adalenv:


swelljoe:
This is unnecessary and can't work (that's not valid Perl syntax, it won't compile).

swelljoe:
What problem are you trying to solve with this change?

adalenv:
i dont know perl  but network service should restart to show up changes

PR #638, opened 2017-09-26
restart networking service on save

adalenv:


jcameron:
Two things : 

1 - This won't work, you can't just start a shell script in the middle of a perl script
2 - Why do you want to do this anyway?
3 - On other linux distributions and operating systems, that command may not even exist

adalenv:
sorry i dont know per its jus a idea 
changes doesnt show up withow restarting network service

PR #643, opened 2017-10-03
init: Add extra set of reboot/shutdown buttons to top

swelljoe:
A user requested having the reboot/shutdown buttons at the top and bottom of the page (because it's so long). That seemed reasonable to me, so I did it. There are probably better ways to make this page more usable, but it'll probably do for now.

### DIFF
--- a/init/index.cgi
+++ b/init/index.cgi
@@ -8,6 +8,22 @@ require './hostconfig-lib.pl';
 &ui_print_header(&text('index_mode', $text{'mode_'.$init_mode}),
 		 $text{'index_title'}, "", undef, 1, 1);
 
+print &ui_buttons_start();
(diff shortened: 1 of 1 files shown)


qooob:
Oh no, oh no! It looks very ugly.

There is _End_ button. I really don't like it so much. Please don't add it.

swelljoe:
OK, I'll close this one. Let's figure out a better way to improve it.

qooob:
In order to have super quick access to any page, for example to _Reboot_ in _Bootup and Shutdown_ module, all you need is to add to theme's configuration _Custom Link 1_ the link of:

```Bash
/init/reboot.cgi
```

Later, anywhere in the theme clicking `Alt+1` will take you there.

PR #644, opened 2017-10-05
Maybe fix the form on edit cron page

swelljoe:
This fixes interleaved forms/tables.

A form can wrap a table in its entirety, or a form cell can wrap a form in its entirety, but form elements cannot be split across form elements or inside of a table but outside of a cell.


swelljoe:
Oops, I clearly screwed up my branching on this one. I'll do it again.

PR #649, opened 2017-10-07
Probably fix upstart vs systemd detection #648

### LINKED ISSUE #648: Detection of init system on Ubuntu is utterly borked
So, Ubuntu systems can literally have all three of the popular init systems installed, even though only one is the "real" init system. Thus, the current detection, even with my fix from a while back, is completley unreliable on Ubuntu (with my fix, it is maybe always wrong on 14.04, without my fix it is maybe always wrong on 16.04).

I don't have a great solution yet. But, I'm posting some findings so I don't forget them as I try to work out what actual detection needs to look like.

Currently, we detect based on existence of systemctl or initctl (upstart). But, on Ubuntu:

```bash
root@scw-e9ac00:~# which systemctl
/bin/systemctl
root@scw-e9ac00:~# which initctl
/sbin/initctl
```

So, that's useless.

And, I've seen other folks suggest checking for the config files, but Ubuntu has all three:

```bash
root@scw-e9ac00:~# ls -d /usr/lib/systemd/
/usr/lib/systemd/
root@scw-e9ac00:~# ls -d /usr/share/upstart/
/usr/share/upstart/
root@scw-e9ac00:~# ls -d /etc/init.d
/etc/init.d
```

So, that's useless.

We can always see what the init process is with:

```bash
# ls -l /proc/1/exe
lrwxrwxrwx 1 root root 0 Oct  6 23:15 /proc/1/exe -> /sbin/init
```

Which means it has..."init", right? Wrong. You're not gonna believe this shit:

```bash
# /sbin/init --version
init (upstart 1.12.1)
Copyright (C) 2006-2014 Canonical Ltd., 2011 Scott James Remnant

This is free software; see the source for copying conditions.  There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A
PARTICULAR PURPOSE.
```

So, that's it, maybe. It looks like we can check first for where /proc/1/exe (which is always the init) points and then, if it's `/sbin/init` check the output of `--version`. Traditional init doesn't have a `--version` flag, so it'd result in:

```bash
root@ubu16:~# /sbin/init --version
/sbin/init: unrecognized option '--version'
```

So...in pseudo-code:

```
if (/proc/1/exe == /sbin/init) {
  if (/sbin/init --version == /upstart/) {
    return 'upstart'
  }
  else {
   return 'init' # Or whatever old initscripts are called
  }
}
elsif (/proc/1/exe == /lib/systemd/systemd {
  return 'systemd'
}
```
Am I missing anything?

swelljoe:
This fixes #648.

My "fix" from before actually broke detection for upstart systems that also have systemd installed (while fixing it for systemd systems that also had upstart installed).

Ubuntu 14+ have *all three* init systems installed in some form, so detection has been broken for some subset of Ubuntu versions ever since the switch to systemd. I believe this gets it right for all currently supported versions of Ubuntu.

### DIFF
--- a/init/init-lib.pl
+++ b/init/init-lib.pl
@@ -57,13 +57,15 @@ elsif ($config{'hostconfig'}) {
 elsif ($config{'rc_dir'}) {
 	$init_mode = "rc";
 	}
+elsif ($config{'init_base'} && -d "/etc/init" &&
+       &has_command("initctl") &&
+       &execute_command("/sbin/init --version") == 0) {
+	$init_mode = "upstart";
+	}
 elsif ($config{'init_base'} && -d "/etc/systemd" &&
        &has_command("systemctl") &&
        &execute_command("systemctl list-units") == 0) {
 	$init_mode = "systemd";
-	}
-elsif ($config{'init_base'} && -d "/etc/init" && &has_command("initctl")) {
-	$init_mode = "upstart";
 	}
 elsif ($config{'init_base'}) {
 	$init_mode = "init";


chris001:
Great work, it's a tricky puzzle to solve.
Here's some more tips:
* https://unix.stackexchange.com/questions/196166/how-to-find-out-if-a-system-uses-sysv-upstart-or-systemd-initsystem
* https://unix.stackexchange.com/questions/18209/detect-init-system-using-the-shell

PR #651, opened 2017-10-10
fix missing  &ui_columns_end()

gnadelwarty:
with authentic theme 19 I recognized a not working button beacuse of a missing call to &ui_columns_end()
pls apply


jcameron:
Thanks!

PR #652, opened 2017-10-11
Check for trigger URL to simply redirect to root

qooob:
This is required for Authentic Theme 19.00+

https://github.com/qooob/authentic-theme/issues/844
https://github.com/qooob/authentic-theme/issues/848

### DIFF
--- a/web-lib-funcs.pl
+++ b/web-lib-funcs.pl
@@ -4686,6 +4686,11 @@ if ($module_name && !$main::no_acl_check &&
 		}
 	$main::no_acl_check++;
 	}
+	
+# Check for trigger URL to simply redirect to root: required for Authentic Theme 19.00+
+if ($ENV{'REQUEST_URI'} =~ /xnavigation=1/) {
+	redirect($gconfig{'webprefix'} || "/");
+	}
 
 # Check the Referer: header for nasty redirects
 my @referers = split(/\s+/, $gconfig{'referers'});


web-lib-funcs.pl
```diff
@@ -4686,6 +4686,11 @@ sub init_config
 		}
 	$main::no_acl_check++;
 	}
+	
+# Check for trigger URL to simply redirect to root: required for Authentic Theme 19.00+
+if ($ENV{'REQUEST_URI'} =~ /xnavigation=1/) {
+	redirect($gconfig{'webprefix'} || "/");
```
qooob: It seems that rediret already dealing with webprefix, so it only should be 

```Perl
redirect("/");
```

jcameron:
Can the check be made tighter, like limiting it to URLs that end with `xnavigation=1` ?

qooob:
No, because in order to to make theme distinguish modes (either default, with left menu and working as SPA or just as module, without the menu and no extra JavaScript part).

You can pull latest git version and try. First open theme as usual. You will see that upon navigating the URL will always be changing and `xnavigation=1` query is forced to be appended all the time. Now right click on some link in the navigation menu and open it in new tab, where URL stays intact. 

The point of adding extra query to URL is reload. Anytime you close/open the browser or hit reload in the browser, it would be reloaded to no-navigation menu mode. To avoid that theme will get user to previously used module upon reloads.

I didn't want to remove the ability to bypass theme's complex JavaScript SPA part and let it work as it is, when opened by link in a separate tab.

In case I'm missing something let me know.

jcameron:
Ok, thanks

PR #656, opened 2017-10-15
fix firewall6 module.info after firewall6 is a symlink

gnadelwartz:
For now /firewall6 is a symlink to /firewall. So a "cloned" module will be created by `makedist.sh`.

After updating to newest webmin and pull the latest changes from github repo I noted that both modules show up with the the name "Linux Firewall". This is not a problem becaue everthing is working, but may irritate users.

I fixed this by replacing the single `firewall6` symlink with individual smylinks from all files/dirs from firewall6 to firewall and placing the old `module.info` file from firewall6 into the firewall6 directory.

### DIFF
--- a/firewall6
+++ b/firewall6
@@ -1,1 +1,0 @@
-firewall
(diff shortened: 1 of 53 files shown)


jcameron:
Thanks!

PR #658, opened 2017-10-19
fix wrong executable permission on manual.sxw

yangfl:


jcameron:
Your other commits look good, but why remove Authen-SolarisRBAC-0.1.tar.gz ?

yangfl:
Authen-SolarisRBAC-0.1.tar.gz is already extracted as folder, thus it should be managed by git.

If you don't agree I can modify my commits and leave that file exist (since that's not the main topic of this pr).

jcameron:
The tar.gz file is needed, as it contains a perl module that gets installed by Webmin on Solaris systems.

yangfl:
So what about https://github.com/webmin/webmin/tree/master/acl/Authen-SolarisRBAC-0.1 ?

yangfl:
ok, fixed

jcameron:
Thanks for the patches!

PR #674, opened 2017-10-29
status: Support 307 and 308 HTTP redirects

mnoorenberghe:
The differences from 302 and 301 are only how methods change.

307: https://tools.ietf.org/html/rfc7231#section-6.4.7
308: https://tools.ietf.org/html/rfc7538


jcameron:
Thanks!

Issue #678, opened 2017-11-03
Proposal: enhance ui_link and ui_links_row to pass hints/classes for theme


gnadelwartz:
in the firewall module janie [accepted](https://github.com/webmin/webmin/pull/676) a "hack" from me to visually improve the links to switch between IPv4 and IPv6. IMHO this should be done by theme.

A possible way to make it easyer for the theme author is to allow passing hints (classes) in ui_link_row.
today all classes from a link `<a class="danger" href="...">xxx</a>` are changed to `<a class="ui-link" href="...">xxx</a>`.

we can use a defined ste of classes e.g. success, danger, warning, info etc. like in bootstrap.
In the firewall  example the code can look like this:

ui_link_row(['<b class="info">IPv4</b>','<a class="primary" href="xxx">IPv6</a>']);

jcameron:
Interesting idea ... but what would the hints be used for? Coloring the links?

gnadelwartz:
clororing and everithing else the theme author likes todo :-) e.g. today Ilja converts ul-link to a button. with this hint he can also set a bgcolor for info.

the general idea comes from my recent use of bootstrap and is not limited to ui_link. Since Ilja already includex bootstrap, we have the possibility to easyly use it from webmin. Other themes without bootstrap can define their own css if they want to support it, if theme does not kown about its dislayed like today.

e.g. we can hint the firewall warnings with "alert alert-danger" and authentic display it automatically this way:
![grafik](https://user-images.githubusercontent.com/4593242/32406608-cb56ae04-c17b-11e7-87f0-ff623602c473.png)
```
print "<div class=\"alert alert-danger\">",
                   &text("index_$word", "$gconfig{'webprefix'}/$word/", $fwconfig),
                   "</div><p>\n";
```
Grey Framed theme:
![grafik](https://user-images.githubusercontent.com/4593242/32406654-87279a6c-c17c-11e7-9f4c-513e551bb6cc.png)

as you can see its backwards compatible, so it may be a nice extentsion.

qooob:
@gnadelwarty Kay, ideally we should separate logic and presentation completely. 

Have a look at `theme_ui_alert_box` sub, just do what you do using it like

```Perl
theme_ui_alert_box('It appears that Shorewall is ....', 'danger')
```

You should get your alert box.

qooob:
My bad, you shouldn't call `theme_ui_alert_box` but `ui_alert_box`.

gnadelwartz:
[...]

PR #680, opened 2017-11-04
use ui_alert_box for warnings  in firewall index.cgi

gnadelwartz:
as advised in https://github.com/webmin/webmin/issues/678


jcameron:
Thanks!

PR #684, opened 2017-11-08
start replace HTML with ui_lib Calls in majordomo

gnadelwartz:
Wow, majordomo modules seems to be very old, because a lot of HTML erverywhere.
I'll need many pulls to conert over to ui_lib.

pls merge


jcameron:
Thanks!

PR #709, opened 2017-11-29
De-uglify multiselect move arrows with fUnicode!

swelljoe:
This changes the ugly '->' and '<-' arrows to Unicode pointy shapes that look much nicer. These glyphs have been part of the Unicode standard for many years and are available in the fonts used in all of our themes. I can't think of any obvious compatibility issues. I've tested across all of the browsers I have access to.


swelljoe:
To update, I just looked it up, and these glyphs were added in Unicode version 1.1.0, which was from June 1993. I think we're safe using these glyphs.

jcameron:
Nice!

qooob:
I even added more coolness to it. It's the last commit I'm making in a bit before releasing 19.04.

screenshot from 2017-12-03 22-55-38

Buttons will react to hover as well.

PR #724, opened 2017-12-17
Provide an ability for user to open link from outside @authentic-theme#841

qooob:
I don't see the other way to make this work. At the moment, user can open any module without navigation menu.

When opening link from the outside (like email notification), referre is not set and the warning message appears.

We use supplied URL, filter it and store it on the disk; in a split second, when theme checks for goto tmp variable, it does checking and then tried to open the link using AJAX call.

### DIFF
--- a/package-updates/update.pl
+++ b/package-updates/update.pl
@@ -46,7 +46,7 @@ if ($tellcount) {
 	&get_miniserv_config(\%miniserv);
 	$proto = $miniserv{'ssl'} ? 'https' : 'http';
 	$port = $miniserv{'port'};
-	$url = $proto."://".&get_system_hostname().":".$port."/$module_name/";
+	$url = $proto."://".&get_system_hostname().":".$port."/$module_name/?xnavigation=1";
 	$body .= "Updates can be installed at $url\n\n";
 	}
 
@@ -67,4 +67,3 @@ if ($config{'sched_email'} && $body) {
 		print STDERR $body;
 		}
 	}
-
--- a/web-lib-funcs.pl
+++ b/web-lib-funcs.pl
@@ -4702,8 +4702,27 @@ if ($module_name && !$main::no_acl_check &&
 	}
 
 # Check for trigger URL to simply redirect to root: required for Authentic Theme 19.00+
-if ($ENV{'REQUEST_URI'} =~ /xnavigation=1/) {
-	redirect("/");
+if ($current_theme =~ /authentic-theme/ && $ENV{'REQUEST_URI'} =~ /xnavigation=1/) {
+
+  # Redirect user to the entered URL upon next page load; store URL into temporary file
+  if ($main::session_id && $remote_user) {
+      my $xnav = "xnavigation=1";
+
+      my $protocol = lc($ENV{'HTTPS'}) eq 'on' ? "https" : "http";
+      my $url = "$protocol://$ENV{'HTTP_HOST'}$ENV{'REQUEST_URI'}";
+      $url =~ s/[?|&]$xnav//g;
+      $url =~ s/[^\p{L}\p{N},;:.%&#=_@\+\?\-\/]//g;
+
+      my $tmp  = 'tmp';
+      my $salt = encode_base64($main::session_id);
+      $salt =~ tr/A-Za-z0-9//cd;
+
+      my %var;
+      my $key = 'goto';
+      $var{$key} = $url;
+      write_file((tempname_dir() . '/.' . $tmp . '_' . $salt . '_' . get_product_name() . '_' . $key . '_' . $remote_user), \%var);
+  		}
+  redirect("/");
 	}
 
 # Check the Referer: header for nasty redirects


iliaross:
Reference:
https://github.com/iliaross/authentic-theme/issues/841

jcameron:
This seems very theme-specific, especially the part where the link is in an email (because you can't guarantee that the same theme will be in use by whoever opens the email).

Also, what is the file in /tmp for?

qooob:
The point is, that this patch is universal. In case the default theme is not Authentic, then it will be ignored and you can see it from my patch. Right now, if you subscribe to receive email notifications about new packages, things will not work correctly when link is clicked from the email, because the theme wouldn't know what to do. What would happen now, is that only right side, without left menu will be displayed or at worse, the referrer error will pop-up.

What my patch does is that it checks if Authentic is in use (`$current_theme =~ /authentic-theme/`). Then, if so, it also checks if it has special trigger, that tells Authentic to use left menu and no only the module part (`$ENV{'REQUEST_URI'} =~ /xnavigation=1/`).

Besides, about your question what `/tmp` is for there. It's the whole point. We need to store currently entered URL, that comes from `REQUEST_URI` on the disk until the redirect part happens (because we have `xnavigation=1` trigger in URL already.) On load the theme will read it and parse it. In case the value passes the checks, it will be opened using AJAX call or otherwise fall back to defaults.

The created file will look like this. It's name session based so in case of multiple browsers it will also work correctly.

```
.tmp_YWI1ODk5Y2Q4Y2ZjOGYwZTIxMjAxZWYzOGZjNGM0ZTU_webmin_goto_root
```

File contents would look like this (just an example, in case of package updates it will look different):

```
goto=https://debug-centos:10000/useradmin/edit_user.cgi?user=root

```

Like I mentioned already, the theme will parse it and then, open it.

It's safe (as far as I see it) and seems like the only way to do it.

jcameron:
Ok, it's starting to make sense. 

However, instead of hard-coding /tmp , can you use the webmin temp dir like the `tempname` function does : https://github.com/webmin/webmin/blob/master/web-lib-funcs.pl#L232

Also, where does this temp file get read?

iliaross:
Sure, Jamie.

I will have to export your function thought in the theme as well, to make it work without WebminCore on XHR calls, for speedy responses.

It's read here:

https://github.com/iliaross/authentic-theme/blob/master/authentic-init.pm#L995

and  here:

https://github.com/iliaross/authentic-theme/blob/master/authentic-lib.pm#L2739

I will commit as I come back a bit later.

qooob:
Have a look at commit above. https://github.com/webmin/webmin/pull/724/commits/1b9b7ce5f8f969af845a08bfee15055b8c498e74

qooob:
It it better now with dehard-coded temp dir?

jcameron:
That's an improvement, although I only just added that new `tempname_dir` function so it won't be included till the next Webmin release. 

However, aren't there other places in Webmin / Virtualmin where a URL for inclusion in an email is constructed? How about if the theme had some way of overriding that URL?

qooob:
I always tend not to disturb you and do as much as possible on the theme side.

It happens before the theme is loaded and before anything I could do.

In case I see everything correctly, there is no way to do it on the theme level but has to be done before referre error message is generated. Without that referre check error I think I could do it.

jcameron:
Regarding the referrer problem - Webmin's existing referer checks shouldn't kick in when the request is to a module's `index.cgi` script, as they are assumed to be safe.

qooob:
I know from looking at the code a while ago, but then we face the issue, where opened links will not have left menu part.

iliaross:
https://github.com/iliaross/authentic-theme/issues/841

jcameron:
Is there a better way this can be handled in the theme? It seems like there is a deeper issue where theme URLs are not "persistent", in the sense that copying and pasting a URL like https://yourserver:10000/software/?xnavigation=1 will not open the same module.

qooob:
My initial point was to let modules work as they are if opened in new tab, without left menu and all other background calls and complexity.

Make sure that you have latest pull and try to go to `/software` for example (using navigation menu) and then reload the page or copy/paste it to the new tab (same browser), it will open correctly, because theme stores currently used page (in case it's allowed to be stored, for instance, pages that are progressive are not stored). After you reopen/reload it, theme reads goto variable from the file and opens it. In case goto file is not there or it's a new session, theme default settings and/or user preferences are used to load defaults.

Without the patch above, all the theme gets is redirect.

It's technically not possible to overcome this, unless I'm wrong.

qooob:
What do we do with that?

jcameron:
Can't the current page be fetched from the URL though? Storing it in a server-side file seems like it could get out of sync with what's actually shown in the browser.

qooob:
There is no no other way to do it, at least I don't see any other way. The page is always in sync and serves the purpose well.

Please give it a try yourself.

I agree, that adding code to miniserv, that looks like dependency is a bit ugly and wrong, but again, I don't see how it's technically possible to do it otherwise.

If you know the solution and its technical logic, share it and I'll do it.

gnadelwartz:
why is it not possible to do a redirect with the content of the file attached as parameter

`https://debug-centos:10000/?xnavigation=1&goto=https://debug-centos:10000/useradmin/edit_user.cgi?user=root`

then parse the url and load the content?

qooob:
Because it happens before the referre check. In case referre check is off, then it's no problem, otherwise the error pops up. This is why I have `xnavigation=1` check and redirects to the `/` before referre check.

Jamie, what makes you confused? The extra code on Webmin lib or the solution? I'm opened to any solutions that would be safe and would work.

qooob:
However, I remember looking at the code that referre check was okay, when opening `index.cgi` links, so probably, fixing existing `xnavigation=1` redirect, might work for limited amount of links. But just not for all.

gnadelwartz:
If I understand correctly, a link in an e-mail will also fail on referer check if it's not index.cgi?

qooob:
Yes, correct. I need to do more testing though.

qooob:
@jcameron I was wrong using single `tempdir_name` function, the best is to use `tempname`. That caused issues obviously.

I'm making a patch now..

qooob:
> If I understand correctly, a link in an e-mail will also fail on referer check if it's not index.cgi?

No, it will not work properly. It will open the link but will have no left menu. Theme doesn't load left menu for module and its pages. It uses them.

This commit https://github.com/webmin/webmin/pull/724/commits/1b9b7ce5f8f969af845a08bfee15055b8c498e74 is wrong. I will fix it for writing file using `tempname()`.

jcameron:
I think the better way to do this is to change the links used in the email to a format the theme can accept. I assume that there's some way to create a URL to a module (which could be any format you like) which shows the left page correctly?

We could implement this as a new "webmin URL" function that the theme could override.

gnadelwartz:
I thought again about it: if the link can be shown, even without menu, it must have already passed the referer check and authentic should be able to detect its running in single page mode.

I see two possibilities to react to this:

- simply place a home button to show the main page with menu
- bust the "frame" by redirect to main page with the "frame" url added as extra parameter e.g webmin:10000/?goto=/software

in the second case the redirect should also pass referrer check because its initiated by user. authentic now can read the page to display from url instead from file and load it for displaying.

to prevent this from misuse the goto parameter has also to pass some referrer check

qooob:
>  to do this is to change the links used in the email to a format the theme can accept

Jamie, it's not all about format. You understand better than me, that everything happens before referre check is not going anywhere. It's either redirect or and error message.

I'm closing this pull request. If you change your mind, let me know, and I'll provide with better patch. My current solution is the only way. The theme will have to store redirects the same way. Working as SPA it's not possible to make things work as they used to be.

Theoretically, in order to have left menu appear directly in module, I would have to completely change the way things work right now.

At the moment theme is loading itself, with left menu and main container. Later, it pulls other module's content, that is free from extra stuff. In case we make things work as you wish, we would have to make each module have all of the content and be treated as complete application. This is not only going to be potentially slower but will also not make possible to simply open module in a separate tab without fancy processing and without having left menu. (I realise that left menu can be collapseable).

The only right-now possibility is to make home/back button to take user back to the very index.

Like I said, if you have concrete, technical solutions, go ahead, I'm ready for everything reasonable/better.

qooob:
> to keep it simple I would prefer the home button ...because sometimes it may useful not to display the menu, eg opening the link on smartphone or debugging

I'm surprised why not using the theme's core code, that is already working fine to let it do the job?

Jamie, you can make your own patch, where Webmin stores URL on each request as:

```
$url = "$gconfig{'webprefix'}$ENV{'REQUEST_URI'}";
```

and saves it to the file as `goto=$url`. Put this to default Webmin `/tmp/.webmin/anyfilename_username` and that would be enough, I would be fine from here.

gnadelwartz:
> The only right-now possibility is to make home/back button to take user back to the very index.

lets start with this small addon and see how to proceed...

qooob:
> lets start with this small addon and see how to proceed...

Added it with latest commit.

jcameron:
@iliaross  - what if links to Webmin from emails already had `?xnavigation=1` appended? This is what I was suggesting in my comment about adding a theme-controllable function to modify URLs used in emails. I'd much prefer that to using a temp file, as this seems vulnerable to problems if the same user is accessing Webmin from multiple tabs at the same time. Also, there are potential XSS / malicious link implications that I haven't even thought about yet :-)

qooob:
In case there is `xnavigation=1` on the link, the page should/will reload. If there is such query, it will hit referre check, - no difference.

I don't see how XSS could be implemented using env referrer, I don't see it's possible anyway. Theme will not open such link during check, besides it wouldn't be opened anyway due to CORS.

I find it safe. Like I mentioned, having prefix+refererre is more than enough. You could check in regex that link has no `//` or `://` if you wish. 😉

PR #730, opened 2017-12-22
Don't break the initial order

qooob:
Jamie, Authentic Theme utilizes `history` and `history -d` command.

Once unique is applied everything breaks, obviously.

It has to go.

### DIFF
--- a/shell/index.cgi
+++ b/shell/index.cgi
@@ -23,7 +23,7 @@ else {
 	open(PREVFILE, $prevfile);
 	chop(@previous = <PREVFILE>);
 	close(PREVFILE);
-	@previous = &unique(@previous);
+	@previous = @previous;
 	}
 $cmd = $in{'doprev'} ? $in{'pcmd'} : $in{'cmd'};
 


jcameron:
I"m confused ... using &unique should just filter out repeated occurences of commands in the history.

qooob:
But theme let's you use `history -d` and it requires proper index to properly parse `shell/previous.username`.

You can test it with my latest commit.

jcameron:
How do we avoid the same command appearing multiple times in the history then?

qooob:
It's must have for Authentic to make `history` and `history -d` work correctly.

I have nothing to offer but make a theme check =~ /authentic-theme/ and left unique for all others.

By the way, theme checks which she'll to use and in case Cloudmin is used, it will call for it's shell.cgi file, so that one would also need changes to be reflected.

jcameron:
Can you point me to the authentic theme code that uses the `history` command?

qooob:
Yes, sure. Open theme's command shell port by clicking `Alk+K` or clicking console button, here:


screenshot from 2017-12-24 11-00-09

Than run at least one command to make sure it's in sync:

```
ls -lsa
```

Then type 

```
history
```

or
 
```
history -d n
```

To remove history by it's number.

qooob:
In case you wanted to look at the code, it's here at [authentic-lib.pm](https://github.com/qooob/authentic-theme/blob/master/authentic-lib.pm#L2742-L2750) at line [L2742-L2750](https://github.com/qooob/authentic-theme/blob/master/authentic-lib.pm#L2742-L2750)

jcameron:
Would it work for you if we kept all the history in the file, but only displayed unique commands in the Command Shell module?

qooob:
Unfortunately, not because theme uses's response and maps it into array, something like this:

```JS
let history = $(res).find('select[name="pcmd"] option').map(function() {
                    return convert.html_scape($(this).val());
              }).get();
```

qooob:
Can't we just go with

```Perl
my @previous;
if ($current_theme =~ /authentic-theme/) {
     @previous = @previous;
} else {
     @previous = &unique(@previous);
}
```

qooob:
You could also create hidden, reflection object (select), that would be hidden and would contain non-unique data. It would have to be with some other name. To keep maintainability, the new select should have the old name/id, which is `pcmd` and old one, should get some new name.

jcameron:
Check out the patch I just added to this bug - it updates the existing hidden "previous" field to contain all commands, not just unique ones.

qooob:
Jamie, at the moment, it doesn't work right.

When the same command that is in the list been entered over again, it's not readded.

Try 

```
ls -lsa
ll
ls -lsa
```

`ls -lsa` will never be added and you end-up having only two commands.

jcameron:
Are you sure? I did a test with the latest code, and commands do appear multiple times in the `previous` field : 

```
<input class='ui_hidden' type='hidden' name="previous" id="previous" value="pwd">
<input class='ui_hidden' type='hidden' name="previous" id="previous" value="ls">
<input class='ui_hidden' type='hidden' name="previous" id="previous" value="pwd">
```

qooob:
Now I'm sure you were right. I was too tiered and stubbornly looked at select that returns  unique values.

I will make a patch for this to work correctly.

qooob:
By the way, Jamie, we need to get rid of having elements with the same id on the page, it's meant to be unique and Chromium has started to throw error about this few releases ago.

qooob:
I'm confused now. :( We had input previous all this time? I tested it and it would work just fine as it were and returned previous command properly. 

Could you just undo latest commit to `shell/index.cgi` and it will be great. I can't believe that I missed that!! What a mess.

I need a break from PC and have more fresh air.

qooob:
https://github.com/qooob/authentic-theme/commit/be5eeafde279dbe4b398a44444fd6b961abe49aa

qooob:
I'm sorry about it!

qooob:
Jamie, I have added another patch for Inbuilt Command Shell to restore support for Cloudmin's shell and also command  history delete feature.

Please fix `/server-manager/shell.cgi` to list previous commands the same way `shell` module does. Right now, the latest command entered there gets at the top of the list. You can see it by using arrows up and down.

When Cloudmin is opened the theme will use  `/server-manager/shell.cgi` and when not `/shell/index.cgi`, in case they are enabled for the user.

https://github.com/qooob/authentic-theme/commit/f0e6d22ddff5eca3d77982a0f071cf65b2aaf50d

I have been making lots of mistakes recently. I need a break.

I will do 19.07 shortly though - bug fixes release.

jcameron:
Sure, I'll do that.

PR #734, opened 2017-12-23
fix :CPU info is now displayed, even no `cpu mhz` entry is availible in `/proc/cpuinfo`

gnadelwartz:
CPU info is now displayed, even no `cpu mhz` entry is availible in `/proc/cpuinfo`

---
original linux `os_get_cpu_info()` returns info about CPU only if `$c{'cpu mhz'}` is found, this leads to not display CPU Info at all, because  on my Synology  `/proc/cpuinfo` has no info about CPU speed!

I changed this to check for `$c{model name}` which is the minimum Info needed to show CPU Type.

image

### DIFF
--- a/proc/linux-lib.pl
+++ b/proc/linux-lib.pl
@@ -308,7 +308,7 @@ if ($c{'cache size'} =~ /^(\d+)\s+KB/i) {
 elsif ($c{'cache size'} =~ /^(\d+)\s+MB/i) {
 	$c{'cache size'} = $1*1024*1024;
 	}
-if ($c{'cpu mhz'}) {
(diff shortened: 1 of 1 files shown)


gnadelwartz:
BTW: deleting the info file solves also my problem with strange HD usage displayed.

now ist correct!
image

gnadelwartz:
I like to have CPU speed present in CPU info, but if no `cpu mhz` is availible in `/proc/cpuinfo` there is nothing to show :-(

to privide at least some speed info, I added `bogomips` to `model info` if `cpu mhz` not availible.

jcameron:
Thanks!

PR #737, opened 2017-12-24
add Synology detection to os_list.txt

gnadelwartz:


gnadelwartz:
next I want to add synology config and libs. how should they are named now?
`config-syno-linux` and `*.syno-linux.pl`?

additionally, if no `*-syno-linux.lib` exist, ist the responding `*-linux.lib` is used?

jcameron:
That depends ... is Synology based on a more common linux distribution like CentOS, Fedora or Debian? Or is it completely different?

gnadelwartz:
its an embedded system , not based on other distro. generic linux worked so far.
it has most basic commands, but you can install custom packages from community.

nevertheless some commands are not availible from syno and community, e.g. vmstat, crontab, fuser ...
for some are alternatives availible, e.g. dstat, for others I need to develop a solution, e.g. crontab.

for now we can stay with the general linux libs and I try to extent them like for dstat, this may also useful for other general distros.

chris001:
* https://a68.demo.synology.com:5001/
* https://sourceforge.net/projects/dsgpl/files/?source=navbar
* https://www.synology.com/en-uk/beta/DSM62Beta

gnadelwartz:
@chris001 I know about them but I use ipkg to install prebuild community packages to synology because I want to stay as close as to the offical released system. So a regular user can is able to to the job :-)

Add `https://www.cphub.net`/ and `http://packages.synocommunity.com` to DSM packge manager.

The easy way to install webmin is the to install regular `perl` package from synologiy and webmin from `qtip` then run webmin and update to latest version :-)
image
image
image

The more felxible way is to use `Easy Bootstrap Installer`  from Qtip.  then ssh to NAS, get root (_sudo -i_)  and install needed software with `ipkg` command, e.g.

`ipkg install bzip2 git lm-sensors lsof man man-pages mc perl perl-net-ssleay sysfsutils util-linux` 

then install `webmin` from package manager

jcameron:
Ok .. you may actually in this case be better off with your initial idea of a new `syno-linux` OS code if it really is that different.

gnadelwartz:
if going down the line with syno-linux: 
- generic-linux uses the *-linux.pl lib, do I have to duplicate/require them as *-syno-linux.pl or will they be included automagically?
- do I have to provide a config-syno-linux for every module?
- are there other modules like software with hard wired dependencies with surprising results?

jcameron:
Yes, you'd need to provide a `config-syno-linux` file for every module, since presumably the paths to various config files are different on Synology Linux vs. other distros.

gnadelwartz:
OK, will prepare all config files an make a new PR

PR #743, opened 2017-12-25
enable webmin to edit user crontabs without crontab command

gnadelwarty:
embedded system like synology may have no crontab command.
add option to copy  user crontab direct to crontabs dir and send SIGHUP to crond to read config again.

image

additionally add  opportunity to create `crontabs` dir if not exist:
image


chris001:
Nice. To simplify this, you can automatically answer the question, "System has crontab command available?", you can automatically detect if the `crontab` command is available, just execute the command `which crontab` and it returns the pathname of the executable it will use.

gnadelwartz:
this magic is already available by the config files per OS, but to be flexible they provide only the default and let the superuser the choice to override. so for synology the default will be no, because there is no crontab command

cron/cron-lib.pl
```diff
@@ -538,14 +545,22 @@ sub copy_crontab
...
-	if ($single_user) {
-		&execute_command($config{'cron_user_delete_command'});
-		}
-	else {
-		&execute_command(&user_sub(
-			$config{'cron_delete_command'}, $_[0]));
+	if (!$config{'cron_crontab'}) {
+	    # we have no crontab command
+		# emulate by deleting user crontab
+		$rv = system("rm >$config{'cron_dir'}/$_[0] 2>/dev/null");
```
jcameron: Instead of this, use `&unlink_logged("$config{'cron_dir'}/$_[0]")`

cron/cron-lib.pl
```diff
@@ -1508,14 +1523,19 @@ sub check_cron_config
...
 # Check for directory
 local $fcron = ($config{'cron_dir'} =~ /\/fcron$/);
 if (!$single_user && !$config{'single_file'} &&
     !$fcron && !-d $config{'cron_dir'}) {
-	return &text('index_ecrondir', "<tt>$config{'cron_dir'}</tt>");
+	if (!$in{'create_dir'}) {
+		return &text('index_ecrondir', "<tt>$config{'cron_dir'}</tt>").
+		"<p><a href=\"index.cgi?create_dir=yes\">".&text('index_ecrondir_create' ,"<tt>$config{'cron_dir'}</tt>")."</a></p>";
+	} else {
+		system("mkdir $config{'cron_dir'}");
```
jcameron: Instead, use `&make_dir($config{'cron_dir'}, 0755);`

gnadelwartz:
changed as advised and tested

Issue #745, opened 2017-12-25
foreign_available() blocks some modules on syno-linux


gnadelwarty:
I wanted to add the ipkg packgae manager to module software, but its not listed.
when I navigate to `/software/index.cgi` or `config.cgi?software` I got the message:

> module is not installed on your system

I check ACLs module access, etc but the module is also there not listed. I tracked it down to `foreign_available()` which denies access to module, but don't understand why :-(

for now I dieabled the `foreign_available` checks on my development synology so I can navigate to module and config "by hand" and start developing.

just for couriosity I checked with authetic-theme and gray-framed theme.
Where is the hidden magic to prevent software module is not accessible and not displayed?

jcameron:
This is most likely caused by the Software Packages module not being compatible with Synology, if you've defined it as a whole new OS type in Webmin (which I guess is the case based on https://github.com/webmin/webmin/pull/737)

Fixing this is complex - does Synology have it's own custom package manager that is unlike YUM or APT?

gnadelwartz:
it has its own package management system plus a community system named ipkg.
currently I defined it as general-linux, no more as syno-linux.

I already guessed the os definition may the reason for this, but cant find where its wired.
 compatibilty of module software should be in a more visible place and  not defined by some deeply wired rules in webmin. an easy way may be an extra field in os_list.txt with the default package manager

I can understand using the software module can be dangerous, but hide it without an option to make it accessible for root  is 🤔

jcameron:
The thing about the `software` module is that it loads in code specific to the OS type, so it can't be opened up to all systems. What you probably want to do is bring back the `syno-linux` type, and in the `software` directory create a `config-syno-linux` file with `package_system=` set to a new package system type. You can see how this works here : https://github.com/webmin/webmin/blob/master/software/software-lib.pl#L11

gnadelwartz:
after I brought back syno-linux I have a similar issue with module mount.

when generic-linux was used mount can be opened and disc usage was shown in system information.
after switching to syno-linux the module is no more displayed and no disc usage is shown.

I can not find the reason why, similar to the software module access is blocked by `foreign_availible()` and friends but don't know why ...

gnadelwartz:
hm,it seems I miss something fundamental

many system modules refuses to open with this message:
![image](https://user-images.githubusercontent.com/4593242/34486083-45acaaca-efce-11e7-8a28-f0acde5160e6.png)

[...]

PR #753, opened 2018-01-03
Stay with generic-linux until module not avail solved

gnadelwarty:
to not break existing installations I will keep the config files, but change OS type back to generic-linux for next release.
meanwhile I'll investigate why so many modules are blocked by foreign_abailible() and friends.

See: https://github.com/webmin/webmin/issues/745


gnadelwarty:
BTW: If some is interested in testing/developing on for Synology NAS, since ysterday I have a working VirtualBox Image availible ...

PR #756, opened 2018-01-06
replace generic-linux with *-linux in module.info

gnadelwartz:
can confirm that syno-linux is now able to open modules from issue https://github.com/webmin/webmin/issues/745


gnadelwartz:
hm, could not relove merge conflict jamie pls do it for me ... keep the line with `syno-linux`

jcameron:
Is you aim here to create a new syno-linux OS type?

gnadelwartz:
correct, but instead of adding syno-linux to every module.info I propose to add *-linux instead of every single linux variant.

If you don't like it this way I can simply add syno-linux

jcameron:
I'd recommend adding syno-linux, since if you add *-linux it will cause the modules to appear on systems detected as generic-linux (basically, minimally supported distros).

gnadelwartz:
i ve done this only where generic was supported ... so there should support ... but yes I will do it the requested way and create a new PR

PR #759, opened 2018-01-07
Fix not to load Xinha HTML editor, as it breaks Authentic Theme https://github.com/qooob/authentic-theme/issues/977

qooob:
In the future, we should do opposite and load it only for the basic themes, leaving other themes deal with it on their own

### DIFF
--- a/mailboxes/reply_mail.cgi
+++ b/mailboxes/reply_mail.cgi
@@ -457,26 +457,31 @@ if ($in{'new'}) {
 print &ui_table_start($text{'reply_body'}, "width=100%", 2, undef,
 		      &ui_links_row(\@bodylinks));
 if ($html_edit) {
-	# Output HTML editor textarea
-	print <<EOF;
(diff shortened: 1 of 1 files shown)


jcameron:
Nice ... however, in future a better solution would be a theme-overridable function that outputs the HTML editor.

qooob:
Maybe, right now I think the best would be is just not to impose any logic on the theme. It's better to make default theme to do exactly what it needs to be done and the theme could decide what and how to load.

However, as long as it;s not hard-code, all would be doable.

tiagoadp:
The problem still happens with 19.08, just tested right now, attached screenshots:

chrome_2018-01-09_10-42-25

chrome_2018-01-09_10-43-28

Goin back to text compose as it is working for now...

qooob:
You need the patch from Webmin:
https://github.com/webmin/webmin/commit/f68039d36030de2452c2f5e23822afc73c43f2f9

PR #760, opened 2018-01-07
return of syno-linux

gnadelwarty:
I checked all modules where genrric-linux is listed and take a hybrid approach as follow:

- when a config file for all 19 linux variants is availible I use `os_support= *.linux`
- when only a few linux variants are supported I check if `syno-linux` is working and add syno-linux
- in both cases I provide a working `config-syno-linux`

In some modules I have to remove  `config-syno-linux` because I created it with an earlyer patch, but it is not needed.

most modules now working with syno-linux if the responsible software package is availible, all other linux distros should work as before.

over time I may modify more modules to work with syno linux.
even I'm a long time linux and perl user I'm very impressed what's possible with webmin :-)


gnadelwarty:
ups, looks like I used PAM module.info for QUOTA module.info by acident, I'll fix this ...

jcameron:
Thanks for the patch!

PR #761, opened 2018-01-08
Fix open theme's link in new tab

qooob:
This one works correctly.

Just copy past any URL and hit enter. As long as URL has `xnavigation=1` query it will work.

In the future, I will try to modify clipboard when user copies the link using right click, so it would also be pasted with `xnavigation=1` trigger.

Tell me if you like it and I could skip doing lots of reworking.

### DIFF
--- a/web-lib-funcs.pl
+++ b/web-lib-funcs.pl
@@ -4702,9 +4702,28 @@ if ($module_name && !$main::no_acl_check &&
 	}
 
 # Check for trigger URL to simply redirect to root: required for Authentic Theme 19.00+
-if ($ENV{'REQUEST_URI'} =~ /xnavigation=1/) {
-	redirect("/");
+if ($current_theme =~ /authentic-theme/ && $ENV{'REQUEST_URI'} =~ /xnavigation=1/) {
+
+  # Redirect user to the entered URL upon next page load; store URL into temporary file
+  if ($main::session_id && $remote_user) {
+      my $xnav = "xnavigation=1";
+
+      my $url = "$gconfig{'webprefix'}$ENV{'REQUEST_URI'}";
+      $url =~ s/[?|&]$xnav//g;
+      $url =~ s/[^\p{L}\p{N},;:.%&#=_@\+\?\-\/]//g;
+
+      my $tmp  = 'tmp';
+      my $salt = encode_base64($main::session_id);
+      $salt =~ tr/A-Za-z0-9//cd;
+
+      my %var;
+      my $key = 'goto';
+      $var{$key} = $url;
+			write_file(tempname('.theme_' . $salt . '_' . get_product_name() . '_' . $key . '_' . $remote_user), \%var);
+  		}
+  redirect("/");
 	}
+
 
 # Check the Referer: header for nasty redirects
 my @referers = split(/\s+/, $gconfig{'referers'});
--- a/web-lib-funcs.pl
+++ b/web-lib-funcs.pl
@@ -4701,30 +4701,6 @@ if ($module_name && !$main::no_acl_check &&
 	$main::no_acl_check++;
 	}
 
-# Check for trigger URL to simply redirect to root: required for Authentic Theme 19.00+
-if ($current_theme =~ /authentic-theme/ && $ENV{'REQUEST_URI'} =~ /xnavigation=1/) {
-
-  # Redirect user to the entered URL upon next page load; store URL into temporary file
-  if ($main::session_id && $remote_user) {
-      my $xnav = "xnavigation=1";
-
-      my $url = "$gconfig{'webprefix'}$ENV{'REQUEST_URI'}";
-      $url =~ s/[?|&]$xnav//g;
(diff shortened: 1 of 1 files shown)


jcameron:
Thanks, this is great! What is `xnavigation=1` normally used for by the theme though?

qooob:
`xnavigation=1` is merely the trigger that tells Webmin or the theme to reload to root and start as theme and not the module.

qooob:
By the way, using this patch, we can control later, by extending it, which pages we want to exclude from opening in the new tab directly, for example, all pages that use buffered header and most other that use post action, like many in MySQL/PostgreSQL modules. Rebuilding theme as we discussed earlier, where left menu integrated in the module itself wouldn't provide such opportunity on the init level.

jcameron:
So with this change, would a request for a URL like `/proc/run.cgi?cmd=pwd&user=root&xnavigation=1` trigger a redirect (before referrer checking) which then causes the same URL to be fetched in the right frame by the theme's index page?

Or are all the URL parameters stripped off in this case? I'm worried about how this could potentially get around referrer checks.

qooob:
The theme will try to open valid URL.

No, parameters will not be stripped off. You just give it a try on your local machine. It will try to open URL as it is.

What theme exactly should worry about when querying the page for module? In case URL is broken, it will not be opened, and that's it.

qooob:
Referre is always overwritten by browser.

http://www.psychology-age.ru/2017/11/blog-post_642.html

Or your concern not about browser's requests?

qooob:
And yes, anytime the theme is set to Authentic and URL contains `xnavigation=1` it will be redirected. Before that URL will be stored.

In your example, the code will run but it's the same as opening module UI and doing it manually. 

Where is concern here, I don't understand?

qooob:
You have this in your referre check

```Perl
elsif ($referer_site eq $http_host &&
       (!$referer_port || !$http_port || $referer_port == $http_port)) {
	# Link came from this website
	$trust = 1;
	}
```

We only store relative URL in 

```Perl
my $url = "$gconfig{'webprefix'}$ENV{'REQUEST_URI'}";
```

Besides, Javascript will not open link from other resource.

jcameron:
My only concern is if a Webmin user is tricked into clicking on a URL (or opens a page with an image from a URL) that performs some action on his server. Normally referrer checking prevents this, but if the URL requested was being saved before the referer check with all params and then re-opened, that could be an issue, 

However, if params are stripped off, hopefully we are safe. Although there are some URLs in Webmin like /init/reboot.cgi that could have immediate impact.

qooob:
Yeah, it's stripped off.

Let's create a function, which will check for URL containing string to be excluded?

I actually have hard time understanding what could be the problem in reality? In case someone could query your server with your session id - is already lost situation.

Other than that, can you describe the real world situation, where one could query `/init/reboot.cgi` for the user with the same session id? Besides, we have great new feature called _ Limit sessions to the same client IP address_, which should be on be default, I believe.

jcameron:
The case I'm thinking of is if you're logged into Webmin, and in the same browser session click on a link on a malicious website like : 

<a href=https://localhost:10000/init/reboot.cgi>click me</a>

It's important that referrer checking prevent this from doing anything.

qooob:
Even though it would be very difficult to perpetrate this kind of attack, now I see what's it all about.

Then, all we need to do is to make sure that the referre is not set at all (on pasting URL to the new tab and hitting enter) or/and referre is our actual application (opened link using context menu).

I'm making another commit to respect the checks. I will also cut the session name to make it more secure.

qooob:
Check the commit above. This should be fairly safe. As far as I'm aware, browser will always set/overwrite referer header.

qooob:
You would also need this latest commit to make patch above work: https://github.com/qooob/authentic-theme/commit/b85d67d56260d8bb285e7a2a93d8620dc8ac4ea5

qooob:
I suppose the link from an email will open as "safe" and pass referre test. 

I think, we should count on the filter (regex), making sure something like `rm -rf /` or any other command, queries containing spaces and other spec chars would not pass it. I would also built a function that would have a list of pages to exclude, like any for reboot and shut down, deleting anything or changing anything.

jcameron:
Thanks, that looks safer! I don't think it's possible to detect all possible dangerous commands via regexps ... instead, the only safe defense in Webmin is referer checking, to ensure that URLs which take some action can only come from inside Webmin itself.

qooob:
Isn't it true that most commands would need a space, and on opposite URLs not.

Do you mind if in the future I will extend it with list of URLs that should be ignored for storing?

jcameron:
I don't think that's a good idea, since the list of possible URLs is infinite :-)  

The only real protection is to ensure that referrer checking is working properly.

qooob:
I didn't check, but I think it will pass referre check with current code and write URL for opening.

Okay, I will do more checking on theme side. Also make sure that reboot and other will not be opened and make sure that cmd=cmd with params will fail to work.

PR #768, opened 2018-01-13
default values if only firewall-lib is included

gnadelwartz:
provide default values if only firewall-lib is included, e.g. foreign_require(firewall, firewall-lib.pl) calls


firewall/firewall-lib.pl
```diff
@@ -9,6 +9,9 @@
 
 $config{'perpage'} ||= 50;	# a value of 0 can cause problems
 
+# provide default values if only firewall-lib is included, e.g. foreign_require(firewall, firewall-lib.pl) calls
+set_ipvx_version(get_ipvx_version);
```
jcameron: Shouldn't this be get_ipvx_version() to make it a function call?

gnadelwartz:
you are rigth, fixed ...

PR #769, opened 2018-01-13
first working version of Synology Package manager IPKG

gnadelwartz:
tis is the first working version of Synology Package manager IPKG
there may be mistakes but its working  :-)

who ever created the module software must be a genius or I'm stupid, nefertheless I figured out how to get a working ipkg-lib.pl.


jcameron:
Thanks for the patch!

Issue #770, opened 2018-01-14
proposal: move module desc to lang files ...


gnadelwartz:
move module desc to lang files and add them to module.info when creating release tarballs.

having different encodings and the mix of unicode and ASCII in `module.info` often leads to confusion by humans and editors. moving the desccription strings to the corresponding lang files should solve this problems and translators have no more to deal with module.info.

after creating the UTF-8 language files for a release, we add the descriptions to module.info for release again, basicly something like this, but more sohpisticated:

 `grep "^desc|^longdesc" lang/* >> module.info`

As a bonus we will have all descritpions in UTF-8 encodings availible in module.info.

jcameron:
I'm actually thinking of going the other way - update all modules' index.cgi pages to not use `index_title` from the lang file, and instead use `$module_info{'desc'}`

gnadelwartz:
that will fit, because after installation/update the desc is in module.info.
I want only to avoid mixing encoding while developing.

gnadelwartz:
here is a script moving all desc from module.info to lang files on a fresh  clone from repo.

- skip move if no lang dir exsit
- keep english desc in module.info as fallback
- add english desc to lang file if no desc_lang exist

if not argument is given it processes only apache as testcaae
for processing all modules run `./move2lang.sh */module.info`

whats now missing is to extend setpup.sh to copy vaules bak while setup

``` .sh
#!/bin/bash
# move2lang.sh
# move descriptions out of module,info
FILES=apache/module.info
[ "$1" != "" ] && FILES="$*"

for file in $FILES
do
    echo "processing $file ..."
    # get dir
    dir=`dirname $file`
    [ ! -d "$dir/lang" ] && echo "skipping $dir: no lang dir" && continue
    # move to temp file
    sed '/desc_.*=/d' $file >$file.tmp
    grep -a -e '^desc' -e '^longdesc' $file | sort >$file.desc
    mv $file.tmp $file

    # move desc to lang files
    for line in `ls $dir/lang`
    do
        lang="_$line"
        [ "$lang" = "_en" ] && lang=""
        sed -n  "s/^desc$lang=/desc=/p" $file.desc >$dir/lang/$line.tmp
        sed -n  "s/^longdesc$lang=/longdesc=/p" $file.desc >>$dir/lang/$line.tmp
        if [ ! -s "$dir/lang/$line.tmp" ]
        then
            sed -n  "/^desc=/p" $file.desc >$dir/lang/$line.tmp
[...]

PR #771, opened 2018-01-14
provide `.gitattributes` file with file encodings for webmin

gnadelwartz:
in addition to https://github.com/webmin/webmin/issues/770 it may helpful to provide git some knowledge about file encoding in webmin.

I set default encoding to iso-8859-1 and force module.info also to iso-8859-1.
the encoding of lang files is extracted from lang_files.txt


jcameron:
Nice, I didn't know that could be done!

PR #773, opened 2018-01-15
Add support for forcing encoding using file comment '-*- coding: encoding -*-'

qooob:
https://github.com/qooob/authentic-theme/issues/987

### DIFF
--- a/filemin/edit_file.cgi
+++ b/filemin/edit_file.cgi
@@ -5,24 +5,29 @@ require './filemin-lib.pl';
 
 get_paths();
 
-my $file = &simplify_path($cwd.'/'.$in{'file'});
+my $file = &simplify_path($cwd . '/' . $in{'file'});
 &check_allowed_path($file);
 my $data = &read_file_contents($file);
 
 my $encoding_name;
 eval "use Encode::Detect::Detector;";
 if (!$@) {
-	$encoding_name = Encode::Detect::Detector::detect($data);
-	}
-if ( lc( get_charset() ) eq "utf-8" && ( $encoding_name && lc($encoding_name) ne "utf-8" ) ) {
+    $encoding_name = Encode::Detect::Detector::detect($data);
+}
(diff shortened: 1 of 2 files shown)


gnadelwartz:
thanks Ilia for your fast coding :-)

qooob:
You're welcome.

In case you want `vim: encoding ..` to work, you would need to modify regex.

gnadelwartz:
if you used to regex from my link to stackoverflow, it should work already:

`/coding[=:]\s*([-\w.]+)/`

qooob:
It will grab any `coding:` and `coding=` - not good. Needs to be more safe!

gnadelwartz:
it's save when only checking the first two lines and it's a comment. currently you are checking all lines.

`/^\s*#.* coding[=:]\s*([-\w.]+)/`

qooob:
If so, yes. You can go ahead and modify my commit it make your own one instead.

jcameron:
Thanks for the patch! Shall I accept this into Webmin, or is there more work still being done?

qooob:
Yes, and later we'll add it.

qooob:
Okay, Jamie, latest commit is fine. 

It will only look for 3 first lines and also support VIM, because I know some guys with beards that told me once or twice that they use VIM. 😉

qooob:
New pull request but, Jamie, for some reason it was added to this one. Probably because it's opened by me?

I added [hardened filter for checking URL](https://github.com/webmin/webmin/pull/773/commits/d13795190b19d3ffc1138bb4913bd3cad29e250b) before saving. Now, even if opened from desktop mail client, it will not be possible to run a command containing space in it, so running `cmd=ls -lsaZ` will result in the same command but with striped spaces.

Take a look.

gnadelwartz:
hey! I have no beard! even I'm using vi ...

qooob:
There are others with beard! 🤣

gnadelwartz:
this one @swelljoe ?

qooob:
Ha-ha, yeah, that one.. particularly nice guy! 😉

PR #781, opened 2018-01-18
Harden checks for redirect and keeping URI

qooob:


jcameron:
What kind of URLs is this checking for?

qooob:
It makes sure that the request is not XHR and the trigger is the same `xnavigation=1`.

PR #783, opened 2018-01-22
Don't filter out URI for trusted source

qooob:
It's important for upcoming 19.09, where right-click and open in new tab, will work correctly with the referre set.

### DIFF
--- a/web-lib-funcs.pl
+++ b/web-lib-funcs.pl
@@ -4791,6 +4791,7 @@ else {
 if ($ENV{'HTTP_X_REQUESTED_WITH'} ne "XMLHttpRequest" &&
     $ENV{'REQUEST_URI'} !~ /xhr/  &&
     $ENV{'REQUEST_URI'} !~ /pjax/ &&
+    $ENV{'REQUEST_URI'} !~ /link.cgi\/\d+/ &&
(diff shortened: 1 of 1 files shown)


qooob:
The latest commit of the theme has support for this patch.

Holding `Ctrl` or `Shift` while making a click will also work, just as context-menu/open in new tab.

qooob:
Does it work safe with you?

jcameron:
This looks reasonable - but how are you defining a "dangerous" URL? I don't quite follow what kinds of URLs the regexp is looking for.

qooob:
Regex is looking for spaces primarily. Fir example, opening targeted link from email, will not result into `cmd=rm -rf /`.

qooob:
Oh my, oh my! Just found a bug that breaks Server Index, as it should skip as well:

```Perl
$ENV{'REQUEST_URI'} !~ /link.cgi\/\d+/ &&
```

qooob:
Please test it with the latest commit:
https://github.com/qooob/authentic-theme/commit/dc63db6873b4fa0686011cf3046e7c8e42a61dcf

qooob:
Ref: https://github.com/webmin/webmin/pull/796

jcameron:
Thanks for that fix!

PR #788, opened 2018-01-27
some minor syno-linux cleanup

gnadelwartz:
be not afraid about the much files, I removed all not needed config-syno-linux which I copied over from my installation stupidly while my first attemp to create syno-linux.


gnadelwartz:
made a fresh install on Synology NAS and everthing works now as it should. PLS merge

PR #789, opened 2018-01-28
activate update manager, package-updates, passwd and useradmin on synology

gnadelwartz:
image


qooob:
@gnadelwartz Kay, what browser is that?

gnadelwartz:
vivaldi aka chrome

jcameron:
This patch appears to have changed a bunch of other lines in the module.info files - is that expected?

gnadelwartz:
ah, l have an other idea, all lines with special characters are affected, so its may be because i forced module.info to iso-8859-1. so github displays the lines because of the forced charset ...

jcameron:
Is all you're really trying to do is add syno-linux to the os_support line of the useradmin/module.info file?

gnadelwartz:
yes, in the same hibrid approach lve done it in the "big bang" PR:

- if all 19 linux variants including generic-linux are listed use *-linux
- in all other cases add only syno-linux

beside this I add the config option ipkg for update system

gnadelwartz:
ok, i make a new PR for package-update

PR #790, opened 2018-01-29
remove wrong </a> in tree.cgi

gnadelwartz:
I'm pretty shure there is no opening `<a>`  ...


gnadelwartz:
@iliaross find an other issue in tree.cgi, I'll provide a combined patch

PR #794, opened 2018-01-31
allow "-" in version string

gnadelwartz:
this should fix the problem of update-from-repo


gnadelwartz:
I found a workaround in update-from-repo, await a reworked version ...

Issue #791, opened 2018-01-30
Latest development version is broken 1.873-0128.1555


qooob:
Hi,

After running `update-from-repo.sh` and waiting for update to finish - there is no left menu printed anymore in Webmin.

Clicking around also shows that things are broken.

![screenshot from 2018-01-30 12-24-50](https://user-images.githubusercontent.com/4426533/35558333-d51dbbae-05b8-11e8-9a44-c77ab2353b88.png)

jcameron:
Is this when running the latest checked-in code? I'm not seeing this on my test systems..

qooob:
Yes, I just ran update script.

It used to work before but yesterday, resulted into broken left menu.

I'll try again today.

gnadelwartz:
hm, it may also a problem of my update script.

l've had this issue also after done multiple updates from script.
no clue why this happens, my solution  was to install the latest dev and then update again ...

gnadelwartz:
I have an Idea where to look, a week ago or so there was the problem that setup.sh can't find mktemp.pl.

I'll have a look ...

gnadelwartz:
yes its my script or more precise makedist.pl doest like my ne version scheme 1.873-xxx and exits

gnadelwartz:
https://github.com/webmin/webmin/pull/794 should fix the problem ...

qooob:
I just cloned the repo and copy/paste it over my existing version, replaced _shebang_ - and it just works, so probably it was your script.

gnadelwartz:
@iliaross you can close this, its adressed with reworked update-from-repo script ...

qooob:
Thanks!

gnadelwartz:
wow! less than 15 seconds, new record!

qooob:
Kay, have a look:


```Bash
root@debug-debian:/usr/share/webmin# ./update-from-repo.sh 
search minserv.conf ... 
Warning: update from non release repository https://github.com/webmin/webmin may break your installation!
Would you like to update Webmin from https://github.com/webmin/webmin [y/N] y
Pulling in latest changes for Webmin  (https://github.com/webmin/webmin) ...
start FULL update for webmin ...
Adding top-level files
Adding module cron
Adding module dfsadmin
Adding module exports
Adding module inetd
Adding module init
Adding module mount
Adding module samba
Adding module useradmin
Adding module fdisk
Adding module format
Adding module proc
Adding module webmin
Adding module quota
Adding module software
Adding module pap
[...]

PR #795, opened 2018-01-31
refactor update-from-repo, also fix the version "bug"

gnadelwartz:
update-from-repo has been grown over time and was hard to maintain, so I refactor it a bit:

- move checks to the begining to avoid if-else hell copied over from qooobs update script
- cleanup warnings and code
- remove -lang option and code, its no more usefull (at least for me)
- check the exit status of makedist.pl to avoid unnoticed problems like qooob reported in https://github.com/webmin/webmin/issues/791
- workaround for rejected version sheme, no need to change  makedist.pl

@qooob may be I reconsider your feature request to pull a specific file or dir only :-)


gnadelwartz:
@jcameron forked refatoring branch with IPKG changes, so pls accept https://github.com/webmin/webmin/pull/793 first ...

PR #799, opened 2018-02-01
fix error detection for makedist and no color if writing to pipe or -…

gnadelwartz:
@qooob  reported wrong abort even if makedist pass ... fixed


gnadelwartz:
@iliaross implemented `-force` option for unattended update incl. no colors and `yes | update-theme` ...

qooob:
Thanks.

PR #801, opened 2018-02-02
more IPKG Treeview improvements (done)

gnadelwartz:
- buttons on top and botton
- filter shown packages
- show if a package is installed 

image


gnadelwartz:
everthing on my list is done for now, have a nice weekend!

PR #802, opened 2018-02-06
add urlize to software/open.cgi close.cgi

gnadelwartz:


gnadelwartz:
@iliaross added support for moving  to #someting in authentic theme and noticed that if you klick in software/tree.cgi of a categire with a blank in the categoriy name., the redirect  is not working correct.

to fix this issue I add &urllize() to the redirect of close.cgi/open.cgi whit preserves the blank.
additionally I applied this fix to ipkg-open.cgi and ipkg-close.cgi .

to be honest, I add also the close all and open all buttons on top of tree.cgi

PR #805, opened 2018-02-11
Change webmin/usermin detection in update.sh

gnadelwartz:
Until now I check for dirnames webmin and usermin, but the name of the installtion dir can be anything different.

So I changed the detection of webmin / usermin  by checking for existence of typical files:

- setup.sh and miniserv.pl not present => no webmin/usermin installation
- existence of usermin-init and uconfig.cgi => usermin detected

### DIFF
--- a/update-from-repo.sh
+++ b/update-from-repo.sh
@@ -8,22 +8,6 @@
 # Kay Marquardt, kay@rrr.de, https://github.com/gandelwartz
 #############################################################################
 
-# Get webmin/usermin dir based on script's location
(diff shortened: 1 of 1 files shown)


gnadelwartz:
@iliaross  may be you what to change it simmilar in your update script also ...

qooob:
@gnadelwartz It's not needed in my setup. Each script runs individually and doesn't care about anything in this regard.

gnadelwartz:
you use also $PROD based on dirname to display messages:

PROD=${DIR##*/}

qooob:
I see what you mean. Thanks. I didn't know that the path could be `/webmin/target/bin` - this is very odd.

Issue #804, opened 2018-02-09
Install Updates Button needs better placement.


f3bruary:
I got Mono installed on my machine and each time there's an update, it's for +150 packages at once. When there are updates, I have to scroll down all the packages and click Update Selected Packages. Then I have to scroll all the way up again to click Install.

Surely this can be improved?

qooob:
It's possible to improve. I will be able to add pagination to the table, I think.  Is it Software Update Packages module?

By the way, why not just hitting end button on the keyboard to be to the bottom of the page?

gnadelwartz:
I provided ad PR to add the buttons also on the top of the list ...

gnadelwartz:
@f3bruary  your request will included in the next release.

if you want to test it now, go to your webmin installtion dir and run `./update-from-repo.sh -force`

gnadelwartz:
@jcameron can be closed

PR #806, opened 2018-02-11
place buttons also on top of update-packages list #804

gnadelwartz:
see issue https://github.com/webmin/webmin/issues/804

### DIFF
--- a/package-updates/index.cgi
+++ b/package-updates/index.cgi
@@ -135,8 +135,8 @@ else {
 		print &text('index_count', scalar(@rows)),"<br>\n";
 		print &ui_form_start("update.cgi", "post");
 		print &ui_submit($in{'mode'} eq 'new' ? $text{'index_install'}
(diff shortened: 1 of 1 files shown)


qooob:
It's not working as not submitting anything.

gnadelwartz:
OK, I'll have a look

gnadelwartz:
now its working for me, give mit a try ...

qooob:
Now we have two `id="refresh"` and two `id="ok"` on the page. This is invalid. We don't need to print `id` at all.

gnadelwartz:
this is from ui_lib ... let me think about it ...

gnadelwartz:
I add `_top` to top buttons names

qooob:
I've also added small margin for the new buttons to look better:
https://github.com/qooob/authentic-theme/commit/e4085ed4ed01eaf638852cf04abed224c542a363

jcameron:
Thanks for the patch!

PR #809, opened 2018-02-11
process also dirs given by -file

gnadelwartz:
now you can also provide dir name(s) with file, so @qooob 's example will work as expected

`-file cpan/*` will work recursive for dirs, btw `-file cpan` does the same but faster
`-file module/lang` pull all lang files of a module
`-file */lang` pull lang files for all existing */lang


qooob:
Alright, this is great, thanks.

gnadelwartz:
one remaining problem is the shebang in perl and cgi scripts, i'll try to implement this when lm back home

PR #810, opened 2018-02-12
 fix insert perl path also with -file

gnadelwartz:
API and behavior  is now considered stable ... 
I'll try to make no backward incompatible changes and no more break functionality :-)


jcameron:
Is this ready for merging?

gnadelwartz:
IMHO yes 😁

jcameron:
Thanks!

PR #815, opened 2018-02-16
fix missing -e for echo

gnadelwartz:
small fix for update-from-repo ...


swelljoe:
echo on some POSIX shells does not have an `-e` flag. This may fail or behave badly if bash isn't the shell being used to run it:

https://wiki.ubuntu.com/DashAsBinSh#echo

printf is more predictable for this use case.

gnadelwartz:
hi joe, you are always a good source of knowlede, thanks for the hint.

In the update-scripts we force use of bash with shebang '#!/bin/bash', so hopefully only bash/posix compatible shells should run them.

PR #823, opened 2018-02-21
using amavisd instead of procmail as spam filter

gnadelwartz:
I use amavisd and not procmail as spamfilter, even there is an option to disable the warning it was not clear if its onyl the warning or more. 

Additionally the spam score is ignored if you use amavisd, as filter. When this hit ms the first time,  it takes hours of resaerch until I noticed that this  setting was overwritten internally by amavisd.

To improve this for other amvisd (other filter) users I've don the following:

- make it claearer you should disable warning if you not use procmail as spam filter
- add a warning to spam score that it is ignored with amavisd

lets see if if I make more improvements like editing amvisd spamconfig, but for now I'm done ...


gnadelwartz:
Seems the translations are not done by Raymond Vetter ...

gnadelwartz:
this can be merged also for release

jcameron:
Great!

PR #842, opened 2018-03-04
language code for Latvian

Kleofass:


jcameron:
It seems like this patch re-ordered the whole file ... is that expected?

Kleofass:
A-Z order - just about a clean design. If it affects a function, then we can leave the old content.

qooob:
It's never good idea to sort things like that as it gets very difficult to watch changes.

Kleofass:
Ok. I will delete this request.

PR #853, opened 2018-03-10
fix -repo option, add check for curl existing

gnadelwartz:
What's in this PR:

- fix for -repo option
- test if curl is availible if needed by option (requested by @qooob )
- additonal check for webmin installation: authentic_theme has to be availible
- authentic does not support alternative repo, so remove code to check for


gnadelwartz:
@jcameron pls merge ...

Issue #833, opened 2018-02-26
PCI compliancy failures cgi abuse vulnerability and others in session_login.cgi 


raidensnake:
I'm running into a pci compliancy issue as it's failing cause of session_login.cgi which affects both webmin and the authentic theme with the same file names.

These are the details from a comodo pci compliancy report.

> Security Hole found on port/service "10000 / tcp / www"
> 
> Status: Automatic Failure as listed by the PCI SSC (This must be resolved for your device)
> 
> Plugin "CGI Generic SQL Injection (blind)"
> Category: "CGI abuses"
> Priority: "Urgent"
> 
> Synopsis: A CGI application hosted on the remote web server is potentially prone to SQL injection attack.
> 
> 
> Description: 
> By sending specially crafted parameters to one or more CGI scripts hosted on the remote web server, Nessus was able to get a very different response, which suggests that it may have been able to modify the behavior of the application and directly access the underlying database. An attacker may be able to exploit this issue to bypass authentication, read confidential data, modify the remote database, or even take control of the remote operating system.
> 
> 
> Risk factor | HIGH / CVSS BASE SCORE :7.5 CVSS2#AV:N/AC:L/Au:N/C:P/I:P/A:P
> Pluginoutput : Using the GET HTTP method, Nessus found that :
> + The following resources may be vulnerable to blind SQL injection :
> + The 'user' parameter of the /session_login.cgi CGI :
/session_login.cgi?pass=&save=1&user=zz&save=1&user=yy

```
-------- output --------
<body class="session_login" data-style="">
<div class="container session_login" data-dcontainer="1">

<form method="post" target="_top" action="/session_login.cgi" clas [...]
<i class="wbm-webmin"></i><h2 class="form-signin-heading">
-------- vs --------
<body class="session_login" data-style="">
<div class="container session_login" data-dcontainer="1">
<div class="alert alert-warning">
<strong><i class ="fa fa-exclamation-triangle"></i> Warning!</stro [...]
<span>Login failed. Please try again.</span>
------------------------

```
> Solution: Modify the affected CGI scripts so that they properly escape arguments.

swelljoe:
This isn't a sensible error. There is no SQL involved in anything that touches any of that code...it's impossible to have an SQL injection in code that doesn't talk to any SQL servers.

And, the behavior that it is saying is problematic is a login failed error, which is exactly what is supposed to happen when given bogus data.

You should notify your PCI compliance testing folks that this test is bogus.

[...]

PR #862, opened 2018-03-12
fix autodetecting crontab command 

gnadelwartz:
whats in this PR:

- `$config{'cron_user_edit_command'}` does not exist, checking for `has_command('crontab');` is enough
- instead change logic for complaining about missing config for  crontab commands:
   - if crontab exist, commands for list and edit must be given
   - if crontab not exist values for list and edit are ignored
- remove no more needed option "is crontab availible"
- fix permissions of manually created user crontab

### DIFF
--- a/cluster-copy/module.info.de
+++ b/cluster-copy/module.info.de
@@ -1,2 +1,2 @@
 desc_de=Cluster - Dateien kopieren
-longdesc_de=Plane den Dateireansfer von diesem zu anderen Servern in einem Webmin-Cluster.
+longdesc_de=Plane den Dateiransfer von diesem zu anderen Servern in einem Webmin-Cluster.
--- a/cron/config.info
+++ b/cron/config.info
@@ -11,7 +11,6 @@ hourly_only=Only allow jobs to be at most hourly?,1,0-No,1-Yes
 add_file=Add new jobs to file,3,User's regular crontab file
 line2=System configuration,11
 cron_dir=Crontab Directory,0
-cron_crontab=System has <tt>crontab</tt> command availible?,1,0-No&#44; webmin processes user files itself,1-Yes&#44; use following crontab commands for users:
 cron_get_command=Command to read a user's cron job,0
 cron_edit_command=Command to edit a user's cron job,0
 cron_copy_command=Command to accept a user's cron job on stdin,0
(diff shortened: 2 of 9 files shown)


jcameron:
Is this building on top of my patch https://github.com/webmin/webmin/commit/938db3252b7d3a78738b1469f9a46df916e7cedb , or a replacement for it?

gnadelwartz:
on top, the only thing I canged is to move the check for edit command from has-crontab-command to check if cron is availible because if crontab is availible it was not complaining about missing edit command

gnadelwartz:
OK, its easier to explain on my office desktop than on tablet :-) , so here is some more detail why I removed you check for edit command from:

```
-my $cmd = $config{'cron_user_edit_command'} || "crontab";
-($cmd) = &split_quoted_string($cmd);
-return &has_command($cmd);

```
- cron_user_edit_command otion should be cron_user_edit_command

jcameron:
Do you mean that there are systems that have the `crontab` command, but it doesn't actually work?

gnadelwartz:
did it read like this? this was not the intention.

I trust crontab more todo it the right way then doing it ourself, so I want to avoid a silent fallback to manual edit if crontab exist on the system.

if I understand  ` $config{'cron_edit_command'} || "crontab";` correctly it check for crontab existence only if edit command is empty. 

the question is: what should happen if user enters a not existing command but crontab exist:

- only test if entered command exist and use manual edit?
- warn the user he entered a not existing command, even crontab exist on his system?

I'm fine with both.

gnadelwartz:
ups, mixed up branches :-/

nevertheless, I you want to keep it like your proposal, decline this PR.

If you think it may worth to do it this way i'll provide an new pill request:

``` perl
 sub has_crontab_cmd
 {
my $cmd = $config{'cron_user_edit_command'};
$cmd =~ s/^su.*-c\s+//;
($cmd) = &split_quoted_string($cmd);
return &has_command($cmd) || &has_command("crontab");
 }
```

jcameron:
Ok, does this commit handle that case OK for you : https://github.com/webmin/webmin/commit/79d56166786b2ea5729a265ed41a6b65ce74413d

gnadelwartz:
great!  now only chmod 600 for file is missing line 601 from here https://github.com/webmin/webmin/issues/833

jcameron:
Fixed with this commit : https://github.com/webmin/webmin/commit/13889515ea414211ebd8b86e0d2e405adea7d6b2

PR #869, opened 2018-03-16
Statisfy PCI scanner with error handler

gnadelwartz:
merge if you think this is the right path to go ...

### DIFF
--- a/error403.cgi
+++ b/error403.cgi
@@ -1,0 +1,14 @@
+#!/usr/local/bin/perl
+
+BEGIN { push(@INC, ".."); };
+use WebminCore;
+
+&init_config();
+&ReadParse();
+
+$msg="error_$in{'code'}";
+&ui_print_header(undef, "$text{'error'} $in{'code'}: $text{$msg}", "", undef, undef, 1, 1);
+print ui_alert_box("<br>$in{'message'}<p>$in{'body'}", "warn");
+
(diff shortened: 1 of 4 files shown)


gnadelwartz:
how will we proceed with this topic?

jcameron:
How does this patch resolve the problem? I'm not seeing how that custom error handler is used?

gnadelwartz:
as stated in the issue I modify the http return code it custom handlers are used and a simple custom handler.

curently the handler has to be activated by hand, so what you possibly missing is UI support for activation?

cooperdjm:
So sorry for my ignorance, I am also experiencing the false flag PCI failure. Can someone point me in the direction of how to activate the handler by hand?  I am running Webmin 1.941

cooperdjm:
> So sorry for my ignorance, I am also experiencing the false flag PCI failure. Can someone point me in the direction of how to activate the handler by hand? I am running Webmin 1.941

I think I figured it out, well no, let's be honest I read some more and saw there are instructions to modify the .pl file.

swelljoe:
Since there are merge conflicts and no follow up in five+ years, and I don't understand why PCI would expect an error page to have status 200 (OK) and how that could possibly be a security improvement...I'm gonna close this. If it's still an issue, we'll need a new PR or this one updated.

PR #878, opened 2018-03-20
shorter german menu entry for "Language and Theme"

gnadelwartz:
The actual menu entry need two lines, this is shorter and the same.
Changed also a reference where the menu entry is mentioned.


gnadelwartz:
same for save config files

PR #881, opened 2018-03-21
Update postfix-lib.pl

R3freshReload:
Postfix version >= 2.2 notation gets ignored if path starts with . or /

Also: Sorry for creating 2 pull requests for the same file. I only discovered this a bit later


R3freshReload:
I will add this to the other pull request

PR #882, opened 2018-03-23
Rename ja_JP.euc to ja_JP.UTF-8

kosugikun:


gnadelwartz:
in `lang.txt` japanense translation is defined as charset `EUC-JP`. so ja_JP.UTF-8 may not used ...

`lang=ja_JP.euc,charset=EUC-JP`

If japanese files are already UTF-8 it must also changed in `lang.txt`

@jcameron ?

jcameron:
We create a separate ja_JP.UTF-8 file based on the ja_JP.euc file, this is done at packaging time so the files aren't in the repository.

PR #885, opened 2018-03-28
fix (ipkg) tree view

gnadelwartz:
What's in this PR:

- fix upper filter form in `tree.cgi`, pointed to ipkg-tree.cgi ...
- fix `ipkg-tree.cgi` to use correct lang strings, filter buttons displayed no text
- optimisation while generating categories in `ipkg-lib.pl`


jcameron:
Thanks!

PR #887, opened 2018-03-29
 provide setup.bat for windows installation

gnadelwartz:
Installation instructions for Windows and Installer seems to be old and (possibly) no more working.
see https://sourceforge.net/p/webadmin/discussion/55378/thread/c86e083e/?limit=25

To understand what's needed to install Webmin on Windows I went to install it manually.
The result of this journey is a` setup.bat` script for windows  to guide the user and install  webmin  in one rush if all prerequsites are fulfilled.

what's in this PR:

- windows batch file `setup.bat` to make installation on windows easier
   checks installation of  `perl`, `process.exe`, `sc.exe` and run `setup.pl` with parameters needed for windows setup
- fix to `setup.pl` to include `web-lib-funcs.pl` from source dir because it will not exist in wadir before installation.   add parameter `/Q` to windows `xcopy` to not display thousands of copied files
- add `setup.bat` to `makedist.pl` to include it in new releases

whats' not in this PR:
- fix to run as Windows Service:
  I can run webmin manually from start script, but Webmin seems not be installed from`setup.pl `as a windows service

---
Prerequisites for Webmin on Windows:
(all of them will be checked from batch file)

- Download webmin source an unpack in a temporary location
https://sourceforge.net/projects/webadmin/
- Download and installl Perl
https://www.activestate.com/activeperl/
- Download Process.exe and place it in C:\Windows
http://retired.beyondlogic.org/solutions/processutil/processutil.htm
- Download and Install Windows Resource Kit (in case `sc.exe` is not on your system)
https://www.microsoft.com/download/details.aspx?id=17657

Webmin Installation:

- go to temporary webmin location and run `setup.bat` by a double click ...

Result:

- Webmin is installed in C:\webmin
   image

### DIFF
--- a/makedist.pl
+++ b/makedist.pl
@@ -14,7 +14,7 @@ $zipdir = "zips";
 
 @files = ("config.cgi", "config-*-linux",
 	  "config-solaris", "images", "index.cgi", "mime.types",
-	  "miniserv.pl", "os_list.txt", "perlpath.pl", "setup.sh", "setup.pl",
+	  "miniserv.pl", "os_list.txt", "perlpath.pl", "setup.sh", "setup.pl, setup.bat",
 	  "version", "web-lib.pl", "web-lib-funcs.pl", "README",
 	  "config_save.cgi", "chooser.cgi", "miniserv.pem",
 	  "config-aix", "update-from-repo.sh",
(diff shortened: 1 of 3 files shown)


jcameron:
thanks!

PR #891, opened 2018-04-06
add macOS 10.13

chrstphrchvz:
For consistency I used "Mac OS X" as the existing entries do, however Apple has changed how they refer to their OS over time (i.e. "OS X" as of 10.7 and "macOS" as of 10.12). Can the new entry adopt such usage, and is it a bad idea if the existing entries change? [**Edit**: I now see where oschooser.pl generates a menu with one entry for each unique name, so any inconsistent entries would show up as another choice.]


jcameron:
Best not to change the OS name, is it could confuse existing Webmin installs.

PR #897, opened 2018-04-15
fix synology regex

gnadelwartz:
while preparing a new release of my synomin SPK a user found a problem with regex, fixed.


jcameron:
Thanks!

PR #920, opened 2018-05-21
Print dev version nicely

qooob:
When installed from repo, instead of having `1.88305210044` all over UI, let's have `1.883.0521.0044 `.


jcameron:
This seems reasonable ... but in what cases is the version number that long??

chris001:
I think it's a long version number like that when you append the month, day (`0521`) and time (`0444`) of the build - this helps distinguish automatic builds downloaded by users directly as updates from this github repo - which can make possibly many updates on the same day, while in development, hotfixes, etc.

jcameron:
Ok, makes sense

PR #924, opened 2018-06-03
Fix size sorting issues

rostovtsev:
This [issue 57297](https://www.virtualmin.com/node/57297) and other, was there for quite a long time. Now, it can be sorted just fine for all languages, even though other other data types present. I made it better, than using _dataTables_ plugins.

Ref.: https://github.com/authentic-theme/authentic-theme/commit/3de119a732d180c68e5ef76962b1bab4fe4413ac

### DIFF
--- a/web-lib-funcs.pl
+++ b/web-lib-funcs.pl
@@ -8633,7 +8633,7 @@ else {
 	}
 my $sz = sprintf("%.2f", ($_[0]*1.0 / $units));
 $sz =~ s/\.00$//;
-return $sz." ".$uname;
+return '<span data-filesize-bytes="' . $_[0] . '">' . ($sz . " " . $uname) . '</span>';
 }
 
 =head2 get_perl_path


jcameron:
Should this be limited to just the authentic theme, like the similar date <span> change was? Also, it should be done only when viewed on the web, to prevent this <span> from appearing in emails.

rostovtsev:
Oh, alright. Then you want to let this accessible from the theme as `theme_nice_size`?

rostovtsev:
It works with other themes as well. Just maybe leaving it for all as it is?

webmin:
Ok .. but we will need to turn this off when the function is not called on the web.

webmin:
Maybe it's best to do it like this : https://github.com/webmin/webmin/commit/82dca72335842cc50ff8d1037c62e311f7ce67ca

rostovtsev:
Yes, yes, I remember perfectly how  it's done. I will do it. 😉

rostovtsev:
Yes, sure. I'll add it on the theme side. I got your point, for the future I will do that.

PR #953, opened 2018-08-20
Always define STATUS_* variables in status monitoring messages

mnoorenberghe:
The corresponding IF- variables also wouldn't be defined either so there was no way to avoid literal STATUS_* output for monitors which didn't provide useful values for these variables.


jcameron:
Thanks!

PR #1000, opened 2018-12-13
Fixed bug in parent column in DHCP list view

dremerb:
Reading the DHCP hosts, if a host was not defined in a subnet, it "inherited" the last read subnet.
This fix simply empties the variable and changes the lang files to enable no parent.


dremerb:
Oh, and btw, congrats to 1000 PRs

jcameron:
Thanks!

PR #1020, opened 2019-02-05
Run stat() for files that are bundled with module

rostovtsev:
This will increase the initial load by 2x times (when gzip third option is enabled and bundled version of the file is included)

Test case. 

Git version of the theme, where some files have their .gz variants, like `bundle.min.js.gz`, `bundle.min.css.gz` and some other.
Buffer size 1024.

Web Server Options and load time before/after the patch.

1. Only if pre-compressed .gz file exists - Load time: 1.46s/1.46s (same)
2. Never - Load time: 4.2s/4.2s (same)
3. Use pre-compressed file and compress dynamically - Load time: 1.40s/700ms (**big difference**)

When .gz files already exist on the theme/module side, miniserv tries to stat them first when 3 option is enabled, and that takes quite a lot of the time. It's much faster, just to gzip them instead.

### DIFF
--- a/miniserv.pl
+++ b/miniserv.pl
@@ -2608,7 +2608,7 @@ else {
 	print DEBUG "handle_request: outputting file $full\n";
 	$gzfile = $full.".gz";
 	$gzipped = 0;
-	if ($config{'gzip'} ne '0' && -r $gzfile && $acceptenc{'gzip'}) {
+	if (!length($config{'gzip'}) && -r $gzfile && $acceptenc{'gzip'}) {
 		# Using gzipped version
 		@stopen = stat($gzfile);
 		if ($stopen[9] >= $stfull[9] && open(FILE, $gzfile)) {


jcameron:
That's surprising .. is running stat really that slow? I'd expect it to be much faster than actually reading and compressing the file.

rostovtsev:
Yes, especially with lowered buffer level.

You can easily reproduce it?

rostovtsev:
It's reading large .gz files as well, I suppose.

jcameron:
I wasn't able to re-produce this on my system .. reading existing files was faster. But I  expect it depends on the relative speed of disk vs CPU on your system.

rostovtsev:
Okay, Jamie. Having new buffer set to 32768 it really becomes irrelevant (this is why you couldn't reproduce it). My initial point was that it's faster to build new files than reading existing, then comparing datetime, especially, if they are large (.7mb and larger).

What is the page initial load time for your case, which is displayed in Network tab (bottom right value) of browser's console.

PR #1031, opened 2019-03-02
Fix encoding and some typos in french translation

zeroheure:


jcameron:
Thanks!

zeroheure:
There is lot of files in lang dir that are not transcoded to UTF-8. Even recent ones. Should I do it ?

jcameron:
They aren't converted in the git repo, but they should exist in the distributed Webmin package.

PR #1032, opened 2019-03-02
transcode French help file to UTF-8.

zeroheure:


jcameron:
Thanks, but this file shouldn't be in UTF-8 encoding - those go in a separate file that is generated from it as part of Webmin's packaging process.

dktcoding:
Is there a specific reason behind this? Related to #1041

rostovtsev:
Can you please commit to non unicode variant of the file, preserving initial encoding?

zeroheure:
@iliaross I don't understand : I've just re-encoded the file to UTF8

rostovtsev:
@zeroheure At the moment, upon compilation, Webmin uses non-UTF variant of language file, and generates UTF-8 file. UTF-8 files are not supposed to be on the repo at all, as they are regenerated.

@jcameron Jamie, how much time it would take to do the opposite. Store all files in UTF-8 and transcode to other encodings if necessary (internally, so users would only face/deal with UTF-8 files to prevent encoding issues)?

jcameron:
Yes, this is in progress. I'm doing some work right now to fix up existing encodings, in preparation for UTF-8 conversion.

zeroheure:
@iliaross Looks like it doesn't work well for french as many generated files are not UTF-8

swelljoe:
I think/hope whatever encoding issues there were in the past have been resolved by the overhaul of translations a while back.

PR #1037, opened 2019-03-27
squid module: support for -n option for ACL

amishmm:
This PR adds support for -n option for Squid ACLs that support it. (dst, dstdomain, dstdom_regex).

ACL parsing code is currently very vague when it comes to ACL options. So I have tried to handle ACL options in a little better way. (by replacing vals variable with aclopts variable). It no more assumes that "-i" option will always be first (index 0) option.

However the new code does not change main logic which parses the ACL options. Parsing is still inefficient and may need changing in future when more squid ACL options are introduced.

Please check. Thank you.

### DIFF
--- a/squid/acl.cgi
+++ b/squid/acl.cgi
@@ -5,14 +5,14 @@
 use strict;
 use warnings;
 our (%text, %in, %access, $squid_version, %config, %acl_types,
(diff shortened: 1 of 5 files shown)


jcameron:
Thanks for the patch!

PR #1043, opened 2019-04-12
Improvement suggestion for output file prefix

seansan:
Improvement suggestion for output file prefix

This allows to create backup files with weekly/daily/montly versions

We can then setup 3 cronjobs same command, only different prefix

### DIFF
--- a/mysql/backup.pl
+++ b/mysql/backup.pl
@@ -14,6 +14,13 @@ else {
 	$cmode = 0;
 	}
 
+$prefix = "";
+if (defined $ARGV[3] and defined $ARGV[4]) {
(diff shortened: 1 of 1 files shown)


jcameron:
Can you update this to be compatible with my recent change to allow a prefix to be specified on the backup form? https://github.com/webmin/webmin/commit/77b3471442070d77545fba9b66865ab08cdc2c9d

seansan:
Maybe the change can be made by someone who knows this better. It is only a small improvement too long ago perl. Hope you understand.

jcameron:
Ok, will do

PR #1055, opened 2019-05-10
Update update_sched.cgi

t-offline:
Fix for:
Failed to update modules : You must enter an email address if only showing what would be updated.
Usermin Configuration ->Upgrade Usermin -> Scheduled update


jcameron:
Thanks!

PR #1072, opened 2019-06-12
Reenable Test Zone Transfer

t-offline:
Allow simple masters clause syntax (not complete)

### DIFF
--- a/bind8/xfer.cgi
+++ b/bind8/xfer.cgi
@@ -23,8 +23,21 @@ my $src = &find("transfer-source", $options->{'members'});
 # Get master IPs
 my $masters = &find("masters", $zconf);
 my @ips;
+my $conf = &get_config();
+my @confmasters = &find("masters", $conf);
 foreach my $av (@{$masters->{'members'}}) {
-	push(@ips, join(" ", @{$av->{'values'}}));
+	my $done = false;
+	foreach my $cm (@confmasters) {
+		if ($av->{'name'} eq $cm->{'value'}) {
+			foreach my $mb (@{$cm->{'members'}}) {
+				push(@ips, $mb->{name});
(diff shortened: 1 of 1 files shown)


jcameron:
Is this patch ready to merge?

t-offline:
My patch does not take care about all options in this clause:

`masters my-masters-name [port gp-num] [dscp gd-num] { ( masters-list |IP-Address [port p-num] [key key] ) ; [...] }; };`

http://www.zytrax.com/books/dns/ch7/masters.html


Same is true for:

`masters [port pg-num] [dscp dscp-num] { (my-masters-name | IP-address )[port p-num] [dscp dscp-num] [key key-name] ; [... ;] };`

http://www.zytrax.com/books/dns/ch7/zone.html#masters



So a "Test Zone Transfer" with my patch will only work with such entries:

```
masters my-masters-name { ( IP-Address ) ; [...] }; };
masters { (my-masters-name | IP-address ) ; [... ;] };
```



Example:

```
masters good-masters {192.168.2.7;10.2.3.15;2001:db8:0:1::15};
...
zone "example.com" {
    type slave;
    file "slave/example.com";
    masters {good-masters;};
};
```


I hope this helps to implement my patch.
TIA

jcameron:
Oh, so you're using a separate list of masters and referring to it in each zone?

t-offline:
> 
> 
> Oh, so you're using a separate list of masters and referring to it in each zone?

Yes.

swelljoe:
This has conflicts and no follow-ups since 2019, so I'm closing it. I hope whatever issue it addresses has been resolved some other way (I see another related PR has a note that it was resolved by Jamie).

Feel free to follow up with a new PR or issue, if the problem remains.

PR #1087, opened 2019-08-06
Freeze and thaw zone when re-signing.

chrisstaite:
When re-signing a zone that has a journal, it *must* be frozen first and thawed afterwards.  The comment in the library (bind8-lib.pl) states "before_editing(&zone) -Must be called before reading a zone file with intent to edit".


jcameron:
Thanks! This will go into the next release..

PR #1101, opened 2019-09-03
Fix renewal error

mzglinski:
I was having problems with certificates renewal. Running a command 

`virtualmin generate-letsencrypt-cert --domain DOMAIN`

would result in following error

```
Requesting SSL certificate for ***** ..
.. failed : Web-based validation failed : Saving debug log to /var/log/letsencrypt/letsencrypt.log
Plugins selected: Authenticator webroot, Installer None
Cert not yet due for renewal

You have an existing certificate that has exactly the same domains or certificate name you requested and isn't close to expiry.
(ref: /etc/letsencrypt/renewal/*******.conf)

What would you like to do?
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
1: Keep the existing certificate for now
2: Renew & replace the cert (limit ~5 per 7 days)
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
Select the appropriate number [1-2] then [enter] (press 'c' to cancel):
** Invalid input **
Select the appropriate number [1-2] then [enter] (press 'c' to cancel): An unexpected error occurred:
EOFError
Please see the logfiles in /var/log/letsencrypt for more details.

DNS-based validation failed : Saving debug log to /var/log/letsencrypt/letsencrypt.log
Plugins selected: Authenticator manual, Installer None
Cert not yet due for renewal

You have an existing certificate that has exactly the same domains or certificate name you requested and isn't close to expiry.
(ref: /etc/letsencrypt/renewal/****.conf)

What would you like to do?
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
1: Keep the existing certificate for now
2: Renew & replace the cert (limit ~5 per 7 days)
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
Select the appropriate number [1-2] then [enter] (press 'c' to cancel):
** Invalid input **
Select the appropriate number [1-2] then [enter] (press 'c' to cancel): An unexpected error occurred:
EOFError
Please see the logfiles in /var/log/letsencrypt for more details.
```

Adding `--keep-until-expiring` fixed the issue, however, I'm not sure whether this is a proper solution.

System info:
System: Debian 10 Buster 
certbot 0.31.0
Webmin and Virtualmin installed from the repository.

### DIFF
--- a/webmin/letsencrypt-lib.pl
+++ b/webmin/letsencrypt-lib.pl
@@ -158,6 +158,7 @@ if (($letsencrypt_cmd && -d "/etc/letsencrypt/accounts") || $wildcard) {
 			join("", map { " -d ".quotemeta($_) } @doms).
 			" --webroot-path ".quotemeta($webroot).
 			" --duplicate".
+			" --keep-until-expiring".
 			" --manual-public-ip-logging-ok".
 			" --config $temp".
 			" --rsa-key-size $size".
@@ -177,6 +178,7 @@ if (($letsencrypt_cmd && -d "/etc/letsencrypt/accounts") || $wildcard) {
 			" --manual-auth-hook $dns_hook".
 			" --manual-cleanup-hook $cleanup_hook".
 			" --duplicate".
+			" --keep-until-expiring".
 			" --manual-public-ip-logging-ok".
 			" --config $temp".
 			" --rsa-key-size $size".
--- a/webmin/letsencrypt-lib.pl
+++ b/webmin/letsencrypt-lib.pl
@@ -158,7 +158,7 @@ if (($letsencrypt_cmd && -d "/etc/letsencrypt/accounts") || $wildcard) {
 			join("", map { " -d ".quotemeta($_) } @doms).
 			" --webroot-path ".quotemeta($webroot).
 			" --duplicate".
-			" --keep-until-expiring".
+			" --force-renewal".
 			" --manual-public-ip-logging-ok".
 			" --config $temp".
 			" --rsa-key-size $size".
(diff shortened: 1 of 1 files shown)


jcameron:
Isn't --force-renewal better, because otherwise Virtualmin may think the cert has been renewed when it really hasn't.

stu177:
I've been getting this issue on my automated renewals (#1124) and another user referenced this PR. I don't know if this file is used for the scheduled renewal tasks as well.

If it is, adding `--force-renewal` to this is not a good idea and could easily cause issues hitting rate limits.

The [Certbot docs](https://certbot.eff.org/docs/using.html) mention that `--force-renewal` should only be used for manual renewals and not for any scheduled/automated runs.

chris001:
Agreed.  `--keep-until-expiring` is best according to the docs.  You don't want to needlessly waste your five forced renewals per week.

stu177:
Any update on this? Just that the changes made in the PR, in my opinion, could be completely functionality breaking by forcing all certificate renewals.

This is what I'm running at the moment and seems to work fine:
````perl
"cd $dir && (echo A | $letsencrypt_cmd certonly".
" -a webroot ".
join("", map { " -d ".quotemeta($_) } @doms).
" --webroot-path ".quotemeta($webroot).
#" --duplicate".
" --keep-until-expiring".
" --manual-public-ip-logging-ok".
" --config $temp".
" --rsa-key-size $size".
" --cert-name ".quotemeta($doms[0]).
($staging ? " --test-cert" : "").
" 2>&1)");
````

On a separate note, I've commented out duplicate as I don't know why this would want to be used rather than renewing the existing?

Issue #1100, opened 2019-09-02
DHCP does not detect vlan network interfaces


vsc55:
Hi,

I had to add a VLAN to my DHCP server and when specifying the Network Interface, I realized that it only detected "lo" and "eth0".
The vlan that are configured in the system do not appear, in my case "eth0.150".

Always start just listening to "eth0" and I had to edit manually "/etc/default/isc-dhcp-server".
I have also detected that when WEBMIN is in Spanish, an option appears in the configuration parameters of the module with the title, "Interfaces donde ejecutar DHCP" which seems to be to force some listening interfaces but it does not work, if WEBMIN is in English this does not appear option.

Config Net:
```
# ifconfig
eth0: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1500
        RX packets 140981  bytes 12924581 (12.3 MiB)
        RX errors 0  dropped 1  overruns 0  frame 0
        TX packets 134309  bytes 34049525 (32.4 MiB)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0

eth0.150: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1500
        RX packets 673  bytes 180174 (175.9 KiB)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 139  bytes 13187 (12.8 KiB)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0

lo: flags=73<UP,LOOPBACK,RUNNING>  mtu 65536
        inet 127.0.0.1  netmask 255.0.0.0
        inet6 ::1  prefixlen 128  scopeid 0x10<host>
        loop  txqueuelen 1000  (Local Loopback)
        RX packets 728  bytes 78591 (76.7 KiB)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 728  bytes 78591 (76.7 KiB)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0
```

List net detect:
![image](https://user-images.githubusercontent.com/13438676/64112004-5cef2900-cd86-11e9-9a43-dd48cadc7c6d.png)

Option not work:
![image](https://user-images.githubusercontent.com/13438676/64111059-05e85480-cd84-11e9-8346-c9ffb337a901.png)

Webmin version: 1.930
System: Raspbian GNU/Linux 10

vsc55:
Pull #1103 fix

PR #1103, opened 2019-09-03
fix error the donot detect interfac vlan (ethx.x).

vsc55:
issues #1100

### DIFF
--- a/dhcpd/edit_iface.cgi
+++ b/dhcpd/edit_iface.cgi
@@ -95,12 +95,12 @@ my $val;
 if (&foreign_check("net")) {
 	%got = map { $_, 1 } split(/\s+/, $iface);
 	&foreign_require("net", "net-lib.pl");
-	@ifaces = grep { $_->{'virtual'} eq '' } &net::active_interfaces();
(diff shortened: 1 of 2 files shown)


vsc55:
> 
> 
> Isn't it better to fix this in the `net` module by making `active_interfaces` include vlans?

Yes, but it is my first time with Perl code and I did not want to go deeper into the code.
Let's see if I have a while and I look more deeply.

vsc55:
Hi @jcameron,
First sorry for my English made in Google Translater.

I've been thinking (sometimes I do it and everything 😄) and apart from solving the "active_interfaces" to detect the vlan.

Don't you think it would be nice to let all the interfaces in the system come out, not just the active ones?
It may happen that you want to leave the server preconfigured and that at the time of configuring that interface you do not have it active.
What do you think about the approach?

Best regards.

vsc55:
Fixed the problem of "active_interfaces".
The vlan detection failure occurred when using the "ip" command. With the "ifconfig" command, the vlan is detected correctly.

jcameron:
Thanks!

PR #1106, opened 2019-09-05
Update password_change.cgi

### LINKED ISSUE #1105: find a bug in password_change.cgi
[line 21](https://github.com/webmin/webmin/blob/master/password_change.cgi#L21)

Always goes into update password branch even got an invlid user.
```perl
	($wuser) = grep { $_->{'name'} eq $in{'user'} } &acl::list_users();
	if ($wuser->{'pass'} eq 'x') {
		# A Webmin user, but using Unix authentication
                # actually not reach here in this case.
		$wuser = undef;
		}
# some code.
if ($wuser) {
	# Update Webmin user's password
        # always reach here even $in{'user'} is invalid or null like `?user=invalid`.
```

[related](https://stackoverflow.com/questions/57788390/)

Minty-He:
check if $wuser exist before touch it.
Fixes #1105


jcameron:
Thanks for the fix!

PR #1121, opened 2019-09-30
Dhcp config list iface listen

vsc55:
New configuration to be able to define the interfaces that appear in the list in the "Edit Network Interface" option.
You can choose between only the active interfaces (current mode), or all the interfaces that appear in the system.


swelljoe:
Since this has merge conflicts and no activity in seven years (though it seems reasonable), and because I think ISC has retired DHCPd and replaced it with a new thing that I guess we should support, I'm going to close this.

If you still need it, and want to make a new PR, I'll try to make sure it doesn't fall on the floor again.

vsc55:
Hi @swelljoe,

I have created a new PR on the new master branch, now there should be no conflicts.

https://github.com/webmin/webmin/pull/2665

PR #1133, opened 2019-10-31
Improved Spanish Translations

Ismael034:
I have changed the structure to look like the default English file and added some translations left. I will translate those in the next days


Ismael034:
All the strings are translated. Ready to merge

jcameron:
Thanks!

Ismael034:
Hi, I don't known why the words with accent's are showing as Ã³ .I don't know if git has some troubles with iso-8859-1 but it doesn't make sense because the first file I uploaded is ok

PR #1138, opened 2019-11-14
Correct yum check-update multiline output

seabres:
For long package names yum check-update produce multi-line output.
This leads to incorrect display of packages and not being able to select for update.
This patch provide a way to combine the multi-line output into single line for each package.


jcameron:
Thanks!

PR #1139, opened 2019-11-14
Search date sort

seabres:
The date in log search and csv export are in standard webmin format.
This means they are not alphabetically sorted, which makes parsing and copying of specific part hard to identify.
This patch modifies the date output (only the output, not the log itself) to be alphabetically (ASCII) sorted.
Additionally it filter the "last" record, which is sometimes inserted.


jcameron:
Thanks!

PR #1141, opened 2019-11-18
Search through rotated action log files

seabres:
When using Webmin for installation with long term logfile archiving requirements, rotating the webmin action log is necessary to wipe them out after some period.
Doing this, the "Webmin Action Log" module is not capable of search through the rotated files.

This patch creates a possible method for handling this with adding a new global config option "logsearch_rotated".
When this is set to 1, the action log modules composes a summary action log file with all rotated contents and uses this as source for index and display.

Alternative way would be to include the filename also in %index. But there is the problem on deleting from %index, when an external logrotate has removed a rotated file.

### DIFF
--- a/webminlog/search.cgi
+++ b/webminlog/search.cgi
@@ -5,6 +5,7 @@
 use strict;
 use warnings;
 use Time::Local;
+use File::Basename;
 require './webminlog-lib.pl';
 our (%text, %config, %gconfig, $webmin_logfile, %in, $in);
 &foreign_require("acl", "acl-lib.pl");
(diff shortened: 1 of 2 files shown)


swelljoe:
I don't recognize the `.sum` file extension convention here. `logrotate` names rotated files with a number (e.g. `webmin.log.1`) and maybe `.gz` when compressed. If you still want to merge this (and I think we probably should be automatically doing something about log rotation of Webmin files, including searching them), it'll need to be updated and needs to more generally handle rotated logs...certainly the standard log rotation tool on Linux is logrotate, I don't know about other OSes.

Anyway, closing since there are conflicts and this isn't general enough for inclusion as-is, even if we resolve the conflicts.

PR #1142, opened 2019-11-18
Allow multiple LDAP server for userdb

seabres:
The userdb string in miniserv.conf allow only one LDAP server in the connection string.
When using LDAP for user management, multiple LDAP server instances is required for failover, which leads to unavailability of Webmin if the specified one fails, especially if LDAP multi master is used.

This patch adds the possibility to specify multiple LDAP hosts in the userdb host part by separating them with comma.
userdb=ldap://user:pass@host1,host2,host3:port/ou=webmin...

It converts the list of hosts into an array for Net::LDAP->new and remembers the used LDAP server Net::LDAP has chosen. This will be used until the next connection error and the reelect again with the host array.
Specified port number is used for all LDAP hosts in the array, which is no limitation, since the scheme and the TLS establishment has to be also the same for all LDAP servers.

There is one drawback with the used implementation. The regex in function split_userdb_string must be extended to allow comma in host parts. But the mysql and postgresql connect methods cannot handle commas used in their connection strings (unlikely to be used, but who knows, what user decides). If this is seen as problem, the only known solution is to create a different split_userdb_string function for use with LDAP connection strings.

### DIFF
--- a/web-lib-funcs.pl
+++ b/web-lib-funcs.pl
@@ -10428,6 +10428,7 @@ On failure returns an error message string. In an array context, returns the
 protocol type too.
 
 =cut
+my $lastldapserver;
 sub connect_userdb
 {
 my ($str) = @_;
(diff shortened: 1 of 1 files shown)


Neustradamus:
@iliaross: Any news about this @seabres PR?

jcameron:
Looks like a reasonable PR - I've merged it!

PR #1143, opened 2019-11-20
Update acme-tiny client to use ACMEv2

2ndkauboy:
Fixes the issue described here: https://www.virtualmin.com/comment/819792


2ndkauboy:
The file was previously deleted in this commit, probably by accident: https://github.com/webmin/webmin/commit/3c97d715745c6b0de431837aed1ce764d40acc77

jcameron:
See discussion on https://www.virtualmin.com/comment/819792 - it's not yet clear if even including the acme_tiny.py client is the best solution.

swelljoe:
There are merge conflicts, so I assume `acme_tiny.py` got updated already via another merge. I think we're generally trying to avoid using `acme_tiny.py`, anyway. And, no follow up in six years, so I'll close this in pursuit of tidiness.

PR #1145, opened 2019-11-21
Add SHA512 to ldap-useradmin

seabres:
Add SHA512 password encryption method to ldap-useradmin module as additional configurable option.
The implementation uses the already defined methods from module useradmin.


jcameron:
Thanks!

PR #1150, opened 2019-11-27
Correct acl usage of ldap-useradmin

seabres:
The acl system of module ldap-useradmin include many acl roles not editable. They are in defaultacl and saved in acl_security_save, but never created in the acl form to be changed.

This patch deletes all not editable acl rights and uses them from the underlying useradmin module., instead of setting some of them to always allowed, as for instance udelete.
This is luckily possible due to exact the same names of acl roles between the two modules. Extending the acl roles for ldap-useradmin is easily possible, as only acl values for not present ldap-useradmin acl settings are copied from useradmin.

The list of secondary groups is limited to only allowed group from the access rights instead of showing all.
Edit masks for user and group consult access rights for options instead of display options always, for instance "change user in other modules".
They also consult module config to not display options not possible to execute, for instance create samba group when not group class is specified.

Access rights are also checked during save of user to reject group membership for not allowed groups. This is to prevent backdoor with direct HTTP request.


jcameron:
Thanks for this patch! However, are you also adding or removing new ACL checks? I see some extra checks like "makehome" and "cothers" are referred to in edit_user.cgi

seabres:
True, ldap-useradmin is very inconsistent in this area.
The configuration has many items with option "same as user and group module", but it has no code to check the ACL for that to apply.
One example you mentioned is homedir. If "makehome" ACL is set to No in useradmin, ldap-useradmin will let the user select to create homedir. It does not either bring its own ACL, nor use useradmin ACL. This is very inconsistent. As administrator, i can allow user creation but without homedir as OS user, but for LDAP user it is impossible to forbid homedir creation.

Yes, the patch implements additional ACL checks, but it does not remove any existing.
I try to implement in ldap-useradmin as much as possible of ACL checking from useradmin to it. But there is not enough time at the moment to do all right now. This is why started with the ones needed right now by myself. More will follow (have already two new in the queue).
Also to see if the way of adding ACL is applicable for the community.

In practice, the patch is written to be extendable in future.
Writing additional code to check ACL is independent of ACL existing in ldap-useradmin or useradmin.
Writing additional code for ldap-useradmin ACL will overwrite used useradmin ACL.

swelljoe:
Since there are merge conflicts and no follow up in six years, I'm going to close this. If it's still an issue, we'll need a new PR (preferably clean...just does the required thing, not open the door for a bunch of other changes).

PR #1151, opened 2019-12-02
Make acl grename editable

seabres:
The module useradmin define in defaultacl the right grename with 1.
Some module code uses this access right, but it cannot be edited in module acl forms.
Even the text is existing in language files.

This patch make the grename acl editable.


jcameron:
Thanks!

PR #1153, opened 2019-12-03
Create home directory in filemin if not exist

seabres:
Filemin stores the user configuration in the home directory of the user itself. But it creates the config directory after switching to the desired user account.

If PAM is configured with makehomedir plugin, home directories may not exist until the user first login and cannot be created by the switched Webmin user.
On use of filemin as replacement for interactive login, this is not possible before creating home directory by root or PAM module with suid rights.. When using Webmin on multiple servers in a LDAP domain, this cannot be handled.

This patch creates the home directory of the user to switch to before switching to the user in case not existing and apply access right defaults for home directories.
Afterwards handling of filemin config files can process as before.

### DIFF
--- a/filemin/filemin-lib.pl
+++ b/filemin/filemin-lib.pl
@@ -46,10 +46,21 @@ sub get_paths {
         @remote_user_info = getpwnam($access{'work_as_user'});
         @remote_user_info ||
             &error("Unix user $access{'work_as_user'} does not exist!");
+        if(!-e $remote_user_info[7]) {
+            mkdir($remote_user_info[7]) or &error("$text{'error_creating_conf'}: $!");
+            chmod(0700, $remote_user_info[7]);
+            chown($remote_user_info[2], $remote_user_info[3], $remote_user_info[7]);
+        }
         &switch_to_unix_user(\@remote_user_info);
     }
     else {
         # The Webmin user we are connected as
+        @remote_user_info = getpwnam($remote_user);
+        if(!-e $remote_user_info[7]) {
+            mkdir($remote_user_info[7]) or &error("$text{'error_creating_conf'}: $!");
+            chmod(0700, $remote_user_info[7]);
+            chown($remote_user_info[2], $remote_user_info[3], $remote_user_info[7]);
+        }
         &switch_to_remote_user();
     }
 


jcameron:
Actually you can get this same behavior already by adding the line `create_homedir=1` to `/etc/webmin/config`

chris001:
> Actually you can get this same behavior already by adding the line `create_homedir=1` to `/etc/webmin/config`

Should this setting `create_homedir=1` be the default setting ?

jcameron:
Probably not, as it applies any time the user logs into Webmin, and so could create home dirs in cases where the admin doesn't really want it.

seabres:
Agree with Jamie, create_homedir=1 shall not be default.
But there is a difference between create_homedir and this patch.

From my current knowledge filemin is the only module, which requires a homedir.
The setting create_homedir creates the homedir unconditional, even if the user has no access to filemin module. In installations with many machines and many user with several user levels (access rights via ldap) this results very fast having homedir for every user on every machine, even if Webmin is used instead of shell.
This patch creates it only on demand, if the user has access to filemin and the user uses it.

Is it preferred to share code with create_homedir ?
If yes, i can rewrite the patch accordingly.

swelljoe:
This feels problematic. It seems like Filemin shouldn't require a home directory. It assumes a Webmin user has a matching system user, which isn't always true.

So the problem is that Filemin makes assumptions that it shouldn't, and the solution probably isn't for it to try to re-shape the world to fit its wrong assumptions.

chris001:
I guess, most service accounts and some user accounts, shouldn't have a home directory created for them, by default.

swelljoe:
I'd guess it's maybe using home to store configuration, but Webmin has user configuration storage already (in various locations in `/etc/webmin`) and functions for handling it.

jcameron:
I assume the problem is that filemin is configured to limit the user to their home directory, but this doesn't work if there is no home dir?

seabres:
filemin stores the configuration in users home directory, regardless of the configured directory limits for the user. The configuration in this case is which columns to display and how.
Agree, it is maybe no proper idea to store it in homedir. In case of using ldap based user accounts, the homedir approach is bad, because it applies only for one server.

So, we would have two options:
- store as ACL
     pro: applies to all servers same
     con: user may not be able to configure display columns any more
- store in /etc/webmin/filemin
     pro: user can configure display columns
     con: applies only to one server
Dependent on the preferred solution, i can rewrite the patch for the desired configuration storage method and avoid homedir creation.

swelljoe:
This has merge conflicts and no follow up in six years, so I'll close it. I seem to recall Ilia fixed the problems in Filemin expecting a home directory and saving configs in the wrong place (I hope).

PR #1155, opened 2019-12-15
Make output of setup.bat more explicit

gnadelwartz:
because of a rant in https://sourceforge.net/p/webadmin/bugs/5332/ I updated the messages of setup,bat:

- the script is unoffical and try to guide unexperienced users
- offical information about setup is found at https://sourceforge.net/p/webadmin/bugs/5332/

As I don't know if Jamie supports and test Windows Installation or it's a "may work or not feature" we should also highligth the Windows support status an the offical page.


gnadelwartz:
there is more to do to install on windows, will file a new PR

PR #1160, opened 2019-12-22
Update german translation to mailboxes module

gnadelwartz:
while testing @rostovtsev new Mail UI I found some not translated or not fitting strings in german translation.

I addtion " Mail" means always "e-mail" in german, because no one use the "correct" translation "e-Post".


gnadelwartz:
To find the missing strings in a webmin / usermin lang file, I used this simple bash script:

```bash
#!/bin/bash
#
# compare_lang.sh
# quick and dirty script to find new/missing lang strings for webmin lang file
#

# $1 = lang file to check for missing strings against en

if [ "$1" == "" ]
then
        echo "no lang file to compare against en given ..."
        echo "usage: $0 file"
        exit 1
fi

while read message
do
        [ "$message" == "" ] && continue
        key="${message%=*}"
        grep -e "^${key}=" $1 >/dev/null || echo "${message}"
done < en

```

PR #1163, opened 2019-12-22
help: change de lang files to real utf-8?

gnadelwartz:
may be you can advise ...

I tried to change the german lang files to real iso-8859-1. "file -i" say they are iso-58859-1.
Also if you view the full file in github, you will see the umlauts.

nevertheless: github and webmin shows <?> instead of german umlauts, any idea?


chris001:
**Umlauts are represented as two characters in `UTF-8`.**  They convert fine and work well.  _The problem comes from legacy programs that expect 1 byte per character._  For these legacy programs, `ISO-8859-1` (aka Latin-1) has 1-byte umlauts. 

*    ASCII: 7 bits. 128 code points.
*    ISO-8859-1: 8 bits. 256 code points.
*    UTF-8: 8-32 bits (1-4 bytes). 1,112,064 code points.

Both `ISO-8859-1` (Latin-1) and `UTF-8` are backwards compatible with `ASCII`, but `UTF-8` is not backwards compatible with `ISO-8859-1`.

Ref: 
1. https://stackoverflow.com/questions/7048745/what-is-the-difference-between-utf-8-and-iso-8859-1
2. https://en.wikipedia.org/wiki/ISO/IEC_8859-1

For the language files, it looks like you're using 7-bit `ASCII` text, with one-byte `UTF-8` special characters indicated by something like `&#221` (one 0-255 decimal value, to be converted to one 8-bit byte), and two-byte `UTF-8` characters indicated by something like `&#226;&#242;` (two decimal values 0-255, supposed to be converted to two 8-bit bytes.)

In fact, `HTML 4.0` documents are **supposed to** output pure `UTF-8`, why not also save these language files in 7-bit `ASCII`, with `UTF-8` characters indicated with two, three, or four "bytes" (0-255 decimal values in the form `&#123` for one byte, `&#123;&#234;&#012;&#111;` for four bytes, etc.).

gnadelwartz:
Yes, Utf-8 would be the best solution.

unfortunately webmin does not use utf-8 as default for lang files, for historical reasons the files are written in local encoding and converted to utf-8 on release.

but as utf-8 is supported I'll try to convert german to an uft-8 only language and adjust the lang config accordingly

gnadelwartz:
what needs to be changed?

- lang-list.txt: lang=de,titles=1,charset=UTF-8	German
- convert all lang/de files to lang/de.UTF-8 files
- convert all ulang/de files to ulang/de.UTF-8 files on usermin
- convert all config.info.de files to config.info.de.UTF-8 files
- convert all help/*.de files to help/*.de.UTF-8 files
- symlink all original files to new ones for compatibility?

anything else?

gnadelwartz:
this PR is not for merging, but for testing and comment on it.


my script has now converted all needed files and I adjusted the language "de" to UTF-8.

- do we need the symlinks for the "old" files like autentic or can they be removed?

chris001:
https://stackshare.io/stackups/crowdin-vs-transifex
https://github.com/marketplace/crowdin
https://github.com/apps/transifex-integration

We should try out CROWDIN and TRANSIFEX because these let you get the user interface text translation maintained by volunteers, and integrate with github, you can pull the latest translation down and ship it with the software on release day.  You just need to write a simple script to automate that.  They're free for open source projects!

gnadelwartz:
I now have a working de.utf-8 Version in my test environment. l'll create separate pr to discuss it

PR #1176, opened 2020-01-09
Convert German Translation to UTF-8 like authentic-theme

gnadelwartz:
There are some issues with german translation, e.g. the original translator choose to encode "umlauts" not as an iso-8859-1 code point but as  iso-8859-1 HTML entitiy.

e.g. an iso-8859-1 'ü' is encoded as ``` &#252;``` 

Some time ago I noticed that ```iconv``` does not convert these entities to UTF-8 (at least on my linux box). in addition authentic-theme escapes strings more and more, so these entities are visible in several places where the browser had converted them automagically in the past.

The intuitive solution would be to convert these entities to real iso-8859-1 codepoints, which I've done in this PR https://github.com/webmin/webmin/pull/1163

even this worked perfect on my local installation, github did not like it and shows mixed charsets and strange characters ...

So why not directly convert german langs files to UTF-8 like it's in authetic-theme and use this as blueprint for other languages. I wrote a script to do this for me https://github.com/gnadelwartz/webmin/blob/translate/convert_lang_de.sh (ups there is a ! missing)

this is whats done by this PR and I want to discuss with you ...


gnadelwartz:
One remaining problem I encontered:

- I I change the change for de lang to UTF-8 in "lang-list.txt" everthing works as expected BUT the lang settings of existing installations are changed to english 

- if I do not change de lang in "lang-list.txt" it works also, but ```chinese-to-utf.pl``` overwrites the created "de.UTF-8" files when creating a new release.

A possible solution may be to do not change the setting in "lang-list.txt" AND do not overwrite existing UTF-8 transaltions https://github.com/gnadelwartz/webmin/commit/fff6c5dee7faf8d0a019f81c23d53c75cd5d73a5

Whats your opronion?

chris001:
The `UTF-8` text looks good, the umlauts are working in there, no funny looking incorrect symbols, so the errors about the international text output seem to be on the way to being resolved.

gnadelwartz:
@jcameron I did not autoconvert the config.info files because I'm not shure if I damage the config Options, e.g  comma in the text of a config description which may misinterpreted as option seperator

do you see any risc, e.g. addtional spaces or minus in the text part?

gnadelwartz:
@jcameron after some time I came to the conclusion that I can safely convert config.info files also, if I convert ```&44;``` to a dot ```.``` instead of a comma ```,```...

rostovtsev:
Kay, can you please not add to one PR, both text files and scripts (like chinese-to-utf8.pl and other)?

gnadelwartz:
@iliaross do you mean a PR for scripts etc and one for the resulting changes?

ok, will do a separate pr for Scripts and manual changes ...

rostovtsev:
Yeah, to separate text/language files from scripts.

gnadelwartz:
if the idea and the scripts are acceped, I'll provide a new PR with the automatic done changes ...

PR #1177, opened 2020-01-10
convert-lang-de.sh to do german conversion like Authentic-Theme

gnadelwartz:
see #1176

### DIFF
--- a/chinese-to-utf8.pl
+++ b/chinese-to-utf8.pl
@@ -107,7 +107,7 @@ foreach $m (@modules) {
 		}
 	foreach $l (@default_langs) {
 		# do convert if readable and not a symlink
-		if (-r "$m/lang/$l" && ! -l "$m/lang/$l") {
+		if (-r "$m/ulang/$l" && ! -l "$m/ulang/$l") {
 			system("iconv -c -f iso-8859-1 -t UTF-8 - <$m/ulang/$l >$m/ulang/$l.UTF-8");
 			}
 		}
@@ -277,7 +277,7 @@ foreach $m (@modules) {
 	foreach $l (@default_langs) {
 		%cinfo = ( );
 		# do convert if readable and not a symlink
-		if (! -l ("$m/config.info.$l" && &read_file("$m/config.info.$l", \%cinfo)) {
+		if (! -l "$m/config.info.$l" && &read_file("$m/config.info.$l", \%cinfo)) {
 			local %ocinfo = %cinfo;
 			foreach $k (keys %cinfo) {
 				$cinfo{$k} = &DefaultToUTF8($cinfo{$k});
--- a/chinese-to-utf8.pl
+++ b/chinese-to-utf8.pl
@@ -69,7 +69,8 @@ foreach $m (@modules) {
 			}
 		}
 	foreach $l (@default_langs) {
-		if (-r "$m/lang/$l") {
+		# do convert if readable and not a symlink
+		if (-r "$m/lang/$l" && ! -l "$m/lang/$l") {
 			system("iconv -c -f iso-8859-1 -t UTF-8 - <$m/lang/$l >$m/lang/$l.UTF-8");
 			}
 		}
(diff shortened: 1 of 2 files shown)


chinese-to-utf8.pl
```diff
@@ -69,7 +69,8 @@
 			}
 		}
 	foreach $l (@default_langs) {
-		if (-r "$m/lang/$l") {
+		# do convert if readable and not a symlink
```
gnadelwartz: checking for symlink should prevent overwriting of an already converted iso-8859-1 lang

gnadelwartz:
@jcameron @iliaross after fixing "module.info.xx..UTF-8" key, the script should convert de lang automatically to UTF-8 like authetic.theme UTF-8 usage and can merged.

may be someone can test it for an other language like russian or frensh, wicth I dont know :-)

gnadelwartz:
@iliaross after you merged the changes for usermin,mhow should we proceed with webmin?

rostovtsev:
@gnadelwartz I am working on the script to automatically translate to over a hundred languages, each and every string on each and every module, preserving already translated parts by humans.

.. keep tight .. :smile:

rostovtsev:
.. everything will be stored in _.utf-8_ encoding and be default, and there will be no more confusion for the future contributors. Committing using GitHub will not break encodings anymore. There will no longer be two files, in different encodings requiring compilation. 

Besides, the script will let developers, automatically add/remove needed pair of `key=value` to language file and automatically translate it to all pre-defined languages, with simply running:

~~~
languager --module=apache,bind --action=add --key=index_module_alternative_title --value="Module's Alternative Title"
~~~

gnadelwartz:
sounds great! now no more missing language strings

gnadelwartz:
may be mark autotranslated strings to make it easier for translators to find them

rostovtsev:
It's gonna flood most of the files badly. I don't think it's needed as you can easily notice if something's wrong by just reading it. 

My goal is to preserve all current commits made by people, without replacing theirs efforts with automated translation. Later, one could just read the file, edit it locally and then just copy/paste it to GitHub for pull request. That's what most people did and would do but having non _.utf8_ on the initial files would always break things out, and shouldn't happen anymore in 2020.

rostovtsev:
> Sounds great! now no more missing language strings

@jcameron @swelljoe @gnadelwartz @chris001

Okay. It's pretty much done.

1. First version of language script is here https://github.com/webmin/webmin/commit/c2006d1dfa438e092d622954a585f6f9dbf0f363
   - Run `manipulate-language -h` for more details.
   - Add/remove and spell check functionality is coming in the near future.

2. Today, at first, as a test run, been done only ACL module:
     - [Language](https://github.com/webmin/webmin/commit/99bb889e8b9483b05bc05b213eea7c319b913cf1) - all in UTF-8 no bad stuff, on example with [de](https://github.com/webmin/webmin/commit/99bb889e8b9483b05bc05b213eea7c319b913cf1#diff-a8b70fe5a499e639235d4a87a035c9fb) file.
   - [module.info](https://github.com/webmin/webmin/commit/09b69941151e74e838e9010d0c5e42d8d678284e) and [config.info](https://github.com/webmin/webmin/commit/876fe286334c530089832f34f8e61a9c47489104) also translated and transcoded (config only transcoded - it's fancy logic there to hack on, probably make it work in the future). Example of how it looked before and how sweet it [looks now](https://github.com/webmin/webmin/commit/876fe286334c530089832f34f8e61a9c47489104#diff-3b64fa6775957f2da96636a3988e7505).

3. Dropped all old conversion scripts and made changes to use utf-8 only, fixed **list_languages()** sub https://github.com/webmin/webmin/commit/adb4ad1797cdd37a87eacaa0755edaad93a8c182.


.. to be continued .. and looking forward for your comments! :wink:

rostovtsev:
.. and forgot to mentioned, that language list will also look very straight forward for the users now .. 

Screenshot from 2020-02-28 01-06-08

gnadelwartz:
@iliaross is working towards the final solution 👍

PR #1189, opened 2020-01-27
Update Mac OS X versions list and add version names

gilcel:
Update Mac OS X / macOS resp. with macOS 10.14 & 10.15 and add the name for every Mac's operating system.
This was tested on macOS 10.14.6 and works for usermin too.


rostovtsev:
Thanks.

PR #1193, opened 2020-02-04
Update masq file location and rename

Omoeba:
The location of the masq file was updated in 5.0.14 according to [the documentation](https://shorewall.org/two-interface.htm#SNAT).

### DIFF
--- a/shorewall/shorewall-lib.pl
+++ b/shorewall/shorewall-lib.pl
@@ -17,13 +17,13 @@ $shorewall_version = &get_shorewall_version(0);
 %access = &get_module_acl();
 
 @shorewall_files = ( 'zones', 'interfaces', 'policy', 'rules', 'tos',
-	   	     'masq', 'nat', 'proxyarp', 'routestopped',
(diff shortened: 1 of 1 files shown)


jcameron:
Should this patch check for the shorewall version before using the different name?

Omoeba:
Actually more changes would be required because the syntax of the `snat` file is different from the `masq` file. I don't know any perl so I won't be able to contribute further. It should definitely check the shorewall version to avoid breaking earlier versions.

swelljoe:
Give the change happened many years ago, and surely no one is running a seven+ year old version of Shorewall, maybe we should just merge this? No non-EOL version of Shorewall has the old location.

But, also Shorewall appears to have been abandoned around the time of this PR, so it may be we just want to remove the module. Last release was 5.2.8 in September of 2020.

swelljoe:
@jcameron any thoughts on this? I'd like to close it, one way or the other, and just merging it seems the simplest/rightest option.

iliaross:
This patch is neither safe nor complete enough to merge!

PR #1212, opened 2020-03-04
detect fdisk output fixed #1211

### LINKED ISSUE #1211: Webmin don't show partitions
Webmin don't show me any HDD partitions (https://localhost:10000/fdisk/index.cgi) on Debian 10 (Buster).

I found a regex problem in `fdisk/fdisk-lib.pl`. When the command `fdisk -l /dev/sda` return a number with decimal point, than this module don't find partitions.

This output works:
```
Disk /dev/sda: 60 GiB, 64424509440 bytes, 125829120 sectors
```

This output don't work:
```
Disk /dev/sda: 232.8 GiB, 249988907008 bytes, 488259584 sectors
```

macmon:


chris001:
Does you regex work with both `60.6 GiB` and `60 GiB` ?

macmon:
Yes, the new regex works with both 60.6 GiB and 60 GiB.

gnadelwartz:
inside  [ ] you don't need to escape the dot, so ` [\d.]+` should be enough.
in addition it may useful to also allow non english decimal seperator ... ` [\d,.]+`

macmon:
When Webmin executes the fdisk command, the decimal separator is always a dot, even if a comma is displayed as decimal separator in the user console.

PS: You're right, the escape is unnecessary.

PR #1215, opened 2020-03-12
German button label index_delete1 fixed

macmon:
The label of button index_delete1 is not correct, because with the button you can activate interfaces also.


iliarostovtsev:
Okay, great! Danke!

gnadelwartz:
nevertheless De-Aktiviere is unclear and no german word.
if its a toggle `(De-) Aktiviere` or `Aktiviere/Deakivere` may better

iliarostovtsev:
:smile: Boys, make up you minds. I will accept the fix, if needed, Kay!

PR #1219, opened 2020-03-18
Update setup.pl

edwardsmarkf:
webmin was not starting properly on my new Centos-8 using `exec`.

note:  this is really just a guess on my part.   somebody more knowledgeable than me should carefully review this.

### DIFF
--- a/setup.pl
+++ b/setup.pl
@@ -525,6 +525,10 @@ else {
 	if ($os_type eq "hpux") {
 		print START "exec '$wadir/miniserv.pl' $config_directory/miniserv.conf &\n";
 		}
+	elsif ($os_type eq "redhat-linux") {
+		## maybe include a test to determine where perl is?
+		print START "/usr/bin/perl '$wadir/miniserv.pl' $config_directory/miniserv.conf &\n";
+		}
 	else {
 		print START "exec '$wadir/miniserv.pl' $config_directory/miniserv.conf\n";
 		}


swelljoe:
Why doesn't `exec` work?

Is it possible Webmin didn't update the Perl path in the shebangs in all the executable files? Look at `webmin.pl`, is the shebang `#!/usr/bin/perl` or is it `#!/usr/local/bin/perl`?

That's the only reason I can think of for exec to fail. Can you also post the error you get when running it with exec? We need to sort the underlying cause; calling Perl directly shouldn't be necessary.

edwardsmarkf:
did you mean `miniserv.pl` ?  i am not seeing a `webmin.pl` file.  perhaps you mean `miniserv.pl`?

I am sure you are correct in that there is a better way to do this other than my suggestion.  my suggestion was more of a "cheap shot" idea, sorry.   however, it did get me past the issue.  

perhaps some command-line switch with the `exec` ?   weird thing is that when i just manually enter in the last line of the `/etc/webmin/start` in the command-line, it flashes some error but then quickly logs me out before i get a chance to see the error.

below is what i am seeing:

> # head  -1  /root/webmin/webmin-1.941/miniserv.pl
> #!/usr/bin/perl
> 
> # cat  /etc/webmin/start
> #!/bin/sh
> echo Starting Webmin server in /root/webmin/webmin-1.941
> trap '' 1
> LANG=
> export LANG
> #PERLIO=:raw
> unset PERLIO
> export PERLIO
> PERLLIB=/root/webmin/webmin-1.941
> export PERLLIB

> #     systemctl   start  webmin.service  ;
> Job for webmin.service failed because the control process exited with error code.
> See "systemctl status webmin.service" and "journalctl -xe" for details.

> # systemctl   status  webmin.service
> ● webmin.service - Webmin
>    Loaded: loaded (/usr/lib/systemd/system/webmin.service; enabled; vendor preset: disabled)
>    Active: failed (Result: exit-code) since Wed 2020-03-18 22:48:26 UTC; 9min ago
>   Process: 14680 ExecStart=/etc/webmin/start (code=exited, status=126)
> 
> Mar 18 22:48:26 virtualmin-2020-03-18 systemd[1]: Starting Webmin...
> Mar 18 22:48:26 virtualmin-2020-03-18 start[14680]: Starting Webmin server in /root/webmin/webmin-1.941
> Mar 18 22:48:26 virtualmin-2020-03-18 start[14680]: /etc/webmin/start: line 11: /root/webmin/webmin-1.941/miniserv.pl: Permission denied
> Mar 18 22:48:26 virtualmin-2020-03-18 start[14680]: /etc/webmin/start: line 11: exec: /root/webmin/webmin-1.941/miniserv.pl: cannot execute: Permission denied
> Mar 18 22:48:26 virtualmin-2020-03-18 systemd[1]: webmin.service: Control process exited, code=exited status=126
> Mar 18 22:48:26 virtualmin-2020-03-18 systemd[1]: webmin.service: Failed with result 'exit-code'.
> Mar 18 22:48:26 virtualmin-2020-03-18 systemd[1]: Failed to start Webmin.

swelljoe:
Yes, miniserv.pl is what I meant.

That's weird I don't see why it'd work calling it with Perl but not with exec. Is `miniserv.pl` executable? (I mean, does it have the x bit set?)

edwardsmarkf:
i assume `systemctl` runs everything under root?   should it matter?

```
# ls -l  /root/webmin/webmin-1.941/miniserv.pl
-rwxr-xr-x. 1 root bin 168446 Mar 18 22:45 /root/webmin/webmin-1.941/miniserv.pl
```
i did try doing  `chown root:root   /root/webmin/webmin-1.941/miniserv.pl;` too but that did not fix it.

swelljoe:
systemd starts as root, but does a su to whatever user is requested by the unit file, but `root` is the default, I would assume.

Is selinux enabled? Maybe it's got some sort of preventative of executables in /root or something. I don't know where selinux logs on CentOS 8...used to be in /var/log/secure, but I think it's probably in the journal now, somewhere.

edwardsmarkf:
`selinux` is new to me, i really dont know anything about it.  but with selinux enabled (the new default), then webmin seems to have issues, but disabled it starts fine.

i went to google-cloud, spun up a nice fresh new VPS with C-8, and installed the absolute bare minimum, then ran the webmin install script.  

here is a logfile i saw, but i do not see anything very informative:
```
# grep   webmin  /var/log/audit/audit.log
type=SERVICE_START msg=audit(1584571706.357:365): pid=1 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:init_t:s0 msg='unit=webmin comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=failed'UID="root" AUID="unset"
type=SERVICE_START msg=audit(1584574740.673:54): pid=1 uid=0 auid=4294967295 ses=4294967295 msg='unit=webmin comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=success'UID="root" AUID="unset"
```
where else shall i look?

swelljoe:
That's it. It's disallowing running things from /root.

Put it in /opt or /usr/local instead and try again.

swelljoe:
No need for this pull request. Problem is SELinux preventing execution in /root.

edwardsmarkf:
https://www.hpe.com/us/en/insights/articles/how-to-set-up-selinux-right-the-first-time-1901.html

> I hasten to stress that you should not adopt SELinux just because the extra-robust security functionality exists. Yes, I know how much security matters to your business, but SELinux isn’t for everyone. Running SELinux takes extra time, requires far more hands-on involvement than other Linux distributions, and demands deep knowledge about Linux architecture.

Joe, this has proven to be a bit more difficult than i expected it to me.  i am thinking about some sort of notice for the centos/redhad people about this issue?

`setenforce permissive ;   ## this is a simple workaround but it does not survive reboot.`

changing `/etc/selinux/config` from "enforcing" to "permissive" and then rebooting is the permanent solution:
```
###SELINUX=enforcing
SELINUX=permissive
```
Using "permissive" as opposed to "disabled" still writes possible security violations in the audit files.

SELinux does not look like something taken trivially.  And the learning curve is steeper than i anticipated. 

Joe, once again, and as always, THANK YOU SO MUCH for your assistance, and especially your ongoing work with webmin.

swelljoe:
Just don't install Webmin into `/root`. Don't make it complicated. ;-)

SELinux can be challenging, but it does *not* prevent Webmin from running. It's just preventing it from running in `/root` as the errors sort of clearly show. (That said, I do end up disabling SELinux most of the time, because it's such a pain in the ass.)

edwardsmarkf:
ha ha Thanks!   excellent advice.    one last, where do you recommend installing?

edwardsmarkf:
`/usr/local` ??   'mkdir -v /usr/local/webmin ;` ??

edwardsmarkf:
replaced with   https://github.com/webmin/webmin/pull/1222

PR #1222, opened 2020-03-23
Update init-lib.pl

edwardsmarkf:
explained here:  https://serverfault.com/questions/817552/systemd-drop-in-fails-to-create-pid-file

```Regrettably, systemd won't create a PID file for a non-forking service even if you specify a PIDFile= line in the service's unit file. But you may be able to cheat with an ExecStartPost= line, such as:```

the second (last) solution appears to work.


edwardsmarkf:
maybe this is not a good idea unless we can verify this is rh/c 8 ?

iliarostovtsev:
We expect PID file to be handled by the service.

PR #1223, opened 2020-03-25
Update web-lib-funcs.pl

edwardsmarkf:
Use `systemctl` to restart webmin.services.

### DIFF
--- a/web-lib-funcs.pl
+++ b/web-lib-funcs.pl
@@ -2124,6 +2124,16 @@ if ($main::webmin_script_type eq 'web' && !$ENV{"MINISERV_CONFIG"} &&
 	return;
 	}
 
+	###  https://www.thoughtco.com/perl-array-exec-system-function-tutorial-2641183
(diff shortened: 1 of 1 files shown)


iliarostovtsev:
No, no! :smile: This shouldn't be done like that!

Thanks for your PR and time. We actually already fixed the restart issue upon restart. It just needs to be fixed on the *Virtualmin-Config/Webmin* script, to restart Webmin like you mentioned. However, it's preferable to use `/etc/webmin/restart` _directly_ instead.

__Tip:__ For the future, when you provide a PR, please make sure that it would also work on other distros, particularly on BSD.

edwardsmarkf:
excellent point, thank you.  

>>> It just needs to be fixed on the Virtualmin-Config/Webmin script

do you have any time-frame on when this might be done?

iliarostovtsev:
> do you have any time-frame on when this might be done?

It's not done personally by me and thus I can not be exact.

PR #1226, opened 2020-03-28
The Japanese translation has been updated.

kosugikun:


iliarostovtsev:
Great, thanks! Now it's all in _UTF-8_. If you revise/fix **ja.auto** files (machine translated), then strings from them, should be move to regular **ja** file (human translated).

PR #1233, opened 2020-04-23
Some Cross-Site-Scripting (XSS) fixes

macmon:
Here are some fixes for XSS.

If you add some bad strings `<<script>alert("xss");//<</script>` into the network module, you can run JavaScript:
image
image

If you deactivate the referers with `referers_none=0`, than you can exploit with an URL:
https://192.168.2.1:10000/net/save_host.cgi?new=1&idx=&address=%3C%3Cscript%3Ealert(%22xss%22)%3B%2F%2F%3C%3C%2Fscript%3E

### DIFF
--- a/net/cygwin-lib.pl
+++ b/net/cygwin-lib.pl
@@ -422,7 +422,7 @@ sub parse_routing
 		    if ($gateway != $_->{'gateway'} ||
 			$gwmetric != $_->{'gwmetric'}) {
 			&check_ipaddress($gateway) ||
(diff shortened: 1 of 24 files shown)


macmon:
I have another fix in time module, one moment please …

macmon:
The output of function `sync_time` is now escape and in a pre block.

jcameron:
Thanks! Fortunately these XSS attacks aren't too scary because the `root` user can only exploit himself.

PR #1234, opened 2020-04-28
hostname update revised

macmon:
If you activate `Update hostname in host addresses if changed?` at `Hostname and DNS Client` (https://192.168.2.1:10000/net/list_dns.cgi) the hostname wasn't update always in `/etc/hosts`. This pull request is fix that and change also FQDN and PQDN if possible.

Example: Hostname was `server1` and Search domains was `example.local`. `/etc/hosts` has the line `127.0.1.1       server1`. If you change the hostname from `server1` to `server2` the line in hosts file wasn't update.

Also the hostname and domain validation had a bug, because the characters [\]^_` are not allowed. This is also fixed in this pull request.


jcameron:
Thanks for the patch!

PR #1239, opened 2020-05-09
Add support for redis session

ardabeyazoglu:
I added redis selection in php session options page. Using file based session is pretty outdated, slow and dangerous.


jcameron:
Thanks!

Issue #1135, opened 2019-11-06
Webmin STILL redirects to domain.com:10000 after login..


MacGyver27:
Hi, after some decent time of troubleshooting and looking into suggestions of #1049 and #1027 and #976 (and also https://sourceforge.net/p/webadmin/discussion/600155/thread/6eb89f60/?page=1&limit=25#8999) still cant fix that redirect problem...

visit -> admin.example.com -> logs into webmin -> redirected to: admin.example.com:10000 -> 404 page not found -> manually delete port -> successfully inside webmin web app
^^^ every time

I am using official Debian package.
Operating system: Debian Linux 10
Webmin version: 1.930
Authentic theme version: 19.40 (forced upgrade since I was not able to find `update-from-repo.sh script to perform update to the latest development version.. how do I do it BTW if using official Debian package?)

For the reverse proxy I am using popular traefik.
traefik:
```
#webmin
  [backends.webmin]
    [backends.webmin.servers.1]
      url = "http://127.0.0.1:10000"
      weight = 1

#webmin
  [frontends.webmin]
    backend = "webmin"
    entryPoints = ["http", "https"]
    passHostHeader = true
	[frontends.webmin.whiteList]
	  sourceRange = ["192.168.255.0/24", "10.8.0.0/24"]
	  useXForwardedFor = true
	[frontends.webmin.headers]
	  forceSTSHeader = true
	  STSSeconds = 315360000
	  STSIncludeSubdomains=true
	  STSPreload = true
    [frontends.webmin.routes.1]
      rule = "Host:admin.example.com"
```

webmin config:
```
relative_redir=0
passwd_mindex=4
product=webmin
real_os_version=10
mobile_theme=virtual-server-mobile
real_os_type=Debian Linux
tempdelete_days=7
passwd_cindex=2
find_pid_command=ps auwwwx | grep NAME | grep -v grep | awk '{ print $2 }'
referers_none=0
log=1
referers=admin.example.com
md5pass=1
ld_env=LD_LIBRARY_PATH
theme=authentic-theme
passwd_uindex=0
passwd_pindex=1
passwd_file=/etc/shadow
os_type=debian-linux
lang=en.UTF-8
by_view=0
os_version=10
path=/bin:/usr/bin:/sbin:/usr/sbin:/usr/local/bin
webprefix=
webprefixnoredir=1
dateformat=dd/mon/yyyy
help_width=
showlogin=0
hostnamemode=0
nofeedbackcc=0
[...]

PR #1271, opened 2020-07-24
Fix reverse proxy redirect for prefixed URL - miniserv.pl

aphalon:
Fix problem where when behind a reverse proxy running as a prefixed url (eg https://abc.com/webmin) that the logon does not redirect correctly to the /webmin part.

This is to complement the other redirect_xxx values that were added in 1.940 per @iliarostovtsev in comment https://github.com/webmin/webmin/issues/1135#issuecomment-565818024

### DIFF
--- a/miniserv.pl
+++ b/miniserv.pl
@@ -1533,6 +1533,9 @@ $portstr = $redirport == 80 && !$ssl ? "" :
 $redirhost = $config{'redirect_host'} || $host;
 $hostport = &check_ip6address($redirhost) ? "[".$redirhost."]".$portstr
 				          : $redirhost.$portstr;
+# If the redirect_prefix exists change redirect base to include the prefix
+$hostport = exists($config{'redirect_prefix'}) ?
+		$hostport.$config{'redirect_prefix'} : $hostport;
 $prot = $ssl ? "https" : "http";
 
 undef(%in);


iliarostovtsev:
What about `webprefix`? Did you check [redirect](https://github.com/webmin/webmin/blob/master/web-lib-funcs.pl#L1346-L1398) sub?

aphalon:
I have the webprefix value set in the config file (see https://github.com/webmin/webmin/issues/1135#issuecomment-663453121 for config)

I took a look at `web-lib-funcs.pl` and expected that that the code `my $wp = $gconfig{'webprefixnoredir'} ? undef : $gconfig{'webprefix'};` in the `redirect` sub should handle this but it did not. 

I've just dig a little and found that the code that does the post logon redirect does not use the `redirect` sub - instead it builds the target Location itself based on the value of `page` within the `$in` hash.

```
		if ($in{'save'}) {
			&write_data("Set-Cookie: $sidname=$sid; path=$cpath; ".
			    "expires=\"Thu, 31-Dec-2037 00:00:00\"$sec\r\n");
			}
		else {
			&write_data("Set-Cookie: $sidname=$sid; path=$cpath".
				    "$sec\r\n");
			}
		&write_data("Location: $prot://$hostport$in{'page'}\r\n");
```

Because my change extends $hostport this works. I guess the question is whether it should be left as per the pull request or moved to be logon specific on the basis that the redir sub is working as expected.

aphalon:
OK... don't normally program in perl - so thought that _was_ the safe way to check it 😮

iliarostovtsev:
Yes, in Perl strict mode it would be different, and technically your implementation was correct but didn't consider 0 as a value.

PR #1278, opened 2020-08-07
Automatically broadcast for debian-linux at bootup configuration

macmon:
If you edit the bootup configuration of an interface, the automatic broadcast calculation will not work. Furthermore, Debian also supports the omission of the broadcast configuration.

An example of the first problem, the `/etc/network/interfaces` contain follow lines:

```
iface ens256 inet static
        address 192.168.2.2
        netmask 255.255.192.0
        broadcast 192.168.63.255
        network 192.168.0.0
```
Then change the netmask from 255.255.192.0 to 255.255.255.0, broadcast to automatic and save (on `Edit Bootup Interface`: https://192.168.2.2:10000/net/edit_bifc.cgi?idx=2)

After saving the file `/etc/network/interfaces` contains the following lines (with wrong broadcast). The broadcast should be `192.168.2.255`.

```
iface ens256 inet static
        address 192.168.2.2
        netmask 255.255.255.0
        broadcast 192.168.63.255
        network 192.168.2.0
```

But (Debian) Linux supports skipping the broadcast configuration and calculates this automatically. This patch supports skipping the broadcast configuration and leaves it to the operating system as far as possible. If the `/etc/network/interfaces` contain the following lines and you configure the bootup interface, you will see the following screenshot (broadcast calculated by Webmin is greyed out):

```
iface ens256 inet static
        address 192.168.2.2
        netmask 255.255.192.0
        network 192.168.2.0
```
grafik

### DIFF
--- a/net/debian-linux-lib.pl
+++ b/net/debian-linux-lib.pl
@@ -98,11 +98,6 @@ foreach $iface (@ifaces) {
 		$cfg->{'edit'} = ($cfg->{'name'} !~ /^ppp|lo/);
 		$cfg->{'index'} = scalar(@rv);	
 		$cfg->{'file'} = $network_interfaces_config;
-		if (!$cfg->{'broadcast'} &&
-		    $cfg->{'address'} && $cfg->{'netmask'}) {
-			$cfg->{'broadcast'} = &compute_broadcast(
-				$cfg->{'address'}, $cfg->{'netmask'});
-			}
 		push(@rv, $cfg);
 		}
 	elsif ($addrfam eq "inet6") {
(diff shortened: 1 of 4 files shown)


jcameron:
Some operating systems do need the broadcast address to be specified though.

Isn't it better for Webmin to correctly compute this from the IP and netmask?

macmon:
On operating systems where the broadcast address must be set, the operating system specific Webmin module can still calculate and set the broadcast address (see redhat-linux-lib.pl). But Webmin currently does not calculate the broadcast address for an existing interface anymore.

My merge-request supports both.

jcameron:
Ok, I see where you're going with this now. I ended up checking in another fix that basically does the same thing, but in a different way that's compatible across operating systems.

macmon:
Thanks for your commit, but you can't set to Automatic if there was manual before.

jcameron:
Thank - I just checked in another fix for that bug.

macmon:
Thanks for fixing.

PR #1284, opened 2020-08-15
LDAP Bind Credentials File Option for ldap-useradmin Webmin Module

ayoder770:
This PR includes an option to allow the ldap-useradmin (LDAP Users and Groups) Webmin module to be pointed to a system file to obtain the bind password instead of being maintained within the module configuration file. The motivation for this is on a system where the ldapsearch utility is used in concert with the "-y" parameter, the system file can be shared and the bind password only needs to be maintained in one location instead of two. Thank you for considering my request.


jcameron:
Thanks for the patch!

PR #1289, opened 2020-08-24
Wrong key printout in edit_zonekey.cgi

MarcHagen:
This fixes the wrong key printout in edit_zonekey.cgi page.
https://www.virtualmin.com/node/69750

keyrec->values consist of a array with:
0: KSK(257) or ZSK (256)
1: DNSSEC Protocol (Fixed to 3)
2: Algorithm identifier
3+: Public Key spliced in multiple chunks...

The public key in chunks is because webmin inserts the key into the zonefile with spaces in between:
https://github.com/webmin/webmin/blob/2d95efa30e587a40642288c9274930f40b2b13fb/bind8/bind8-lib.pl#L3335


iliarostovtsev:
Thanks for the patch. However, I don't think that's the right way to go. I'll comment on the parent, linked node.

PR #1299, opened 2020-09-16
Update README.md

zero77:
This quickly shows all package versions across different distros at a glance.
Also, Libraries.io dependency status for GitHub repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,7 @@
+[![Dependency](https://img.shields.io/librariesio/github/webmin/webmin)](https://libraries.io/github/webmin/webmin)
+[![Packages versions](https://repology.org/badge/latest-versions/webmin.svg)](https://repology.org/metapackage/webmin/versions)
+[![Packages](https://repology.org/badge/tiny-repos/webmin.svg)](https://repology.org/metapackage/webmin/versions)
+
 ## Contents
 * [Changelog](https://github.com/webmin/webmin/blob/master/CHANGELOG.md)
 * [About](#about)


swelljoe:
Adding libraries.io doesn't make sense. It's showing nonsensical npm dependencies.

chris001:
This would explain why, [in 2014-2015 a kind node dev published this `node` package `node-webmin` which made node devs able to install `webmin` directly with a Node Package Manager command!](https://libraries.io/npm/webmin)   Thus what you saw was their node-based installer package depends on a particular version of the `npm` command!  

I'd be in favor of having `Libraries.io` enabled so that it would show just `webmin`'s of actual dependencies and their versions, which the `webmin` installer in turn installs thru each supported linux (or bsd or...) distro's version's package manger (`apt`, `pacman`, `rpm`, `dnf`, etc.), e.g. if you install `webmin` on Fedora Linux latest version right now, then you'll get `perl` version 5.xxxx and `openssl` version xxx and `bind9` version 9.xxx etc), that'd helpful information and bring a significant time savings to have `Libraries.io` constantly update it every day and show it on the repo main page in the `README`.  Plus version incompatibilities between all the libraries, if any!  Often enough, users require a particular version of a system service which, for instance, has a feature they need to test before they could roll their critical `webmin` or `Virtualmin` server out to real users e.g. test `bind9` latest stable which has fully auto secure DNS key management renewing zone signing and zero-downtime key rolling by publishing the next key in advance of the current key expiration date, so they could see `Libraries.io` has already scanned all supported distro repos and shows which distros have that version of bind9 or newer, and pick the right distro to match their requirement immediately.

Issue #1296, opened 2020-09-11
webmin ver 1.955 cron malfunction


Mate20x:
Working fine the other day

Today I found a problem with the Cron schedule editor:

![2020-09-11_205442](https://user-images.githubusercontent.com/46883314/92928071-2e179280-f471-11ea-8515-35266e7fafef.jpg)

Another new window will pop up after clicking

Display blank:

![2020-09-11_205605](https://user-images.githubusercontent.com/46883314/92928215-5c956d80-f471-11ea-8f70-8cc4d2a8233f.jpg)

iliarostovtsev:
Hi,

I can see that the encoding is broken on the link in navigation menu. It is due to old errors in human translated files. I will fix this.

Speaking about initial issue - simply reload the page. It should fix your issue.

iliarostovtsev:
Okay, issues with chars have been fixed. https://github.com/webmin/webmin/commit/cc9689d204177c8a85978b538a4e08658d3d62ea

I will keep it open and see if I can reproduce the initial issue you've reported.

Do you remember doing anything before this issue popped up?

www3cc:
@iliaross 

I open webmin cron and enter the command wget command

as follows:
wget -o /root/notify.xxx.com/index.log -t 5 -S https://notify.xxx.com
wget -o /root/notify.xxx.com/meets.log -t 5 -S https://notify.xxx.com/meets

![2020-09-12_231817 - 複製](https://user-images.githubusercontent.com/46883314/92999013-398dbb00-f550-11ea-8595-ca4406cf77d7.jpg)

The next day,
 when webmin clicks the menu cron, 
it will automatically open a new window and display a blank screen.

![2020-09-11_205605](https://user-images.githubusercontent.com/46883314/92998971-d6038d80-f54f-11ea-9fba-20b173887449.jpg)

Try to use webmin to adjust English
 cron still shows a white screen
The current emergency solution can only add index.cgi to the browser URL
Join index.cgi emergency method
Will not see the left menu of webmin

![2020-09-12_231817](https://user-images.githubusercontent.com/46883314/92999066-9c7f5200-f550-11ea-9cba-85ca7d5fc781.jpg)



Thanks for your assistance

Looking forward to the next release of a new solution version

Please forgive me for my poor English

www3cc:
@iliaross 

When will the new version be released?

Webmin version> 1.955

Thank you

iliarostovtsev:
[...]

PR #1301, opened 2020-09-18
Prevent exec process writing into module folder https://github.com/webmin/webmin/issues/1296

iliarostovtsev:
Jamie, it looks like a security issue to me. This pull request should address it. I made a pull request, as I wasn't sure if it would affect any other modules in unexpected way? If there is a better way, please let me know.

..to reproduce an issue run with Cron:

~~~
wget -o /root/index.log -t 5 -S https://virtualmin.com
~~~

### DIFF
--- a/proc/proc-lib.pl
+++ b/proc/proc-lib.pl
@@ -103,16 +103,6 @@ elsif ($access{'uid'}) {
 	}
 }
 
-sub print_array
-{
-    print "Content-type: text/html\n\n";
-    my ($____v) = @_;
-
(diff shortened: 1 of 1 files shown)


jcameron (commented):
Is this always safe? There might be cases where the process being run expects to find a file in the current directory, or chdir() was already called.

Maybe change it to only chdir() if the current directory is under the webmin root.

iliarostovtsev:
> Maybe change it to only chdir() if the current directory is under the webmin root.

Okay, done. Thanks!

PR #1313, opened 2020-10-21
fixed pl.auto

mateusz-szafraniec:
fixed translation GiB means GiB in Polish (not "Klin")


jcameron:
@iliaross should we exclude these abbreviations from automatic translation files?

iliarostovtsev:
> .. should we exclude these abbreviations from automatic translation files?

Well, I think, no, as it depends on the language and it really should translate it correctly, it does, for example, for Russian. Automatic translations might have glitches and this is where human's help is needed.

@mateusz-szafraniec Thanks, Mateusz! We will fix it in other places too.

PR #1326, opened 2020-11-12
Use `encrypt_password` in the right context https://github.com/webmin…

iliarostovtsev:
Jaime, have a look. It's a bug t me.

### DIFF
--- a/changepass.pl
+++ b/changepass.pl
@@ -3,6 +3,8 @@
 # Script for the user to change their webmin password
 
 # Check command line arguments
+do "./miniserv.pl";
+require "./acl/md5-lib.pl";
 usage() if (@ARGV != 3);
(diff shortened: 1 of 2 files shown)


changepass.pl
```diff
@@ -3,6 +3,8 @@
 # Script for the user to change their webmin password
 
 # Check command line arguments
+do "./miniserv.pl";
```
jcameron, line 6: I don't think this is a good way of bringing in the needed functions from miniserv.pl, because it will execute al the main code! Best to split them into a separate file.

iliarostovtsev: Okay, understood! Thanks. Here is the final fix. https://github.com/webmin/webmin/commit/d1d475bbd14a294f04ebf4636b7057814eded9a7

PR #1359, opened 2020-12-23
Update fr

ljdj:


iliarostovtsev:
Thank you for all your translations!

PR #1385, opened 2020-12-24
Create fr

ljdj:


iliarostovtsev:
This one is also Identical to its `.auto` variant?

ljdj:
Of course it is identical, I indent it like the English file so that if future language updates, it is easier to find the line to translate

PR #1386, opened 2020-12-24
Update config.info.fr

ljdj:
Complete and indented translation in French for IPV4


ljdj (commented):
Complete translation in French

PR #1392, opened 2020-12-24
Update config.info

ljdj:
Removing lines from IPV4


ljdj:
error of file

iliarostovtsev:
Yes, correct, symlinks **_must not_** be replaced with actual files.

PR #1399, opened 2020-12-24
Create fr

ljdj:


iliarostovtsev:
This file is also identical to `iscsi-client/lang/fr.auto` - is this intended?

ljdj:
Of course it is identical, I indent it like the English file so that if future language updates, it is easier to find the line to translate

iliarostovtsev:
It's identical, because lines from `fr.auto` file with translations made by Google Translate, where put to `fr` file, which is meant for human translations only. Is this intended? Have you checked translations from `fr.auto` file?

swelljoe:
If all of these strings are verified correct by a human, then we want to merge it, but if this is just an exact copy of the auto-translated strings without human review, we don't want them to be in the human translated file (`fr`). Translations generated by machine should always be clearly identifiable (`fr.auto`). The quality standards of the human translations should be higher than the auto-translated files, i.e. corrected when the machine uses non-idiomatic language or confusing language.

Human translations can be one string at a time, and they will take precedence over the auto translations.

I'll close this for now, since there has been no follow up.

PR #1402, opened 2020-12-25
Update fr

ljdj:


iliarostovtsev:
Look, you should not just move lines from `fr.auto` to `fr` file without checking, improving them. There is no reason in doing it, as there is a [language manager script](https://github.com/webmin/webmin/blob/master/bin/language-manager) that can do all of this automatically.

ljdj:
I prefer to check and indented (as the original file) the auto.fr files and put them on a fr file.

iliarostovtsev:
I will not be able to merge that, unless .auto files are human verified, not just copy pasted.

Thanks.

ljdj:
Auto lines have been checked and modified depending on whether the translation was good or not

iliarostovtsev:
> Auto lines have been checked and modified depending on whether the translation was good or not

.. then we're good. Merci!

PR #1447, opened 2021-02-27
Fix for restore job locking when using SQLite

ctrlbru:
The SQLite db handle left open prevents the job from starting and permanently freezes "s dir" command in bconsole. Not tested with mysql and postgresql (but it shouldn't hurt)


jcameron:
Thanks!

Issue #1449, opened 2021-03-02
Unclear how to allow IPv6 range for IP Access Control


tomldickson:
On the IP Access Control page it shows how to add IPv4 ranges in the help, but trying to add an IPv6 range in the same manner as _hosts.allow_ gives an error:

    Failed to update access list : Failed to find IP address for '[2600:b00n:dcaf::]/48'

I would like to add the whole range as my client computer has privacy settings enabled and so changes it's IPv6 "outgoing" address every few hours.

Allowing a single IPv6 address does work - but I'm not sure how to put in a range.

jcameron:
You should be able to enter it without the [] , like `2600:b00n:dcaf::/48`

tomldickson:
Thank you! Created a pull request to update that description: #1451

tomldickson:
The pull request closes this issue and has been accepted.

PR #1451, opened 2021-03-04
Update access_desc to make it clear how to use IPv6

tomldickson:
To prevent others from encountering #1449 I've updated the translations for that page, the line access_desc to use the example IPv6 range defined by RFC3849.

I was (I think) able to update all but "fa" which appears to write right to left and confuses my editor.


iliarostovtsev:
Thanks!

PR #1452, opened 2021-03-06
Change: Do not list Legacy Theme in dropdown

iliarostovtsev:
I think listing Legacy Theme (Old Webmin Theme) must not be an option

### DIFF
--- a/acl/edit_user.cgi
+++ b/acl/edit_user.cgi
@@ -218,6 +218,7 @@ my @overlays = grep { $_->{'overlay'} } @all;
 if ($access{'theme'}) {
 	# Current theme
 	my @topts = ( );
+	push(@topts, !$user{'theme'} ? [ '', $text{'edit_themedef'} ] : ());
 	foreach my $t (@themes) {
 		push(@topts, [ $t->{'dir'}, $t->{'desc'} ]);
 		}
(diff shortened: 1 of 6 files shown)


jcameron:
What if somehow a user actually has the old theme selected though?

iliarostovtsev:
Then, by visiting one of those pages there will simply be no option for Old Theme, and to save the form a user would need to either select Gray or Authentic Theme.

iliarostovtsev:
Legacy Theme is a fall back solution only. We don't expect anyone selecting it.

jcameron:
In that case, at least set the 5th param of ui_select to 1 (the "missing" flag), so that the user doesn't get an unexpected theme change.

iliarostovtsev:
> In that case, at least set the 5th param of ui_select to 1 (the "missing" flag), so that the user doesn't get an unexpected theme change.

That wouldn't work as expected but this commit does it - https://github.com/webmin/webmin/pull/1452/commits/4882e980f6ddd949d9efb6bae41a9389c1e3cda5

The users with Legacy Theme selected will have it as an option. Other users won't. I think it's all ready.

jcameron:
Looks good, thanks!

PR #1453, opened 2021-03-06
Change: Dev/no list blue theme

iliarostovtsev:
Do not list symlinked themes properly

### DIFF
--- a/acl/edit_user.cgi
+++ b/acl/edit_user.cgi
@@ -211,7 +211,7 @@ if ($access{'cats'}) {
 			    [ 0, $text{'default'} ] ]));
 	}
 
-my @all = &webmin::list_themes();
+my @all = &webmin::list_visible_themes();
 my @themes = grep { !$_->{'overlay'} } @all;
 my @overlays = grep { $_->{'overlay'} } @all;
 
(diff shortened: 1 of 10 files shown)


cluster-usermin/cluster-usermin-lib.pl
```diff
@@ -113,7 +113,7 @@ sub all_themes
 {
 local %done;
 return sort { $a->{'desc'} cmp $b->{'desc'} }
-	grep { !$done{$_->{'dir'}}++ }
+	grep { ! -l "$root_directory/$_->{'dir'}" && !$done{$_->{'dir'}}++ }
```
jcameron, line 116: Is this function used anywhere other than to display a UI for the user? If so, the filtering should be done at the UI level, because a user might still have a linked theme selected.

acl/edit_user.cgi
```diff
@@ -211,7 +211,7 @@ if ($access{'cats'}) {
 			    [ 0, $text{'default'} ] ]));
 	}
 
-my @all = &webmin::list_themes();
+my @all = &webmin::list_visible_themes();
```
jcameron, line 214: Thanks, this is the right way of doing it.

cluster-usermin/cluster-usermin-lib.pl
iliarostovtsev: Yes, correct. Unrolled. 
https://github.com/webmin/webmin/pull/1453/commits/38deb9c1328e9a3be6f775811aa52a62267b5868

PR #1468, opened 2021-03-28
Replaced java applet with jsTree in Bacula restore files selector

ctrlbru:
This is an attempt to a long due update of the file selector for backup restores in the Bacula module. The java applet has been replaced with a javascript implementation of the file selector using [jsTree](https://www.jstree.com/).

It has been tested (not too much!) with Linux and Windows style filenames, Chrome 89 and Firefox 78, Grey Framed Theme and Authentic Theme.

I'm not experienced in the Webmin developement rules, sorry if I'm proposing something that cannot be accepted for any reason. I just did it for my personal use and wanted to share it.

**Changes:**
- treechooser.cgi: replaced java applet with jsTree control
- list.cgi: added optional JSON output, preserving old text output behavior in case it's still used by other parts of the module
- web-lib-funcs.pl: added blob to child-src in CSP
- bacula-backup-lib.pl: changed bacula_file_button to hijack Authentic Theme attempt to replace popup with modal, since it doesn't work for the Bacula module ("undefined" message)
- added jstree folder containing minimized jsTree and jQuery


web-lib-funcs.pl
```diff
@@ -958,7 +958,7 @@ sub PrintHeader
 	print "X-Frame-Options: SAMEORIGIN\n";
 	}
 if (!$gconfig{'no_content_security_policy'}) {
-	print "Content-Security-Policy: script-src 'self' 'unsafe-inline' 'unsafe-eval'; frame-src 'self'; child-src 'self'\n";
+	print "Content-Security-Policy: script-src 'self' 'unsafe-inline' 'unsafe-eval'; frame-src 'self'; child-src 'self' blob:\n";
```
jcameron, line 961: What does "blob" do here exactly?

bacula-backup/bacula-backup-lib.pl
```diff
@@ -417,9 +417,10 @@ sub directive_lines
 
 # bacula_file_button(filesfield, [jobfield], [volume])
 # Pops up a window for selecting multiple files, using a tree-like view
+# Brackets around "window" are to hijack Authentic Theme replacement of popups with modals ("undefined")
 sub bacula_file_button
 {
-return "<input type=button onClick='ifield = form.$_[0]; jfield = form.$_[1]; chooser = window.open(\"treechooser.cgi?volume=".&urlize($_[2])."&files=\"+escape(ifield.value)+\"&job=\"+escape(jfield.value), \"chooser\", \"toolbar=no,menubar=no,scrollbar=no,width=500,height=400\"); chooser.ifield = ifield; window.ifield = ifield' value=\"...\">\n";
+return "<input type=button onClick='ifield = form.$_[0]; jfield = form.$_[1]; chooser = (window).open(\"treechooser.cgi?volume=".&urlize($_[2])."&files=\"+escape(ifield.value)+\"&job=\"+escape(jfield.value), \"chooser\", \"toolbar=no,menubar=no,scrollbar=no,width=500,height=400\"); chooser.ifield = ifield; window.ifield = ifield' value=\"...\">\n";
```
jcameron, line 423: So this is to prevent the authentic theme from hijacking the popup? 

@iliarostovtsev is there a better way to do this?

web-lib-funcs.pl
ctrlbru: This is needed to use inline js workers when tree data is fetched, but it might be a security issue and I found there should be a more strict way to allow it only for the specific script, so let me check. I'll report as soon as I find time to work on it again.

ctrlbru:
Made a bit of a mess because git is not my everyday vcs, but in the end web-lib-funcs.pl requires no change, blob has been removed.

jcameron:
Thanks for the patch!

PR #1469, opened 2021-04-01
Variable $tostr used 100 lines above its declaration !

nikokio:
Hello,

When using fastrpc, it hangs because variable $tostr is used before its declaration


jcameron:
Thanks, I don't know how I missed this!

nikokio:
You're welcome. I'm sorry, I know this bug since august 2019, but I was busy at this time and I also was not using git at all...
Things have changed since then and I recently wanted to upgrade, so here we are :)

Issue #1421, opened 2021-01-01
Can MySQL module change labels to MariaDB when that database is used?


swelljoe:
Jamie, would it be feasible/reasonable and not messy code-wise to make the MySQL module display `MariaDB` instead of `MySQL`, just to eliminate confusion among users about which database is installed and in use?

They're close enough to interchangeable for our purposes most of the time, but we still get folks who are confused about what database they have, sometimes. (Though maybe changing the label will make more people think they need to "upgrade" to MySQL...folks will always find a way to make their lives harder.)

I don't see an obvious way to do it, since it's all generated with variables from the text hash, but you're much more clever than I, so maybe you see an easy/clean way to do it.

jcameron:
For the text within the MySQL module, it wouldn't be too hard to do this. The only potentially hard part would be the module title on the left menu.

Are you also thinking we should do this in the Virtualmin UI?

swelljoe:
Everywhere that MySQL currently appears on a system that actually has MariaDB. I wouldn't want you to spend a lot of time on it, though. It's a minor thing that I think documentation can also address. It's just been a source of occasional confusion...but, I think it's fine to kick it down the road for later, if it's at all difficult.

I think some of our users don't really understand that MySQL and MariaDB are effectively interchangeable. and having the label say MySQL when it's really MariaDB confuses some of them. (But, then again, those same users might be confused into trying to install MySQL if they see they "only" have "MariaDB" and know that their apps need "MySQL".)

This may just not be a problem with a technical solution, I'm not sure! It feels righter to call it what it is, but it probably won't actually solve the problem of user confusion.

iliarostovtsev:
> For the text within the MySQL module, it wouldn't be too hard to do this. 

Yes, how about [this](https://github.com/webmin/webmin/commit/371be571af50c45b1cbe654d239d3aaf2ae9f39e)? 

.. but we also have _config_, _module.config_ and _help_ to deal with. It can be also pretty easily doable but I am not sure about _not messy code-wise_, as Joe mentioned before. Jamie, how about modifying `get_module_info()` and `help_file` subs?

jcameron:
@iliaross yes something like that, however we'd want to aggressively cache the variant of MySQL in use so that we don't have to hit the database everytime the `mysql` module is loaded.

Ironically, I often see people confused in the other direction - they still refer to "MySQL" when it's actually MariaDB!

iliarostovtsev:
[...]

PR #1470, opened 2021-04-06
Add ability to override module's info #1421

iliarostovtsev:
https://github.com/virtualmin/virtualmin-gpl/issues/275
https://github.com/webmin/webmin/issues/1421#issuecomment-772137324

### DIFF
--- a/web-lib-funcs.pl
+++ b/web-lib-funcs.pl
@@ -5644,7 +5644,7 @@ if (-r "$mdir/$call.pl") {
 	eval {
 		local $main::error_must_die = 1;
 		&foreign_require($mod, "$call.pl");
-		&foreign_call($mod, $call, \%{$data});
+		&foreign_call($mod, $call, $data);
 		};
 	}
 }
--- a/web-lib-funcs.pl
+++ b/web-lib-funcs.pl
@@ -5639,20 +5639,13 @@ return if (!$mod);
 my $mdir = &module_root_directory($mod);
 
 # Call module specific overrides
-if (-r "$mdir/module.overrides") {
-	my %overrides;
-	&read_file_cached("$mdir/module.overrides", \%overrides);
-	my $funcs = $overrides{'funcs'};
-	if ($funcs) {
-		eval {
-			local $main::error_must_die = 1;
-			my @funcs = split(/\s+/, $funcs);
-			&foreign_require($mod, "$mod-lib.pl");
-			foreach my $func (@funcs) {
-				&foreign_call($mod, $func, \%{$data});
-				}
-			};
-		}
+my $call = 'module_overrides';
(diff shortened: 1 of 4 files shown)


web-lib-funcs.pl
```diff
@@ -5639,19 +5639,19 @@ sub get_module_overrides
...
+if (-r "$mdir/module.overrides") {
 	my %overrides;
 	&read_file_cached("$mdir/module.overrides", \%overrides);
 	my $funcs = $overrides{'funcs'};
 	if ($funcs) {
 		eval {
 			local $main::error_must_die = 1;
 			my @funcs = split(/\s+/, $funcs);
-			&foreign_require($mod);
+			&foreign_require($mod, "$mod-lib.pl");
```
iliarostovtsev, line 5650: This is an important fix as we need to define module's lib manually as `foreign_require` will call `get_module_info` in auto detection mode causing endless loop.

We should assume any module's overrides sub reside in main module's lib file.

jcameron:
I like this solution, but it could be simplified by instead of having the override function be defined in `module.overrides` and implemented in the main -lib.pl file, do it in a separate file. Sort of like how modules have a `backup_config.pl` file already.

iliarostovtsev:
Okay, simplification is done. https://github.com/webmin/webmin/pull/1470/files

web-lib-funcs.pl
```diff
@@ -5622,6 +5625,30 @@ sub get_module_info
...
+
+my $mdir = &module_root_directory($mod);
+
+# Call module specific overrides
+my $call = 'module_overrides';
+if (-r "$mdir/$call.pl") {
+	eval {
+		local $main::error_must_die = 1;
+		&foreign_require($mod, "$call.pl");
+		&foreign_call($mod, $call, \%{$data});
```
jcameron, line 5647: Why \%{$data} rather than just $data ?

iliarostovtsev: It wouldn't alter the data, if not passed as reference?

jcameron: Yes, but isn't the whole idea to alter the data?

iliarostovtsev: 🙂 Yes, the whole point is to alter the data and passing it as `$data` the second time wouldn't work. No?

jcameron: Since $data is a hash reference, you can alter the contents by just passing it as $data

iliarostovtsev: Actually you're right! Fixed! Thanks for the review.

jcameron:
Looks good, thanks!

iliarostovtsev:
Cool, thanks.

PR #1471, opened 2021-04-08
Add ability for module to call custom sub upon ACLs save

iliarostovtsev:
Reviewing Usermin support for utilizing ACLs set for a user in Webmin User module to, for example, allow custom paths.

### DIFF
--- a/filemin/acl_security.pl
+++ b/filemin/acl_security.pl
@@ -30,6 +30,13 @@ sub acl_security_form {
     print &ui_table_row($text{'acl_max'},
 	&ui_opt_textbox("max", $access->{'max'}, 10, $text{'acl_unlimited'}).
 	" ".$text{'acl_bytes'}, 3);
+
+    # Update for Usermin
+    if (&foreign_installed("usermin")) {
+        print &ui_table_hr();
+        print &ui_table_row($text{'acl_update_usermin'},
+            &ui_yesno_radio('save_usermin_acls', 0)."<br><br>");
+        }
 }
 
 sub acl_security_save {
@@ -59,7 +66,9 @@ sub acl_security_save {
         $access->{'work_as_user'} = $in->{'acl_user'};
     }
     $access->{'max'} = $in->{'max_def'} ? undef : $in{'max'};
-    $access->{'_module_acl_func'} = 'save_module_acl_usermin';
+
+    $access->{'_module_acl_func'} = 'save_module_acl_usermin'
+        if ($in{'save_usermin_acls'});
 }
 
 sub acl_security_noconfig {
--- a/filemin/acl_security.pl
+++ b/filemin/acl_security.pl
@@ -59,6 +59,7 @@ sub acl_security_save {
         $access->{'work_as_user'} = $in->{'acl_user'};
     }
     $access->{'max'} = $in->{'max_def'} ? undef : $in{'max'};
+    $access->{'_module_acl_func'} = 'save_module_acl_usermin';
 }
 
 sub acl_security_noconfig {
--- a/filemin/acl_security.pl
+++ b/filemin/acl_security.pl
@@ -66,9 +66,6 @@ sub acl_security_save {
         $access->{'work_as_user'} = $in->{'acl_user'};
     }
     $access->{'max'} = $in->{'max_def'} ? undef : $in{'max'};
-
-    $access->{'_module_acl_func'} = 'save_module_acl_usermin'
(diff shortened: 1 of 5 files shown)


jcameron:
This seems complex ... what problem are you trying to solve here?

iliarostovtsev:
The ACLs saved for the given user can be chosen to be saved for `/etc/usermin/filemin/$user.acl`, letting set the path.

iliarostovtsev:
It technically can be any module if the sub is defined.

iliarostovtsev:
https://forum.virtualmin.com/t/allow-usermin-user-to-browse-outside-of/110245/2?u=ilia

iliarostovtsev:
Okay, Jamie. This one is also simplified - https://github.com/webmin/webmin/pull/1471/commits/246f274822c428bd4756ddbd92e6fd94d83476a5

jcameron:
OK, so the goal is in Usermin to be able to configure the allowed paths on a per-user basis?

iliarostovtsev:
Yes, allowed paths and other per user options, like max upload.

iliarostovtsev:
> OK, so the goal is in Usermin to be able to configure the allowed paths on a per-user basis?

Yes!

I have just added patches to provide complete support for File Manager tree and listing when allowed path doesn't include just root level directory.

For example, it will be possible to set now for any user:


~~~
$HOME
/var/lib/mysql
/tmp/.webmin
~~~ 

.. and get correct and only allowed paths on both the tree and listing panel.

https://github.com/authentic-theme/authentic-theme/compare/19.74...master#diff-60f74f044d7e19bfac1f45dc6373aa563b2cc6abf587e2d306c3f8410785c211

jcameron:
I think we need a more comprehensive solution that potentially allows module ACLs to be set for users who aren't also in Webmin. Something that can define permissions or configs that apply to entire Unix groups or lists of Unix users for example..

iliarostovtsev:
> Something that can define permissions or configs that apply to entire Unix groups or lists of Unix users for example..

What is wrong in having control over ACLs per each user the way it is intended by Webmin Users module?

> I think we need a more comprehensive solution that potentially allows module ACLs to be set for users who aren't also in Webmin.

Do you think it could be an extra solution while keeping this as well?

What is wrong in letting an admin to control ACLs for a user if needed? This is a very rare case but still could be userful?

iliarostovtsev:
> Something that can define permissions or configs that apply to entire Unix groups or lists of Unix users for example..

By the way, Jamie, I think we don't need to make it more complicated than it already is, because filesystem permissions will do the rest of the job!

jcameron:
In Usermin, unix permission still apply and always did. But I assume what's wanted here is a way to lock down visibility of directories even when unix permissions would allow access?

iliarostovtsev:
> In Usermin, unix permission still apply and always did. But I assume what's wanted here is a way to lock down visibility of directories even when unix permissions would allow access?

The thing is, that I have made patches that are going to 19.75, which will work perfectly fine with ACLs (allowed paths) having sub-directories like:

~~~
/var/lib/userdir
/etc/apache/userdir
~~~

.. it will display a file tree and file listing only for given directories, excluding all others. Isn't this enough already? I cannot imagine what else would be needed for more granular control? Is this what you meant in your comment?

iliarostovtsev:
Jamie, please try latest theme patches and see if that's something you wanted? If something is missing, let me know please.

jcameron:
Which theme patch are you referring to? Is it needed for this Usermin change to work?

iliarostovtsev:
> Which theme patch are you referring to?

I talk about these patches [<sup>patch-1</sup>](https://github.com/authentic-theme/authentic-theme/commit/bd431a0422d9cf2cc0faacca4d326b7f5adb4f34) [<sup>patch-2</sup>](https://github.com/authentic-theme/authentic-theme/commit/a78abef1cbcb7cd1864f2fbf5475f5885de6164c).

> Is it needed for this Usermin change to work?

It's not required but it makes things work right, when the path is not just one of the root level directories, i.e. `/etc/apache2` or `/home/user/path/allowed_path`. The permissions will still apply but it's now possible to set any paths on user _.acl_ file and make things work accordingly.

iliarostovtsev:
Jamie, where we at with this? :slightly_smiling_face:

jcameron:
Didn't I fix the original issue in another patch? So this shouldn't be needed anymore ..

iliarostovtsev:
I don't remember already. If you did - great! Thanks.

PR #1477, opened 2021-04-20
Use new API with old `changepass.pl` for compatibility

iliarostovtsev:
Jamie, I suggest completely dropping old `changepass.pl` code and use new API in it:

```Bash
# webmin passwd  --help
Usage:
    passwd [options]

Options:
    --help, -h
        Print this usage summary and exit.

        Examples of usage:

         - passwd --user root
 
         - passwd --user root --password ycwyMQRVAZY
 
         - passwd --config /usr/local/etc/webmin --user root --password ycwyMQRVAZY

    --config, -c
        Specify the full path to the Webmin configuration directory.
        Defaults to "/etc/webmin"

    --user, -u
        Existing Webmin user to change password for

    --password, -p
        Set new user password. Using this option may be unsecure.
```

New API can be seen here - https://github.com/webmin/webmin/blob/master/bin/passwd

### DIFF
--- a/changepass.pl
+++ b/changepass.pl
@@ -18,7 +18,7 @@ if ($status != 0) {
 	else {
 		print "Error: $!\n";		
 		}
-}
+	}
 exit $status;
 
 sub usage
--- a/changepass.pl
+++ b/changepass.pl
@@ -3,109 +3,11 @@
 # Script for the user to change their webmin password
 
 # Check command line arguments
-require "./acl/md5-lib.pl";
 usage() if (@ARGV != 3);
-($config, $user, $pass) = @ARGV;
-if (!-d $config) {
-	print STDERR "The config directory $config does not exist\n";
-	exit 2;
-	}
-if (!open(CONF, "<$config/miniserv.conf")) {
-	print STDERR "Failed to open $config/miniserv.conf : $!\n";
-	print STDERR "Maybe $config is not the Webmin config directory.\n";
-	exit 3;
-	}
-while(<CONF>) {
-	if (/^([^=]+)=(\S+)/) { $config{$1} = $2; }
-	}
-close(CONF);
-if (!open(CONF, "<$config/config")) {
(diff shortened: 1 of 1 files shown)


iliarostovtsev:
image


LInk - https://github.com/webmin/webmin/commit/41b15fb1010ebc561163098aa81e6121ae4392c7

changepass.pl
```diff
@@ -3,109 +3,11 @@
...
-local ($pass, $salt) = @_;
-if ($use_perl_crypt) {
-	return Crypt::UnixCrypt::crypt($pass, $salt);
-	}
-else {
-	return crypt($pass, $salt);
-	}
-}
+my ($config, $user, $pass) = @ARGV;
+my $status = system("webmin passwd --config $config --user $user --pass $pass");
```
jcameron, line 9: Is the `webmin` command guaranteed to be in the path? Maybe use /bin/webmin , or some other known correct location.

iliarostovtsev:
> Is the webmin command guaranteed to be in the path? Maybe use /bin/webmin , or some other known correct location.

Yes, great suggestion. Fixed on the commit above.

changepass.pl
jcameron: What about systems installed using setup.sh? I don't think they have the /usr/bin/webmin command.

iliarostovtsev: How about this - https://github.com/webmin/webmin/pull/1477/commits/3b8861b1c92eda5c78f53735db5ac3b783b63048

jcameron: Thanks, that seems better

swelljoe:
In UNIX, commands that do what you told them should be silent, unless told to be verbose. The "Success" stuff is unnecessary noise.

iliarostovtsev:
I can remove "Success:" part, thanks for pointing out.

iliarostovtsev:
@swelljoe FIxed, alongside with few other improvements.

PR #1478, opened 2021-04-22
Fixed Simplified Chinese translation issues

kasuganosoras:
Updated Simplified Chinese translation.


iliarostovtsev:
Thank you very much! For the future PRs (if you do any) all human translated strings must be put to `zh` file, while `zh.auto` file be deleted.

kasuganosoras:
Thanks. I see. : )

iliarostovtsev:
@kasuganosoras Meiko, this should be the right way of doing it - https://github.com/webmin/webmin/commit/5dc2ec2b46273822092412a677625be67d365154

kasuganosoras:
ok, I will do this in the next localization update.

PR #1480, opened 2021-04-27
New setting to be compatible with external services monitoring cron jobs and similar periodic processes

ypapouin:
The purpose of this PR is to make the Webmin Scheduled Monitoring compatible with external services like https://healthchecks.io or https://deadmanssnitch.com/

A new option `As long as the service is up` is added: 

image

With this option, a `ping` is sent every minutes to say _"Yes I'm still here"_.

image

The big advantage of this logic is to be sure to receive a failure notification if no  _"Yes I'm still here"_ message arrives after few minutes.


jcameron:
Thanks for the patch!

PR #1496, opened 2021-06-16
Don't require being run as root unless base_dir is /usr/src/redhat

lanurmi:
If base_dir is under $HOME, no root permissions are required.


jcameron:
Thanks!

PR #1529, opened 2021-09-28
fix wrong translation

gueno:
Pâte is pasta(nouilles/nuddles)  in french ʘ‿ʘ


gueno:
do not forget cheez

jcameron:
Thanks!

PR #1534, opened 2021-10-05
R3 and E1: consider all active certificates

iliarostovtsev:


iliarostovtsev:
Do we even need this download code to exist nowadays?

jcameron:
It's a fallback - but I suppose we could clean up a bit to only downloading these certs if no CA cert is included in the acme_tiny response.

PR #1535, opened 2021-10-09
Update /webmin japanese translation.

niyari:
Add rel and target attribute to Webmin links : rel="noopener noreferrer" target="_blank"
line 152:ui_rgb Changed the text color of "RRGGBB" to stand out. (gray-theme)
line 206:mods_standard Radio button labels will not be generated if there is an a tag at the beginning.


iliarostovtsev:
Arigato gozaimasu!

PR #1548, opened 2021-11-21
Fix to copy links first, no matter if dir or file

iliarostovtsev:
It's important for File Manager to use `copy_source_dest($src, $dest)` with symlinked dirs and actually copy the link and not resolve it. This change does exactly this.

Works for me.


jcameron:
Thanks for the patch!

PR #1588, opened 2022-01-27
Add support for custom redirect URL for Usermin

iliarostovtsev:
It should be possible to define redirect URL for users behind proxies (like from **Virtualmin / Edit Users: Edit User**).

### DIFF
--- a/usermin/usermin-lib.pl
+++ b/usermin/usermin-lib.pl
@@ -954,7 +954,7 @@ my $cookie = "$sidname=$sid; path=/$sec";
 
 # Work out redirect host
 my @sockets = &webmin::get_miniserv_sockets(\%miniserv);
-my ($host, $port);
+my ($host, $port, $url);
 my %uconfig;
 &get_usermin_config(\%uconfig);
 if ($uconfig{'host'}) {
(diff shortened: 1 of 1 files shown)


jcameron:
Where does `redirect_url` that this patch uses get set?

iliarostovtsev:
Nowhere - for now, by editing manually only. That was the question I later asked on the initial support request ticket.

Where do you think it could be added to?

jcameron:
Oh I see. Wouldn't it make more sense for this to be an option in /etc/usermin/config , as it's not really related to miniserv?

iliarostovtsev:
I assumed that as we already have `redirect_host`, `redirect_port` and etc placed in `miniserv.conf`, it would be more intuitive to do the same for `redirect_url`. Yet, do you wish that go to `config`?

jcameron:
Ok, that makes sense

iliarostovtsev:
So now, would that be possible to make all those **`redirect_`** configurable on UI somewhere, to make it more straight forward to use, as now you need to look at the code or search for some deeply buried forum posts?

jcameron:
Yes, I think adding a UI for this would be a good idea ... somewhere at Webmin -> Usermin Configuration.

iliarostovtsev:
What about Webmin? Could you be more specific where in Webmin/Usermin Configuration it should be added?

jcameron:
I was thinking on the **Web Server Options** page?

iliajie:
> I was thinking on the **Web Server Options** page?

Jamie, please have a look, as this has been implemented on this commit - https://github.com/webmin/webmin/commit/086bd55ddd22a35a1e8344cbe22dde73d0ed48f5.

Almost all looks good to me, and the only thing I'm not certain about is text labels as **Internal redirect URL overrides** and **Post-login switch redirect URL** (only for **Usermin Configuration: Web Server Options**):

image

Feel free to reword those labels.

jcameron:
This looks fine to me - thanks!

PR #1589, opened 2022-01-27
LetsEncrypt Staging URL update

bctrainers:
LetsEncrypt staging URL changed per https://letsencrypt.org/docs/staging-environment/

The old URL `acme-staging.api.letsencrypt.org` in `letsencrypt-lib.pl` is showing as NXDOMAIN.


jcameron:
Thanks for the patch!

PR #1621, opened 2022-04-09
Add support for displaying service ports/protocols

iliajie:
This improvement will make it clear what ports and protocols are implied for each service.

**Example:**

image

### DIFF
--- a/firewalld/firewalld-lib.pl
+++ b/firewalld/firewalld-lib.pl
@@ -103,7 +103,7 @@ if (-r $service_file) {
 	}
 @ports = &unique(@ports);
 @protos = &unique(@protos);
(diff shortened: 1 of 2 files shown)


iliajie:
It will look like this when editing/adding a service. It will also be possible to quickly filter/search for needed port with a dropdown:

image

**Listing with for services with multi ports:**
image

jcameron:
Looks good, but can you change the main page UI to format the port like `ssh (22)` instead of `ssh:22` ?

iliajie:
> Looks good, but can you change the main page UI to format the port like `ssh (22)` instead of `ssh:22` ?

Yeah, why not. Hold on..

iliajie:
Done.

jcameron:
Thanks for the patch!

PR #1628, opened 2022-04-19
Dev/postinstall procedures

iliajie:
Fixes all STDOUTs to be void or printed to logs when necessary.


init/atboot.pl
```diff
@@ -7,6 +7,8 @@
 $product = $ARGV[0] || "webmin";
 $ucproduct = ucfirst($product);
 
+$< == 0 || die "updateboot.pl must be run as root";
```
jcameron, line 10: Shouldn't this be atboot.pl ?

init/atboot.pl
```diff
@@ -117,8 +119,22 @@
 	$perl_path = &get_perl_path();
 	&enable_at_boot($product, $ucproduct, $perl_path." ".&quote_path("$root_directory/miniserv.pl")." ".&quote_path("$config_directory/miniserv.conf"));
 	}
-elsif ($init_mode eq "rc" || $init_mode eq "upstart" ||
-       $init_mode eq "systemd") {
+elsif ($init_mode eq "systemd") {
```
jcameron, line 122: What's going to happen when atboot.pl is run a second time when Webmin is upgraded? Will we end up with an init script and a systemd script?

init/uninstall.pl
```diff
@@ -59,6 +59,10 @@ sub module_uninstall
 	# Delete win32 service
 	&delete_win32_service($product);
 	}
+elsif ($init_mode eq "systemd") {
```
jcameron, line 62: Similarly, what if the user had an init script before?

makedebian.pl
```diff
@@ -265,19 +261,15 @@
 print SCRIPT <<EOF;
 #!/bin/sh
 
-# Fix old versions of Webmin that might kill the UI process on upgrade
-if [ -d /etc/webmin ]; then
+# Fix old versions of Webmin that might kill the UI
```
jcameron, line 264: I deleted this whole block in a commit just now - please merge it.

makedebian.pl
```diff
@@ -61,7 +60,6 @@
 chmod(0755, $tmp_dir);
 mkdir($debian_dir, 0755);
 system("mkdir -p $pam_dir");
-system("mkdir -p $init_dir");
```
jcameron, line 64: I assume the plan here is to not include the init script in the RPM at all, and instead create it dynamically in atboot.pl ? If so, how will this behave when a user upgrades?

makedebian.pl
```diff
@@ -331,28 +323,24 @@
 	fi
 fi
 rm -f /var/lock/subsys/$baseproduct
-which systemctl >/dev/null 2>&1 && systemctl daemon-reload
```
jcameron, line 334: Do we no longer need this?

setup.pl
```diff
@@ -533,23 +537,132 @@
...
+
+	# Clear existing
+	my $csupp = sub {
+		unlink("$config_directory/start");
+		unlink("$config_directory/stop");
+		unlink("$config_directory/restart");
+		unlink("$config_directory/force-reload");
+		unlink("$config_directory/reload");
+	};
+	&$csupp();
```
jcameron, line 605: Why put this into a sub instead of just executing the statements?

setup.pl
```diff
@@ -533,23 +537,132 @@
...
+	symlink("$config_directory/stop-init", "$config_directory/stop");
+	# Restart init.d
+	symlink("$config_directory/restart-init", "$config_directory/restart");
+	# Force reload init.d
+	symlink("$config_directory/force-reload-init", "$config_directory/force-reload");
+	# Reload init.d
+	symlink("$config_directory/reload-init", "$config_directory/reload");
+
+	# For systemd
+	my $systemctlcmd = `sh -c 'command -v systemctl'`;
```
jcameron, line 620: Are we sure the `command` command always exists?

setup.sh
```diff
@@ -601,53 +602,116 @@ if [ "$noperlpath" = "" ]; then
...
 echo "..done"
 echo ""
 
+# Re-generating supplementary
+
+# Clear existing
+csupp () {
+	rm -f $config_dir/stop $config_dir/start $config_dir/restart $config_dir/force-reload $config_dir/reload
+}
+csupp
```
jcameron, line 670: Why does this need to be a sub? Also, is this syntax correct?

init/atboot.pl
iliajie: Yes, it's for `atboot.pl` but I forgot to change the name in die.

Fixed.

init/atboot.pl
iliajie: > Will we end up with an init script and a systemd script?

Not for **.rpm** and not for **.deb** packages though. But I missed the fact it may be install from source. What do we do for this? Should we also try to remove init script?

init/uninstall.pl
iliajie: Similarly, should we uninstall for both?

makedebian.pl
iliajie: Done - merged `makedebian.pl` only.

makedebian.pl
iliajie: Yes, this is the plan. If user upgrades it will be clear by the package. We don't need to clean it manually.

makedebian.pl
iliajie: No, as `systemd` must and will be restarted right after `webmin.service` file changes (modified, deleted or added).

setup.pl
iliajie: Because there are two places where you need to call exact the same code. If we ever again change this code we would need to remember to change it two times, in this case only once.

setup.pl
iliajie: Yes, it must be available on all systems, at least on all those systems which have `systemd`.

setup.sh
iliajie: The same as above, we call it twice.

This is Bash, why wouldn't syntax be incorrect?

init/atboot.pl
iliajie: Actually, even if installed from source there would be no _init script_ left on `systemd` systems as there is no one installed. It was only done by RPM and DEB and those are auto cleaned, right?

init/uninstall.pl
iliajie: Similarly, appears to be not a problem.

init/atboot.pl
jcameron: Oh, we don't need to support upgrading from source to RPM / DEB. However, we do need to support upgrading from an older RPM that included the init script to an RPM with these changes.

init/uninstall.pl
jcameron: The best option would be to to search for all kinds of init scripts, and remove whichever one the user has.

makedebian.pl
jcameron: I also updated makerpm.pl

makedebian.pl
jcameron: Got it! Yeah, that makes sense ..

setup.pl
jcameron: AH, ok .. makes sense.

setup.pl
jcameron: I mean, can we be sure it's available on other Unix variants like Solaris or FreeBSD?

setup.sh
jcameron: It's not safe to use bash-specific syntax - users might run setup.sh on systems like FreeBSD that have the older Bourne shell. Instead, just repeat this line.

init/atboot.pl
iliajie: Yes, when I meant an upgrade, I meant upgrading when installed from source.

init/uninstall.pl
iliajie: Jamie, main old `/etc/init.d/webmin` script will be removed (cleared) automatically upon package upgrade (i.e. 1.990->1.991).

makedebian.pl
iliajie: Yes, I merged that as well.

setup.pl
iliajie: Can't we? 🙂 I really don't know about Solaris.

setup.sh
iliajie: Okay, but if this is holding the merge and a release just go ahead and do it.

jcameron: I mean we couldn't do a release with this code as-is, as it would break non-Linux systems.

setup.pl
jcameron: The only command I think you can count on being available is `which`

makedebian.pl
jcameron: Thanks!

makedebian.pl
jcameron: Excellent, thanks

init/uninstall.pl
jcameron: Ok, cool. Is there any possible conflict if this file is removed upon upgrade, but then re-created by the postinstall script?

init/atboot.pl
jcameron: Ok, great

setup.pl
iliajie: I'll fix that!

init/uninstall.pl
iliajie: > Is there any possible conflict if this file is removed upon upgrade, but then re-created by the postinstall script?

Not that I'm aware of! I don't see a way for `init.d/webmin` to be recreated with existing code found (in PR). Do you?

iliajie:
I have made another change, that will not let a user to call directly `start-init`, `stop-init` and etc scripts (by adding a dot before the name) - https://github.com/webmin/webmin/pull/1628/files/e84bb09b8124fce1d6ddca38ccd3c731de006796..d865737a4599df0f3e92504ab242e34818e8e23f

Also, fixed `atboot.pl` to call those new scripts directly.

jcameron:
What's the benefit of changing the script name to add a dot at the start?

setup.sh
jcameron: Just run the `rm` statement directly where you're currently calling `csupp`

iliajie:
We have the scripts (commands in `/etc/webmin`) that are meant to be executed by the user, like `/etc/webmin/restart`, `/etc/webmin/start` or `/etc/webmin/stop` and those that must _never_ be executed by the user, like `/etc/webmin/.start-init` or `/etc/webmin/.stop-init`. By saying _never_ I refer primarily to `systemd` systems.

For example, current version of `systemd` (239) on RHEL 8 (and all derivatives) makes a lot of mess, if Webmin is handled outside of `systemd`. I cannot speak for other initialization systems but for `systemd` that's the case for version 239. This is directly related to why we had stuck Webmin process back in time when CentOS 8 was just released. However, it may well be `systemd` 239 bug, as `systemd 245` and `systemd 249` on Ubuntu 20.04 and Ubuntu 22.04 respectively don't have such issue.

Either way, we needed a proper `system.service` unit which was added with this PR. The key element of the new service unit was to allow `miniserv` to escape the service manager's lifecycle done by `systemd` using `KillMode=none` option to let Webmin decide how (on which signal) and when the service is reloaded or stopped. This allows properly restarting Webmin from UI as well.

Also a new helpful feature was added, which is to print to `STDOUT` a message, if Webmin cannot be restarted properly upon upgrade, which will look explicitly clear about what is wrong and what should be done:

```
warning: Webmin server cannot be restarted. It is advised to restart it manually by
running "/etc/webmin/restart-by-force-kill" when upgrade process is finished
```

And to summarize, we must hide all those scripts that user must not use directly and expose only those that can be used, which is:

```
/etc/webmin/stop
/etc/webmin/start
/etc/webmin/reload
/etc/webmin/restart
/etc/webmin/restart-by-force-kill
```

jcameron:
Ok, got it. So are scripts like `/etc/webmin/.start-init` just called by `systemd` ?

iliajie:
Yes, those "dot" scripts are only meant to be called by programs.

jcameron:
So I just did some extensive testing of this new feature, including upgrades from old versions, and it worked great!

jcameron:
Wait ... I ran into an issue with a tar.gz install. It didn't create the /etc/webmin/start and /etc/webmin/stop scripts correctly - should they be running `systemctl` ?

iliajie:
What do you mean by "didn't create start and stop scripts correctly" - what was on the inside of those scripts? Also, did the timestamp change for those files?

jcameron:
I mean the start/stop scripts just looked like they did in older releases - they don't run `systemctl`

iliajie:
Which OS and version was it?

Were those start/stop scripts symlinks or files?

jcameron:
It was CentOS 7. And yes, those files are just links to `/etc/webmin/.start-init` and `.stop-init`

iliajie:
Are you sure that it was CentOS 7, as it supposed to be `systemd` system, right?

So to reproduce an issue, I could install previous version of Webmin from the source on CentOS 7 and then upgrade it (Webmin) using what? `./setup.pl` script with the latest code?

iliajie:
Jamie, sorry. I really ran few actual tests and on CentOS 7 I can see that even upgrading using installation from source worked just as expected. No symlinks for start/stop scripts and `systemctl start/stop webmin` commands are in the those files. `systemd` _webmin.service_ file is also updated. 

If you can reproduce the issue again, please let me know.

jcameron:
Did you install with setup.sh , or setup.pl ?

iliajie:
I think I installed using .sh and upgraded using.pl.

iliajie:
Please have another look at it. It works for me. If you can reproduce a problem, please let me know how exactly I could do it and on which system.

jcameron:
So I re-installed from a tar.gz file on a CentOS 7 system using the most updated code, and I found that `/etc/webmin/start` and `stop` were symlinks to  `/etc/webmin/.start-init` and `.stop-init` respectively. 

A systemd service is created though, with the config : 

```
[Unit]
Description=Webmin server daemon

[Service]
ExecStart=/etc/webmin/.start-init
ExecStop=/etc/webmin/.stop-init
PIDFile=/var/webmin/miniserv.pid
Type=forking
KillMode=none

[Install]
WantedBy=multi-user.target
```

iliajie:
Here is the fix - https://github.com/webmin/webmin/commit/3e01c24434a57a1f35078b0b1d2b6647d82fda47#commitcomment-73047983

iliajie:
I mean that was a discussion the fix is here - https://github.com/webmin/webmin/commit/c6e2df146a8102f2f72d898184c37f72706e836d

PR #1648, opened 2022-05-24
Fix to consider Webmin privileged user

iliajie:
This is a better way of doing this. [Previous PR](https://github.com/webmin/webmin/pull/1647/files) is probably less secure, unless you added `_safe` flag at the same time when adding this feature?


SamSirry (approved):
It seems to solve the issue.

jcameron:
This commit is wrong - mode 2 means only allow users named "root" or "admin".

iliajie:
Here is the right PR, I think - 
https://github.com/webmin/webmin/pull/1649/files

PR #1649, opened 2022-05-25
Add unobtrusive way to test if user is privileged; separate RPC logic

iliajie:
Jamie, I have taken a deeper look and it appears that we should take a different approach in testing admin user.

What are your thoughts on this PR?


web-lib-funcs.pl
```diff
@@ -12118,18 +12118,61 @@ sub delvar
...
+
+# Get current user and base user global permissions
+my %uaccess = &get_module_acl($remote_user, "");
+my %access = &get_module_acl($base_remote_user, "");
+
+# Check if mode must be restricted
+if ($uaccess{'_safe'} == 1 || $access{'_safe'} == 1 ||
+    $uaccess{'rpc'} == 0 || $access{'rpc'} == 0 ||
+    $prod ne "webmin") {
+        $mode = 'user';
```
jcameron, line 12156: I don't think we can just rely in the _safe field, as this was introduced relatively recently - many users have less-privileged Webmin logins that weren't created as "safe users", but should not be considered root-equivalent.

jcameron:
This change is mostly good ... but I think we need to consider where in Webmin it makes to call it? Permissions are generally granted for specific modules or actions, rather than having classes of users that permissions are then based on.

web-lib-funcs.pl
iliajie: It would still work, and be set to `$mode=user` because of `$access{'rpc'} == 0` test?

iliajie: > I don't think we can just rely in the _safe field, as this was introduced relatively recently

I have checked deeper and it appears that all main functionality for the Safe Webmin user was added with [this](https://github.com/webmin/webmin/commit/2075fffe70a8c593d5045f76993793fe5c511356) commit. It already includes `_safe` flag.

Also, as I mentioned before, I expect that `$access{'rpc'} == 0` would also force a regular user mode, even if `_safe` flag is missing.

jcameron: So is there any value in checking for the `_safe` flag then?

web-lib-funcs.pl
```diff
@@ -12175,6 +12180,35 @@ sub webmin_user_is_admin
...
+# Returns 1 if currently logged in user belongs to one
+# of the requested types: root, safe-user, mail-user,
+# cloud-owner, virtual-reseller, virtual-reseller
+# Simply a convenience wrapper function
+sub webmin_user_is
+{
+my ($user_type) = @_;
+
+# Is user root/admin
+return &webmin_user_login_mode() eq 'root'
```
jcameron, line 12202: Better to call webmin_user_login_mode just once, store the result in a variable, and check that below.

web-lib-funcs.pl
iliajie: > So is there any value in checking for the _safe flag then?

Yeah, I test if `$access{'_safe'} == 1` or `$access{'rpc'} == 0` and if true it will fall to safe user mode.

web-lib-funcs.pl
iliajie: > Better to call webmin_user_login_mode just once, store the result in a variable, and check that below.

Yes, great suggestion! Have a look at this patch - https://github.com/webmin/webmin/pull/1649/commits/16e8f9ea9d0be6a6e165ba9b7a20edaeda46b3d1

jcameron: No I meant to just call it once in this function. I don't like the use of `state` to do caching in this case.

iliajie: >  I don't like the use of state to do caching in this case.

Why not? Do you expect that mode can change during the call?

iliajie: > Better to call webmin_user_login_mode just once, store the result in a variable, and check that below.

I see now, done - https://github.com/webmin/webmin/pull/1649/commits/5df581e103238dcd691981b01f4896a520ed9f04

> I don't like the use of state to do caching in this case.

This function can be called dozens of times per call. Unless we expect mode to be changed during the call, I would keep caching. It's very effective.

iliajie:
Removed caching. Anything else here?

web-lib-funcs.pl
jcameron: I'd just prefer to use an explicit global variable as the cache then, rather than using `state`

iliajie: > Thanks, this looks a lot better

Okay, great then.

jcameron:
Looks good!

iliajie:
Thanks!

PR #1651, opened 2022-05-27
Add support for `systemd-journald` systems (without `rsyslog` package)

iliajie:
All contemporary Linux distros do not install `rsyslog` package by default and using `systemd-journald` to collect and store logging data.

What this PR does:

  1. Adds support to display logs on **System ⇾ System Logs** page when there is no `rsyslog` package installed and `journalctl` command is available
  2. Makes the module displayed in _System_ category rather than in _Un-used modules_
  3. Hides all config options related to `rsyslog` functionality
  4. Hides `rsyslog` service start/stop button
  5. Hides _Add a new system log_ button
  6.  Add proper help to manual reference
  7. Changes to display latest lines on the top (even thought logs go only one way in command line, and it's expected to have last lines at the bottom of the screen, in UI logs expected to flow from the top to bottom, which makes it easier to refresh logs and look for new lines instantly at the top of the screen, rather than scrolling down first. For example Cockpit does exactly the same).


iliajie:
Example of how pages will look on `systemd-journald` systems:

**Index page:**
image

**Config page:**
image

jcameron:
Does `systemd-journald` completely replace `syslog` or `rsyslog` ? If so, how are the log files to which messages are written configured?

iliajie:
> Does systemd-journald completely replace syslog or rsyslog 

Yes, although `rsyslog` has an API (`imjournal` and `omjournal`), which enables those two to interact (if needed). It's very possible to have `rsyslog` installed and running alongside with `systemd-journald` and have no problem.

Although, newer Linux systems have no `rsyslog` package installed by default. The point of my patch is to make logs still viewable via UI even in this case scenario.

>  If so, how are the log files to which messages are written configured?

The files like `/var/log/messages`, `/var/log/secure` and etc are not written by default, however they can present on the system for various of reasons. if that's the case we display them, if files exist.

iliajie:
Add severity (priority) support for `journald` systems:

image

jcameron:
On systems without `syslog` or `rsyslog`, do they use `syslog-ng` instead? Webmin has a separate module for that..

swelljoe:
No. systemd-journald is how logging works by default on modern systems.

swelljoe:
I couldn't name a single OS that uses syslog-ng, currently, but maybe some BSD or something uses it? Nope, it's been 10+ years since anybody was using it as a default. https://en.wikipedia.org/wiki/Syslog-ng#Distributions

iliajie:
Yeah, so we should look towards `systemd-journald` but be prepared that roughly half of users (my wild guess) will end-up installing `rsyslog`, as they won't be able to find `/var/log/messages` file, so our System Logs module must be able to work with both at the same time. Besides, as I already mentioned `systemd-journald` and `rsyslog` can operate together pretty well.

swelljoe:
No way it's going to be half. I'd be surprised if even 5% know the difference and even fewer will install rsyslog. They may be confused by the relative lack of text logs in /var/log, but they aren't going to go installing additional (unnecessary) services.

swelljoe:
The journal just just how logging happens on a Linux system today. We should address that, if we can. It's easier to use than syslog (and provides a syslog interface to services, if they need it, but logging to journald is as simple as writing to the console because if it is started with systemd, it will automatically capture that to the journal and you can select it with `journalctl -u <unit-name>`. The journal is good. No reason to fight it.

iliajie:
I hope not. Although, you will get my point if you Google "logs sshd" and see where it leads to.

iliajie:
The journal is really good, no one is fighting it. This is the reason why I made this PR.

>  it will automatically capture that to the journal and you can select it with journalctl -u <unit-name>

Also, this is not really necessary as we already have `journalctl` command and when this opened you can easily filter a service (or anything else) by name.

image

jcameron:
So I'm thinking that if `rsyslog` has been completely replaced, we should have a separate log viewing module for it rather than using the existing module.

swelljoe:
That'd be fine. But, with the syslog module being named "System Logs", if we name the journald module "Journald", people are going to still go to System Logs looking for...system logs.

Nobody really thinks of Systems Logs are being a syslog module...they think of it as a system log viewer. So...if we do make a new module, it probably still needs to show up as System Logs or something along those lines, maybe?

I don't know. The move to journald and systemd has confused a lot of people. Not sure how we can best help them with that confusion.

swelljoe:
Maybe just "Log Viewer" or whatever. There are still lots of plain text logs on a Linux system, even with journald. So we still need that functionality. Those logs just aren't being written by syslog.

iliajie:
So do you want me to take it to a separate module?

jcameron:
Yeah, we should separate log viewing from managing where logs are written to. 

@iliaross if you wanted to work on this, you could leave the current `syslog` module untouched and instead create a new `logviewer` module that just shows the contents of log files. And it would be smart enough to detect if `syslog`, `syslog-ng` or `journald` is in use, and fetch the appropriate list of log files from each.

iliajie:
Yeah, sure, will do!

iliajie:
Added a new module `logviewer` on [this PR](https://github.com/webmin/webmin/pull/1655).

PR #1652, opened 2022-05-28
Add support for `yescrypt` password hashing scheme

iliajie:
Add support for `yescrypt` as requested in [this thread](https://sourceforge.net/p/webadmin/discussion/55377/thread/01e65be3f9/).


useradmin/md5-lib.pl
```diff
@@ -205,14 +205,39 @@ sub check_sha512
...
 $salt ||= '$6$'.substr(time(), -8).'$';
 return crypt($passwd, $salt);
 }
 
+# unix_crypt_supports_yescrypt()
+# Returns 1 if the built-in crypt() function can already do yescrypt
+sub unix_crypt_supports_yescrypt
+{
+my $hash = '$y$j9T$waHytoaqP/CEnKFroGn0S/$fxd5mVc2mBPUc3vv.cpqDckpwrWTyIm2iD4JfnVBi26';
+my $newhash = eval { crypt('test', $hash) };
```
jcameron, line 221: Curious, does this work on many systems?

useradmin/md5-lib.pl
```diff
@@ -205,14 +205,39 @@ sub check_sha512
...
+return &unix_crypt_supports_yescrypt() ? undef : 'Crypt::NaCl::Sodium';
+}
+
+# encrypt_yescrypt(password, [salt])
+# Hashes a password, possibly with the given salt, with yescrypt
+sub encrypt_yescrypt
+{
+my ($passwd, $salt) = @_;
+$salt ||= &substitute_pattern('$y$j9T$[A-Z]{4}.[a-zA-Z0-9]{16}.$[a-zA-Z0-9]{14}.[a-zA-Z0-9]{7}/[a-zA-Z0-9]{15}/[a-zA-Z0-9]{4}');
+return crypt($passwd, $salt);
```
jcameron, line 238: Shouldn't you call `Crypt::NaCl::Sodium` if installed?

useradmin/md5-lib.pl
iliajie: It's meant to work on Ubuntu 22.04 and Debian 11. I didn't expect it to work on Ubuntu 20.04 but it actually did, however at the same time, it failed on Debian 10. Unexpected.

useradmin/md5-lib.pl
iliajie: No, cause system supports it.

iliajie: The most important part here is to pass salt in expected format, otherwize it just won't work. I'm glad I have found another application to my `substitute_pattern` sub. 🙂

useradmin/md5-lib.pl
jcameron: Yeah, the crypt() function doesn't always support all hashing types. That's why elsewhere in this file we fall back to using a Perl module if it's installed.

useradmin/md5-lib.pl
jcameron: Why check if `Crypt::NaCl::Sodium` is installed above though?

iliajie: > Why check if `Crypt::NaCl::Sodium` is installed above though?

It does the same as your `check_sha512` - if system doesn't support it then it will provide a link to install it from CPAN.

iliajie: I have checked once more, and it looks that `usave_edigestsha512` and as a result `usave_edigestyescrypt` won't suggest installing from CPAN disregard, and will only suggest to change encryption method. Is this intended that you didn't include `Crypt::SHA` at the first time to be installable from CPAN?

jcameron: Hmm, I don't know why I did that with SHA512 either! It should be implemented more like `encrypt_md5`

iliajie: > Hmm, I don't know why I did that with SHA512 either! It should be implemented more like encrypt_md5

I have made a [patch](https://github.com/webmin/webmin/pull/1652/commits/dae7f8a397d31902c6a50421cbe4297e993ce4ac) for this to improve error message handling. Although, I assume we cannot (should not) install any Perl modules in case of `crypt` cannot handle it .. as I don't think it will be possible to authenticate even if we will be able to get the right hash.

iliajie:
I think we're done here.

useradmin/md5-lib.pl
jcameron: `crypt` doesn't depend on the perl modules though.

Take a look at how `encrypt_md5` works - if crypt doesn't support MD5, it falls back to calling a perl module.

iliajie: This is because, I assume MD5 is supported by all systems. But, I just think that if we install Perl `Crypt::NaCl::Sodium` module, and generate `yescrypt` hash and then put this to the shadow file, user still won't be able to authenticate! So I would expect, if system `crypt` doesn't happen to work then it won't work system-wide. This is why installing those isn't an option. Also, building from source is required, which isn't a good thing for a production system.

iliajie:
Tell me if there is anything else can be done. It works for me on various OS.

useradmin/md5-lib.pl
jcameron: Whether `yescrypt` is actually supported by PAM and the rest of the system is a separate question though. It seems quite possible that even if the older `crypt` libc method doesn't support this format, it might still be supported when logging in.

iliajie: > Whether yescrypt is actually supported by PAM and the rest of the system is a separate question though.

I was trying to say that this is frutial point.

> It seems quite possible that even if the older crypt libc method doesn't support this format, it might still be supported when logging in.

Certainly not. It can be easily tested by editing `/etc/shadow` file on Debian 10 and re-placing a hash for the _root_ user with _yescrypt_ from Ubuntu 22.04, and it won't work.

~~~
May 29 03:28:05 debug-debian10-gpl.local sshd[16488]: pam_unix(sshd:auth): authentication failure; logname= uid=0 euid=0 tty=ssh ruser= rhost=10.211.55.2  user=root
May 29 03:28:07 debug-debian10-gpl.local sshd[16488]: Failed password for root from 10.211.55.2 port 49967 ssh2
May 29 03:28:07 debug-debian10-gpl.local sshd[16488]: Connection closed by authenticating user root 10.211.55.2 port 49967 [preauth]
~~~

iliajie:
I had to force remove 2 last commits. They were meant for https://github.com/webmin/webmin/pull/1654

useradmin/md5-lib.pl
jcameron: Understood, but do we know for sure that all systems on which `crypt` supports `yescrypt` also support it in `/etc/shadow` ?  That would depend on how the PAM library interprets the shadow file entries.

iliajie: Yes, this should always be the case, as we read it from PAM common password config itself.

jcameron:
Question - which Linux distro are you testing this on?

iliajie:
I tested it on Debian 10, Ubuntu 20.04 and Ubuntu 22.04.

swelljoe:
RHEL and derivatives do not have scrypt or yescrypt. SHA512 is cross-distro compatible. (I don't have a problem with using whatever the system default is, if that's what this is doing, though.)

iliajie:
Yeah, this is what it's doing. Let me try on RHEL and derivatives. I'll let you know a bit later.

iliajie:
Alright, I have made more tests.

For example, RHEL 8 and derivatives indeed do not support `yescrypt`. However, Fedora 36 and Alma 9 do support it.

Although, this is not really important as the code would prevent from writing a hash that is not supported.

I also just added a [patch](https://github.com/webmin/webmin/pull/1652/commits/f431aaafde0964c557f59e7ae1960ebb16e666c6) that will simply hide `yescrypt` from module's config if it's not supported by the system.

FYI, I made a familiar patch to Webmin today and updated `webmin passwd` command to be able to always rely on the system defaults, and end up with secure hashes, rather than DES or MD5.

iliajie:
Alright, I have made more tests.

For example, RHEL 8 and derivatives indeed do not support `yescrypt`. However, Fedora 36 and Alma 9 do support it.

Although, this is not really important as the code would prevent from writing a hash that is not supported.

I also just added a [patch](https://github.com/webmin/webmin/pull/1652/commits/f431aaafde0964c557f59e7ae1960ebb16e666c6) that will simply hide `yescrypt` from module's config if it's not supported by the system.

FYI, I made a familiar [PR](https://github.com/webmin/webmin/pull/1654) for Webmin today and updated `webmin passwd` command to be able to always rely on the system defaults, and end up with secure hashes, rather than DES or MD5.

useradmin/user-lib.pl
```diff
@@ -2724,4 +2724,20 @@ sub generate_random_password
...
+sub config_pre_load
+{
+my ($modconf_info, $modconf_order) = @_;
+my $unix_crypt_supports_yescrypt =
+		unix_crypt_supports_yescrypt();
+
+# Tweak available formats on the config if yescrypt
+# not available and cannot be used on given system
+if (!$unix_crypt_supports_yescrypt) {
+	$modconf_info->{'md5'} =~ s/,5-<tt>yescrypt<\/tt>//;
```
jcameron, line 2739: Wouldn't it instead be better to show a warning on the module's main page if the selected hash format isn't supported? That would be less fragile ..

iliajie: > Wouldn't it instead be better to show a warning on the module's main page if the selected hash format isn't supported?

Yes, this is what was happening already.

> That would be less fragile

This is already quite solid it terms of the match. It cannot fail really (at least I don't see how it could) but I can easily remove it. I just thought that displaying an option that cannot be used is pointless.

So, do you really want me to remove it?

jcameron: Yeah, editing the hash from config.info like this just seems a little sketchy to me. And also it doesn't prevent selection of the wrong option via other means, like editing `/etc/webmin/useradmin/config` directly.

chris001:
Very nice update. [FYI](https://github.com/openwall/yescrypt):
```
As of this writing, yescrypt is the default password hashing scheme on
recent ALT Linux, Debian 11, Fedora 35+, Kali Linux 2021.1+, and Ubuntu
22.04+.  It is also supported in Fedora 29+, RHEL 9+, and Ubuntu 20.04+,
and is recommended for new passwords in Fedora CoreOS.
```

useradmin/user-lib.pl
iliajie: Alright, reverted.

jcameron:
Ok I think we're good ... merging this now.

iliajie:
Thanks!

iliajie:
Jamie, there is a familiar patch to address Webmin side issue, as Webmin passwords must also be properly hashed.

https://github.com/webmin/webmin/pull/1654

Issue #1653, opened 2022-05-28
Webmin default password hashing format


iliajie:
Based on my [latest PR](https://github.com/webmin/webmin/pull/1652) to support _yescrypt_, and when trying to set **Webmin password hashing format** to **Use standard Unix crypt encryption** on **Webmin ⇾ Webmin Configuration: Authentication** page doesn't seem to use _yescrypt_ on Ubuntu 22.04, while it should have.

Also, the default must not be set to MD5 in 2022. If we fix the default to work properly (like Ubuntu 22.04 use _yescrypt_ and Debian 10 use _sha512_ and etc whatever system default is there), then **Use standard Unix crypt encryption** should be a new reasonable default. :--)

jcameron:
I think `yescrypt` is too new to make the default yet, but `sha512` seems reasonable.

iliajie:
This is not the point. We don't care. We rely on the system. System must decide which hashing format to use.

Let's continue on [this PR](https://github.com/webmin/webmin/pull/1654).

PR #1654, opened 2022-05-29
Add Webmin to support system default hashing format

iliajie:
As discussed here https://github.com/webmin/webmin/issues/1653, we **must not use** MD5 by default. The point is to let Webmin try to use default hashing format using `crypt`.

The good thing with this patch, that each system will use for Webmin user passwords a hashing format  that it suggests. For example, Debian 10 will use _SHA512_, while Ubuntu 22.04 will use _yescrypt_ automatically.

Also, we would need to set `md5pass=0` in `/etc/webmin/config` to make it default and work automatically.

### DIFF
--- a/acl/acl-lib.pl
+++ b/acl/acl-lib.pl
@@ -1693,7 +1693,7 @@ else {
 	# Try detecting system default first
 	if (&foreign_available('useradmin')) {
 		&foreign_require('useradmin');
-		return &useradmin::encrypt_password($pass, $salt, 'force_system_detection');
+		return &useradmin::encrypt_password($pass, $salt, 1);
 		}
 	else {
 		# Use Unix DES
--- a/acl/acl-lib.pl
+++ b/acl/acl-lib.pl
@@ -1690,10 +1690,17 @@ elsif ($mode == 2) {
 	return &encrypt_sha512($pass, $salt);
 	}
 else {
-	# Use Unix DES
-	&seed_random();
-	$salt ||= chr(int(rand(26))+65).chr(int(rand(26))+65);
-	return &unix_crypt($pass, $salt);
+	# Try detecting system default first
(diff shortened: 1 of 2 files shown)


iliajie:
Alright, this is done.

acl/acl-lib.pl
```diff
@@ -1693,7 +1693,7 @@ sub encrypt_password
 	# Try detecting system default first
 	if (&foreign_available('useradmin')) {
 		&foreign_require('useradmin');
-		return &useradmin::encrypt_password($pass, $salt);
+		return &useradmin::encrypt_password($pass, $salt, 'force_system_detection');
```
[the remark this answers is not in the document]
iliajie: Because, I have these magic numbers. You can never know what's happening, unless looking at the sub desription or code itself.

I could change it to `1` for you if you find it cleaner.

acl/acl-lib.pl
```diff
@@ -1690,10 +1690,17 @@ sub encrypt_password
 	return &encrypt_sha512($pass, $salt);
 	}
 else {
-	# Use Unix DES
-	&seed_random();
-	$salt ||= chr(int(rand(26))+65).chr(int(rand(26))+65);
-	return &unix_crypt($pass, $salt);
```
jcameron, line 1696: Isn't this change going to break any validation of existing passwords that are encrypted with old-style crypt() ?

iliajie: No, because it's not deleted. Please check slightly below. It will use system default and if module useradmin not available it will fall-back to old DES.

iliajie: Validation will happen in `validate_password` sub in `md5-lib.pl`, which already supports latest formats and been merged few days ago.

jcameron:
Thanks for the patch!

PR #1655, opened 2022-05-29
Add new System Logs Viewer (`logviewer`) module

iliajie:
This module will allow users to view all log files, in a familiar way to what `syslog` module did. The `systemd-journald` service is supported.

From this point it can grow more on its own.

Features:

  1. Keep to display all other modules' log files
  2. Keep ability to add custom logs
  3. Keep ACLs support
  4. Add better `journalctl` support
  5. Add a message, when `syslog` or `syslog-ng` is not installed, to tell a user to try a new module, instead of installing `rsyslog` package

**Examples:**

image
image

image
image


jcameron:
Nice! Will this discover logs belonging to `syslog` though?

logviewer/logviewer-lib.pl
```diff
@@ -0,0 +1,133 @@
...
+return 0;
+}
+
+# 
+# Returns standard 
+sub get_systemctl_cmds
+{
+my $lines = $config{'lines'} || 1000;
+return !&has_command('journalctl') ? () : (
+	{ 'cmd' => "journalctl --lines $lines -p 0..3",
```
jcameron, line 36: What do the numbers 0 to 3 mean?

swelljoe: Priority.

```
-p, --priority=
           Filter output by message priorities or priority ranges. Takes
           either a single numeric or textual log level (i.e. between
           0/"emerg" and 7/"debug"), or a range of numeric/text log
           levels in the form FROM..TO. The log levels are the usual
           syslog log levels as documented in [syslog(3)](https://www.man7.org/linux/man-pages/man3/syslog.3.html), i.e.
           "emerg" (0), "alert" (1), "crit" (2), "err" (3),
           "warning" (4), "notice" (5), "info" (6), "debug" (7). If a
           single log level is specified, all messages with this log
           level or a lower (hence more important) log level are shown.
           If a range is specified, all messages within the range are
           shown, including both the start and the end value of the
           range. This will add "PRIORITY=" matches for the specified
           priorities.
```

I believe the text version would be more self-documenting. e.g. `emerg` instead of `0`.

iliajie:
> Nice! Will this discover logs belonging to `syslog` though?

Yes, it will, logs from Webmin, Apache and etc. Do you mean these ones?

> What do the numbers 0 to 3 mean?

That means (as described on the first page) in "Messages selected" column, to display messages that are error, critical, alert and emergency.

> I believe the text version would be more self-documenting. e.g. emerg instead of 0.

Yes, this is what I did at first (locally) but then I couldn't sort it (easily) properly (based on severity level) and I got back to using numerical ranges. I think you're right. I will change it to text based range.

iliajie:
Alright, the commit above fixed it. I reworked how the messages are displayed and named. Fixed sorting and few other bugs.

**Example:**
image

image

iliajie:
> Nice! Will this discover logs belonging to syslog though?

Oh, I see what you mean now. I didn't think before that the new module would need to list any `syslog` files. I have fixed that on [this commit](https://github.com/webmin/webmin/pull/1655/commits/88c6e06c038831b3a118b7d88773c875c4d87bcc). Now, if `syslog` is installed and available it will also include the logs from it.

image

jcameron:
Excellent! Can you also do the same for logs managed by the `syslog-ng` module?

iliajie:
> Excellent! 

Thanks!

> Can you also do the same for logs managed by the `syslog-ng` module?

Well, yeah, if you think it should be considered too.

jcameron:
We certainly still have users running `syslog-ng` .. and it should be pretty trivial to add code to fetch the list of log files from the Syslog-NG config if the module is installed.

iliajie:
Can do, tomorrow.

iliajie:
Where are we at here?

jcameron:
I think we can call this done, and ready for inclusion in the next release!

Also, maybe in a future PR you could link to this module from the syslog and syslog-ng modules for viewing logs?

iliajie:
It's already linked from `syslog` and `syslog-ng`. I mean, I already added it.

Is this what you mean?
https://github.com/webmin/webmin/pull/1655/commits/3a7dda67cd9425a9b08308ed906d8c57fac6a77d

jcameron:
I meant the other way - if the user clicks on a link to view logs on the `syslog` module, it could just link to this new one

jcameron:
I meant the other way - if the user clicks on a link to view logs on the `syslog` module, it could just link to this new one

iliajie:
Well, this is what my patch does? It mentions a new `logview` module but only if `syslog` or `syslog-ng` not installed.

Like that -

image


Is this what you mean?

jcameron:
No, I meant that if the Syslog is installed, in the `syslog` module we could link to the `logview` module to view logs.

iliajie:
Like FirewallD message in Firewall? Like yellow alert on the top?

jcameron:
It might be easier for me to just make the change myself so you can see what I mean :-)

PR #1659, opened 2022-06-03
Dev/systemd improvements

iliajie:
* Fix change bind page, including port redirects
 * Fix Webmin systemd service unit stability
 * Upgrade page needs more work


useradmin/md5-lib.pl
```diff
@@ -2,6 +2,7 @@
 
 use strict;
 use warnings;
+no warnings 'redefine';
```
jcameron, line 5: This seems like it belongs in a different PR?

webmin/change_bind.cgi
```diff
@@ -175,15 +175,19 @@ else { $url = $ENV{'SERVER_NAME'}; }
 if ($ENV{'HTTPS'} eq "ON") { $url = "https://$url"; }
 else { $url = "http://$url"; }
 
-if ($tconfig{'inframe'}) {
-	# Theme uses frames, so we need to redirect the whole frameset
+# Theme redirect if port changed
+if ($miniserv{'port'} ne $oldminiserv{'port'}) {
```
jcameron, line 179: Should be != instead of ne

iliajie:
Well, because (if I understand this correctly) it is simply impossible to have control over `SIGTERM` the old ways, as _miniserv_ no longer controlling the process, where `systemd` is in charge. As you call`/etc/webmin/stop` (even with `$SIG{'TERM'} = 'ignore';`) `miniserv` will just stop with `systemd` (as designed). Also, you cannot call `/etc/webmin/.stop-init` either, as it will mess with `systemd` perception of the process status.

I made it possible to have port change work smoothly. However, taking down miniserv with `stop` command won't just work in an expected way.

> Why do this in index.cgi ? It seems odd to force stop webmin here ..

If we do it the old way, the process will just die and no redirection will happen. 

I will keep looking more today in doing it in a better way.

iliajie:
Alright, there is much better way of doing it - https://github.com/webmin/webmin/pull/1659/commits/f662edb8147911eb45f468ed8250c8daa097edd9.

Few more small fixes to this PR and done.

iliajie:
Alright, unless you have questions, this can be merged. Works perfectly for me now on all distros!

init/updateboot.pl
```diff
@@ -20,7 +20,14 @@
 			unlink("$p/$product.service");
 			unlink("$p/$product");
 			}
-		copy_source_dest("../webmin-systemd", "$systemd_root/$product.service");
+		my $temp = &transname();
+		&copy_source_dest("../webmin-systemd", "$temp");
```
jcameron, line 24: Use an absolute path rather than ..

web-lib-funcs.pl
```diff
@@ -2401,6 +2401,13 @@ sub reload_miniserv
 {
 my ($ignore) = @_;
 return undef if (&is_readonly_mode());
+if (&has_command('systemctl')) {
+	my $restart_target =
```
jcameron, line 2405: Couldn't this just be `my $restart_target = &get_product_name();` ?

init/updateboot.pl
jcameron: I still don't see why an absolute path like `$root_directory/webmin-systemd` can't be used here?

iliajie:
> I still don't see why an absolute path like $root_directory/webmin-systemd can't be used here?

Yes, alright. Added it.

iliajie:
> Couldn't this just be my $restart_target = &get_product_name(); ?

Yes, I could by I wanted to be super specific as this is going to system command.

init/updateboot.pl
```diff
@@ -20,7 +20,14 @@
 			unlink("$p/$product.service");
 			unlink("$p/$product");
 			}
-		copy_source_dest("../webmin-systemd", "$systemd_root/$product.service");
+		my $temp = &transname();
+		&copy_source_dest("$root_directory/webmin-systemd", "$temp");
```
jcameron, line 24: $temp doesn't need to be in quotes here

PR #1662, opened 2022-06-07
Simply replace Gray Framed with Virtualmin Framed theme

iliajie:
For the start, let's just replace Gray Theme with Virtualmin Framed Theme. It will just work for users with the same setup.

`virtualmin-server-theme` repo can be archived.


jcameron:
This is a big big change .. let's not do it now when we have multiple other large changes in flight.

iliajie:
Alright, fair point.

We can do it with 1.996. Although, the change is not changing anything for the user. It's the same theme from the user's config perspective.

jcameron:
Yeah, let's look into this after we do a release

iliajie:
I assume not yet or 1.997 is good for it?

jcameron:
Let's see if we need a new bugfix release first on top of 1.996

iliajie:
Yeah, I think we do need 1.997 have things done right and have it released to Virtualmin users finally.

I think this can go to 1.998 if you're afraid it may break something.

jcameron:
Yeah let's put this in 1.998.

iliajie:
Great, I'm fine with this!

iliajie:
Are we putting this to 1.998?

jcameron:
If we need 1.998 to fix bugs in 1.997, let's hold off on this. I want to get a release out that's bug-free

iliajie:
Alright, makes perfect sense. Let's put it off until 1.999.

Also, where do you potentially see bugs in here, so I could try to prevent any issues before we later add it to 1.999?

jcameron:
I just assume that all large changes will have bugs :-)

iliajie:
Okay! :-)

iliajie:
This can be merged.

jcameron:
Did you also apply this fix : https://github.com/virtualmin/virtual-server-theme/commit/958a961b689117b35195636f1403dd1823811c89

iliajie:
I ran tests with builds and it worked just fine for all users that previously had either `blue-theme` or `gray-theme` set.

PR #1664, opened 2022-06-09
Fix to use reload over restart when possible

iliajie:
Using restart required for changing binds and re-reading `userfile` and `keyfile`.

Re-reading ACLs, user files and etc works with _reload_.


jcameron:
Thanks for this patch!

PR #1667, opened 2022-06-15
Add new `webmin server` command to control `miniserv`

iliajie:
This PR made to draw attention and doesn't need to be reviewed.

New Webmin 1.995 will have a new `webmin` CLI sub-command to control Webmin web-server, for example:

```
webmin restart
```

or 

```
webmin status
```

The full command description is available by running `webmin server -h`, i.e.:

```Bash
Usage:
     webmin server [command]
     webmin [command]

Options:
    --help, -h
         Print this usage summary and exit.

         Examples of usage:
           - webmin server status
           - webmin server restart
           - webmin server --config /usr/local/etc/webmin --command start
           - webmin status
           - webmin restart

    --config, -c
         Specify the full path to the Webmin configuration directory. Defaults to C</etc/webmin>

    --command, -x
         Available commands:
           - status
           - start
           - stop
           - restart
           - reload
           - force-restart
           - force-reload
           - kill

         Alias commands:
           - force-restart | force-reload
           - kill | force-kill
```

### DIFF
--- a/bin/disable-proxy
+++ b/bin/disable-proxy
@@ -129,7 +129,7 @@ Disable proxy-related features in Webmin.
 
 =head1 SYNOPSIS
 
-disable-proxy [options]
+webmin disable-proxy [options]
 
 =head1 OPTIONS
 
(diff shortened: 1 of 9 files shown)


swelljoe:
In a world with systemctl, this seems counterproductive. Like, why make it confusing with multiple ways to manage the server?

swelljoe:
Really, the more I think of this, the less I like it. It is almost a negative to have more than one obvious way to do things when it comes to user experience. This is almost, but not quite, identical to interacting with systemd, so it's adding no features/capabilities or ease of use, but it is adding complexity, both for the user and for maintaining the code. All cost, no benefit.

PR #1669, opened 2022-06-15
Add support for `yescrypt` and `sha512` upon installation; fix other bug

iliajie:
This PR adds support for having default user initial password comply with system strongest hashing format available.

It also resets Webmin default hashing format to auto detect (system default), so when a new Webmin user is created or a password changed for existing Webmin user in `acl` module, it would be hashed properly.

### DIFF
--- a/setup.pl
+++ b/setup.pl
@@ -299,9 +299,14 @@ else {
 		$crypt = $ENV{'crypt'};
 		}
 	else {
+        system("stty -echo");
 		chop($password = <STDIN>);
-		print "Password again: ";
+        system("stty echo");
+		print "\nPassword again: ";
+        system("stty -echo");
 		chop($password2 = <STDIN>);
+        system("stty echo");
+        print "\n";
 		if ($password ne $password2) {
 			&errorexit("Passwords don't match");
 			}
--- a/setup.pl
+++ b/setup.pl
@@ -435,8 +435,9 @@ else {
 		}
 
 	# Generate random
-	$salt8 = substr(time(), -8);
-	$salt2 = substr(time(), -2);
+	@saltbase = ('a'..'z', 'A'..'Z', '0'..'9', split(//, time()));
+	$salt8 = join('', map ($saltbase[rand(@saltbase)], 1..8));
+	$salt2 = join('', map ($saltbase[rand(@saltbase)], 1..2));
 
 	# Create users file
 	open(UFILE, ">$ufile");
--- a/setup.pl
+++ b/setup.pl
@@ -718,10 +718,8 @@ if (!defined($miniserv{'passwd_mode'})) {
 	$miniserv{'passwd_mode'} = 0;
 	}
 
-# If Perl crypt supports MD5, then make it the default
-if ($md5pass) {
-	$gconfig{'md5pass'} = 1;
-	}
+# Use system default for password hashing
+$gconfig{'md5pass'} = 0;
 
 # Set a special theme if none was set before
 if ($ENV{'theme'}) {
--- a/setup.pl
+++ b/setup.pl
@@ -636,8 +636,7 @@ else {
 	symlink("$config_directory/.reload-init", "$config_directory/reload");
 
 	# For systemd
-	my $systemctlcmd = `which systemctl`;
-	$systemctlcmd =~ s/\s+$//;
+	my $systemctlcmd = &has_command('systemctl');
 	if (-x $systemctlcmd) {
 
 		# Clear existing
--- a/setup.pl
+++ b/setup.pl
@@ -299,14 +299,14 @@ else {
 		$crypt = $ENV{'crypt'};
 		}
 	else {
-        system("stty -echo");
+		system("stty -echo");
 		chop($password = <STDIN>);
-        system("stty echo");
+		system("stty echo");
 		print "\nPassword again: ";
-        system("stty -echo");
+		system("stty -echo");
 		chop($password2 = <STDIN>);
-        system("stty echo");
-        print "\n";
+		system("stty echo");
+		print "\n";
 		if ($password ne $password2) {
 			&errorexit("Passwords don't match");
 			}
--- a/setup.pl
+++ b/setup.pl
@@ -435,7 +435,7 @@ else {
 		}
 
 	# Generate random
-	@saltbase = ('a'..'z', 'A'..'Z', '0'..'9', split(//, time()));
+	@saltbase = ('a'..'z', 'A'..'Z', '0'..'9');
 	$salt8 = join('', map ($saltbase[rand(@saltbase)], 1..8));
 	$salt2 = join('', map ($saltbase[rand(@saltbase)], 1..2));
 
--- a/setup.pl
+++ b/setup.pl
@@ -418,21 +418,37 @@ else {
 		}
 	&put_miniserv_config(\%miniserv);
 
-	# Test MD5 password encryption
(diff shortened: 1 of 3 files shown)


iliajie:
Jamie, all these commits on the PR were tested and ready to be reviewed and merged.

setup.pl
```diff
@@ -299,9 +299,14 @@
 		$crypt = $ENV{'crypt'};
 		}
 	else {
+        system("stty -echo");
```
jcameron, line 302: Can you fix the indents here?

setup.pl
```diff
@@ -418,21 +423,37 @@
...
+		}
+	if (&unix_crypt('test', '$6$Tk5o/GEE$zjvXhYf/dr5M7/jan3pgunkNrAsKmQO9r5O8sr/Cr1hFOLkWmsH4iE9hhqdmHwXd5Pzm4ubBWTEjtMeC.h5qv1') eq '$6$Tk5o/GEE$zjvXhYf/dr5M7/jan3pgunkNrAsKmQO9r5O8sr/Cr1hFOLkWmsH4iE9hhqdmHwXd5Pzm4ubBWTEjtMeC.h5qv1') {
+		$sha512pass = 1;
+		}
+	if (&unix_crypt('test', '$1$A9wB3O18$zaZgqrEmb9VNltWTL454R/') eq '$1$A9wB3O18$zaZgqrEmb9VNltWTL454R/') {
 		$md5pass = 1;
 		}
 
+	# Generate random
+	$salt8 = substr(time(), -8);
```
jcameron, line 438: Pity we can't generate a random salt to be safer? The last 2 digits of time() has only 100 possibilities, compared to 26^2 that are allowed. You could instead do something like :

`chr(int(rand(26))+65) . chr(int(rand(26))+65);`

setup.sh
```diff
@@ -535,16 +535,26 @@ else
 		cat "$wadir/miniserv-conf" >>$cfile
 	fi
 
+	# Test availble hashing formats
+	yescryptpass=`$perl -e 'print crypt("test", "\\$y\\$j9T\\$waHytoaqP/CEnKFroGn0S/\\$fxd5mVc2mBPUc3vv.cpqDckpwrWTyIm2iD4JfnVBi26") eq "\\$y\\$j9T\\$waHytoaqP/CEnKFroGn0S/\\$fxd5mVc2mBPUc3vv.cpqDckpwrWTyIm2iD4JfnVBi26" ? "1\n" : "0\n"'`
+	sha512pass=`$perl -e 'print crypt("test", "\\$6\\$Tk5o/GEE\\$zjvXhYf/dr5M7/jan3pgunkNrAsKmQO9r5O8sr/Cr1hFOLkWmsH4iE9hhqdmHwXd5Pzm4ubBWTEjtMeC.h5qv1") eq "\\$6\\$Tk5o/GEE\\$zjvXhYf/dr5M7/jan3pgunkNrAsKmQO9r5O8sr/Cr1hFOLkWmsH4iE9hhqdmHwXd5Pzm4ubBWTEjtMeC.h5qv1" ? "1\n" : "0\n"'`
 	md5pass=`$perl -e 'print crypt("test", "\\$1\\$A9wB3O18\\$zaZgqrEmb9VNltWTL454R/") eq "\\$1\\$A9wB3O18\\$zaZgqrEmb9VNltWTL454R/" ? "1\n" : "0\n"'`
 
+	salt8=`tr -dc A-Za-z0-9 </dev/urandom | head -c 8 ; echo ''`
```
jcameron, line 543: Do all operating systems have a `/dev/random` file? I'm curious if this will work on FreeBSD or Solaris.

setup.sh
```diff
@@ -535,16 +535,26 @@ else
...
+	salt2=`tr -dc A-Za-z0-9 </dev/urandom | head -c 2 ; echo ''`
+
 	ufile=$config_dir/miniserv.users
 	if [ "$crypt" != "" ]; then
 		echo "$login:$crypt:0" > $ufile
 	else
-		if [ "$md5pass" = "1" ]; then
-			$perl -e 'print "$ARGV[0]:",crypt($ARGV[1], "\$1\$XXXXXXXX"),":0\n"' "$login" "$password" > $ufile
+		if [ "$yescryptpass" = "1" ]; then
+			$perl -e 'print "$ARGV[0]:",crypt($ARGV[1], "\$y\$j9T\$'$salt8'"),":0\n"' "$login" "$password" > $ufile
```
jcameron, line 551: Pass in $salt8 as a parameter like $login , rather than inserting it into the command.

web-lib-funcs.pl
```diff
@@ -11001,16 +11001,15 @@ =head2 unix_crypt(password, salt)
 sub unix_crypt
 {
 my ($pass, $salt) = @_;
-return "" if ($salt !~ /^[a-zA-Z0-9\.\/]{2}/);   # same as real crypt
-my $rv = eval "crypt(\$pass, \$salt)";
-my $err = $@;
-return $rv if ($rv && !$@);
+return "" if ($salt !~ /^[\$a-zA-Z0-9]{2}/);   # same as real crypt
+my $rv = eval { crypt($pass, $salt) };
```
jcameron, line 11005: I think the use of eval "" should stay, as there are some rare systems on which `crypt` doesn't work.

setup.sh
iliajie: `</dev/urandom`, yes, it works on FreeBSD. No idea about Solaris.

setup.pl
iliajie: Let's do better something like this - https://github.com/webmin/webmin/commit/cd1ae41ad557af9bb6fa48540a08a124a039c76b

setup.sh
iliajie: I don't think it makes any difference in this case scenario. If you think it does, could you please show how you expect this to be done?

web-lib-funcs.pl
iliajie: Yes, but I didn't remove `eval`. Also, I really doubt those systems are still around.

setup.pl
iliajie: Indent fixed - https://github.com/webmin/webmin/commit/defa41f0b3098f742062b126bd1c7099a11f20af.

setup.sh
jcameron: Looks like it does exist, so this should be safe

setup.sh
jcameron: The line would be like : 

```
 $perl -e 'print "$ARGV[0]:",crypt($ARGV[1], "\$y\$j9T\$$ARGV[2]"),":0\n"' "$login" "$password" "$salt8" > $ufile
```

setup.sh
iliajie: Great.

web-lib-funcs.pl
jcameron: Ok, I think this will be fine then

iliajie: Yeah, it is. Tested it.

setup.sh
iliajie: Alright, let's do it that way https://github.com/webmin/webmin/commit/ebfb5dbae1606536ac37bd5779180bb263ff78bc https://github.com/webmin/webmin/commit/a21626bd219e4bb68883957b6298dde8ece6cb52.

jcameron:
Thanks for the patch!

iliajie:
Welcome!

Issue #1675, opened 2022-07-01
Upgrade to 1.995 on Ubuntu & Debian fails


TheBudda:
A little heads-up! I think you got a bug! Just done an upgrade to 1.995, via webmin, the upgrade fails lost connection and webmin stopped working

I had to run sudo dpkg --configure -a to fix it! Please see screenshot

![webmin_update01 07 2022](https://user-images.githubusercontent.com/16983376/176877080-880fdf31-ff6d-4845-b0a8-53ded88a007b.png)

iliajie:
Damn, I'm afraid you're right! 😞 

`makerpm.pl` and `makedebian.pl` call `/etc/webmin/stop` which in turn call `systemctl stop webmin`. I cannot believe I missed that! 

Thanks for a report.

TheBudda:
> Damn, I'm afraid you're right! disappointed
> 
> `makerpm.pl` and `makedebian.pl` call `/etc/webmin/stop` which in turn call `systemctl stop webmin`. I cannot believe I missed that!
> 
> Thanks for a report.

You are very welcome. Happy to always help

jcameron:
@iliaross why does `systemctl stop webmin` kill Webmin when upgrading via the UI when we have have this code to prevent the process from being killed by a signal? https://github.com/webmin/webmin/blob/master/package-updates/update.cgi#L45

iliajie:
Because default `systemd` kill mode (alternatively set with `KillMode=`) is `control-group` and running `systemctl stop` will completely stop all remaining processes in control group (including the main process). The work around for us could be is to set `KillMode` to `none` but `systemd` documentation advises _strongly against_ doing that.

Now, as you fixed `miniserv` bug before, restarts will work correctly in all cases and sending _SIGHUP_ will work rock solid with all `systemd` versions.

---

Moreover, after thinking more about this problem and doing additional tests, I think I will change my **PR** to _always_ send _SIGHUP_ on update, meaning that we must be able to restart even older versions of Webmin and make transition smooth.

My initial extra precaution potentially would  cause more issues, as even though sending _SIGHUP_ on the older versions of `miniserv` changes _PID_ it really shouldn't crash Webmin in most case scenarios.

jcameron:
@TheBudda which Webmin version were you upgrading from? 

Also, which Debian or Ubuntu version are you running there?

TheBudda:
> @TheBudda which Webmin version were you upgrading from?
> 
[...]

PR #1676, opened 2022-07-01
Fix Webmin updates from UI

iliajie:
This has been sorted!

 1. Webmin will be restarted from UI correctly (starting 1.996+)
 2. On `systemd` systems it will respect current status after upgrades (i.e. Webmin won't be restarted if it was stopped)
 3. Signals are only sent from scriptlets 
 4. Bonus: Fix `miniserv.pem` generation in `setup.pl`

It won't be reliably possible to send `SIGHUP` to the process running Webmin 1.994 as it will simply may die (and most likely will). Upgrades made from Webmin 1.994 to 1.996 will have to be restarted manually. There is nothing we can do, unfortunately (due to that `miniserv` bug that was recently fixed). After version 1.996 all upgrades (including from UI) will flow smoothly. Upgrading from 1.994 to 1.996 will show a nice message in package manager explaining what to do (how to restart it). This is the best safest we can do.

I am 1000% sure that it must be resolved now! Please have a look.

### DIFF
--- a/makedebian.pl
+++ b/makedebian.pl
@@ -260,11 +260,19 @@ system("chmod 755 $preinstall_file");
 open(SCRIPT, ">$postinstall_file");
 print SCRIPT <<EOF;
 #!/bin/sh
+justinstalled=1
+if [ -e "/etc/$baseproduct" ]; then
+	justinstalled=0
+fi
+supportupdprepost=0
 inetd=`grep "^inetd=" /etc/$baseproduct/miniserv.conf 2>/dev/null | sed -e 's/inetd=//g'`
-if [ "\$1" = "upgrade" -a "\$1" != "abort-upgrade" ]; then
+if [ "\$1" = "configure" ]; then
 	# Upgrading the package, so stop the old Webmin properly
 	if [ "\$inetd" != "1" ]; then
-		/etc/$baseproduct/stop >/dev/null 2>&1 </dev/null
+		if [ -e "/etc/$baseproduct/.pre-install" ]; then
+			supportupdprepost=1
+			/etc/$baseproduct/.pre-install >/dev/null 2>&1 </dev/null
+		fi
 	fi
 fi
 cd /usr/share/$baseproduct
@@ -292,7 +300,8 @@ autothird=1
 noperlpath=1
 nouninstall=1
 nostart=1
-export config_dir var_dir perl autoos port login crypt host ssl nochown autothird noperlpath nouninstall nostart allow atboot makeboot
+nostop=1
+export config_dir var_dir perl autoos port login crypt host ssl nochown autothird noperlpath nouninstall nostart nostop allow atboot makeboot
 tempdir=/tmp/.webmin
 if [ ! -d \$tempdir ]; then
 	tempdir=/tmp
@@ -308,10 +317,18 @@ fi
 rm -f /var/lock/subsys/$baseproduct
 
 if [ "$inetd" != "1" ]; then
-	/etc/$baseproduct/stop >/dev/null 2>&1 </dev/null
-	/etc/$baseproduct/start >/dev/null 2>&1 </dev/null
-	if [ "\$?" != "0" ]; then
-		echo "W: $ucproduct server cannot be restarted. It is advised to restart it manually by\nrunning \\"/etc/webmin/restart-by-force-kill\\" when upgrade process is finished"
+	if [ "\$justinstalled" = "1" ]; then
+		/etc/$baseproduct/start >/dev/null 2>&1 </dev/null
+		if [ "\$?" != "0" ]; then
+			echo "E: Webmin server cannot be started. It is advised to start it manually\n   by running \\"webmin force-restart\\" command"
+		fi
+	elif [ "\$supportupdprepost" = "1" ]; then
+			/etc/$baseproduct/.post-install >/dev/null 2>&1 </dev/null
+		if [ "\$?" != "0" ]; then
+			echo "W: Webmin server cannot be restarted. It is advised to restart it manually\n   by running \\"webmin force-restart\\" command when upgrade process is finished"
+		fi
+	else
+		echo "W: Webmin server will not be restarted. It is advised to restart it manually\n   by running \\"webmin force-restart\\" command when upgrade process is finished"
 	fi
 fi
 
@@ -360,10 +377,10 @@ if [ "\$1" != "upgrade" -a "\$1" != "abort-upgrade" ]; then
 	if [ "\$?" = 0 ]; then
 		# Package is being removed, and no new version of webmin
 		# has taken it's place. Run uninstalls and stop the server
+		/etc/$baseproduct/stop >/dev/null 2>&1 </dev/null
 		if [ "$product" = "webmin" ]; then
 			(cd /usr/share/$baseproduct ; WEBMIN_CONFIG=/etc/$baseproduct WEBMIN_VAR=/var/$baseproduct LANG= /usr/share/$baseproduct/run-uninstalls.pl) >/dev/null 2>&1 </dev/null
 		fi
-		/etc/$baseproduct/stop >/dev/null 2>&1 </dev/null
 		/bin/true
 	fi
 fi
--- a/makedebian.pl
+++ b/makedebian.pl
@@ -299,7 +299,7 @@ noperlpath=1
 nouninstall=1
 nostart=1
 nostop=1
-export config_dir var_dir perl autoos port login crypt host ssl nochown autothird noperlpath nouninstall nostart nostop allow atboot makeboot
+export config_dir var_dir perl autoos port login crypt host ssl nochown autothird noperlpath nouninstall nostart allow atboot makeboot nostop
 tempdir=/tmp/.webmin
 if [ ! -d \$tempdir ]; then
 	tempdir=/tmp
--- a/makedebian.pl
+++ b/makedebian.pl
@@ -264,13 +264,11 @@ justinstalled=1
 if [ -e "/etc/$baseproduct" ]; then
 	justinstalled=0
 fi
-supportupdprepost=0
 inetd=`grep "^inetd=" /etc/$baseproduct/miniserv.conf 2>/dev/null | sed -e 's/inetd=//g'`
 if [ "\$1" = "configure" ]; then
 	# Upgrading the package, so stop the old Webmin properly
 	if [ "\$inetd" != "1" ]; then
 		if [ -e "/etc/$baseproduct/.pre-install" ]; then
-			supportupdprepost=1
 			/etc/$baseproduct/.pre-install >/dev/null 2>&1 </dev/null
 		fi
 	fi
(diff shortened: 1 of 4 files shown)


jcameron:
This is a BIG change ... is everything in here necessary to fix the restart issue?

Before we make any changes, can you explain here why the restart isn't working from the UI?

iliajie:
> Before we make any changes, can you explain here why the restart isn't working from the UI?

Yes, explained here https://github.com/webmin/webmin/issues/1675#issuecomment-1172859762 on your familiar question.

> This is a BIG change ... is everything in here necessary to fix the restart issue?

Yes, everything (aside from a bonus fix for `miniserv.pem` generation). I also just pushed a new commit to always try to restart Webmin, which should work even for older versions of Webmin. I was extra precautious about not always performing post-install restart.

makedebian.pl
```diff
@@ -260,11 +260,17 @@
 open(SCRIPT, ">$postinstall_file");
 print SCRIPT <<EOF;
 #!/bin/sh
+justinstalled=1
```
iliajie, line 263: N: In case of a new install we always need to start Webmin after setup is done.

makedebian.pl
```diff
@@ -260,11 +260,17 @@
...
+if [ -e "/etc/$baseproduct" ]; then
+	justinstalled=0
+fi
 inetd=`grep "^inetd=" /etc/$baseproduct/miniserv.conf 2>/dev/null | sed -e 's/inetd=//g'`
-if [ "\$1" = "upgrade" -a "\$1" != "abort-upgrade" ]; then
+if [ "\$1" = "configure" ]; then
```
iliajie, line 268: N: `configure` is performed on upgrades and installs (including `--reinstall`) (this is what we want).

makedebian.pl
```diff
@@ -292,7 +298,8 @@
 noperlpath=1
 nouninstall=1
 nostart=1
-export config_dir var_dir perl autoos port login crypt host ssl nochown autothird noperlpath nouninstall nostart allow atboot makeboot
+nostop=1
```
iliajie, line 301: N: We cannot allow `setup.sh` to perform any actions on the process when run from scriptlets. It must happen once and in predictable manner.

makedebian.pl
```diff
@@ -308,10 +315,16 @@
 rm -f /var/lock/subsys/$baseproduct
 
 if [ "$inetd" != "1" ]; then
-	/etc/$baseproduct/stop >/dev/null 2>&1 </dev/null
-	/etc/$baseproduct/start >/dev/null 2>&1 </dev/null
-	if [ "\$?" != "0" ]; then
-		echo "W: $ucproduct server cannot be restarted. It is advised to restart it manually by\nrunning \\"/etc/webmin/restart-by-force-kill\\" when upgrade process is finished"
+	if [ "\$justinstalled" = "1" ]; then
+		/etc/$baseproduct/start >/dev/null 2>&1 </dev/null
```
iliajie, line 319: N: If newly installed, always start Webmin/Usermin.

makedebian.pl
```diff
@@ -308,10 +315,16 @@
...
+		if [ "\$?" != "0" ]; then
+			echo "E: Webmin server cannot be started. It is advised to start it manually\n   by running \\"webmin force-restart\\" command"
```
iliajie, line 321: N: If for some unexpected reason Webmin/Usermin cannot be started (which should never happen) it will display an error message.

makedebian.pl
```diff
@@ -260,11 +260,17 @@
...
+	justinstalled=0
+fi
 inetd=`grep "^inetd=" /etc/$baseproduct/miniserv.conf 2>/dev/null | sed -e 's/inetd=//g'`
-if [ "\$1" = "upgrade" -a "\$1" != "abort-upgrade" ]; then
+if [ "\$1" = "configure" ]; then
 	# Upgrading the package, so stop the old Webmin properly
 	if [ "\$inetd" != "1" ]; then
-		/etc/$baseproduct/stop >/dev/null 2>&1 </dev/null
+		if [ -e "/etc/$baseproduct/.pre-install" ]; then
+			/etc/$baseproduct/.pre-install >/dev/null 2>&1 </dev/null
```
iliajie, line 272: N: In case of update or reinstall on all Webmin versions 1.996+ run `.pre-install` command, which will simply either run your older code from `/etc/webmin/.stop-init` script or in case of a `systemd` systems it will send _SIGSTOP_ on the main process (only!) while upgrading.

makedebian.pl
```diff
@@ -308,10 +315,16 @@
...
-	/etc/$baseproduct/start >/dev/null 2>&1 </dev/null
-	if [ "\$?" != "0" ]; then
-		echo "W: $ucproduct server cannot be restarted. It is advised to restart it manually by\nrunning \\"/etc/webmin/restart-by-force-kill\\" when upgrade process is finished"
+	if [ "\$justinstalled" = "1" ]; then
+		/etc/$baseproduct/start >/dev/null 2>&1 </dev/null
+		if [ "\$?" != "0" ]; then
+			echo "E: Webmin server cannot be started. It is advised to start it manually\n   by running \\"webmin force-restart\\" command"
+		fi
+	else
+		/etc/$baseproduct/.post-install >/dev/null 2>&1 </dev/null
```
iliajie, line 324: N: In case of upgrade or reinstall always run conditional `.post-install` script, which will simply either run your older code from `/etc/webmin/.start-init` script or in case of a `systemd` systems it will send two signals _SIGCONT_ followed by _SIGHUP_ (making Webmin completely reload `miniserv` configuration).

makerpm.pl
```diff
@@ -170,19 +170,13 @@
 
 %post
 inetd=`grep "^inetd=" /etc/webmin/miniserv.conf 2>/dev/null | sed -e 's/inetd=//g'`
-startafter=0
-
 if [ "\$1" != 1 ]; then
```
iliajie, line 173: Logic for RHEL is identical as for Debian - no comments here.

setup.pl
```diff
@@ -143,7 +143,9 @@
 			system("$config_directory/stop.bat >/dev/null 2>&1");
 			}
 		else {
-			system("$config_directory/stop >/dev/null 2>&1");
+			if (-r "$config_directory/.pre-install") {
```
iliajie, line 146: Also, call conditional `.pre-install` to handle old and new systems (with `systemd`) correctly.

setup.pl
```diff
@@ -465,7 +467,7 @@
 		$host = &get_system_hostname();
 		$cert = &tempname();
 		$key = &tempname();
-		open(SSL, "| openssl req -newkey rsa:512 -x509 -nodes -out $cert -keyout $key -days 1825 >/dev/null 2>&1");
+		open(SSL, "| openssl req -newkey rsa:2048 -x509 -nodes -out $cert -keyout $key -days 1825 -sha256 >/dev/null 2>&1");
```
iliajie, line 470: N: Users reported it in the past, that they couldn't start Webmin with `miniserv.pem` generated with 512 key length, as 512 key length is forbidden nowadays and it will always fail with an error upon starting Webmin, saying something like `cannot create security context`.

I think in 2022 we have to switch to 4096 to be default key length. What do you think?

setup.pl
```diff
@@ -560,7 +562,14 @@
 		print START "exec '$wadir/miniserv.pl' $config_directory/miniserv.conf\n";
 		}
 	close(START);
-	$start_cmd = "$config_directory/start";
+
+	# Define final start command
+	if ($upgrading) {
```
iliajie, line 567: N: Start script must be conditional. In case of install - it must always call `start` command and in case of upgrade it must always call conditional `.post-install`.

setup.pl
```diff
@@ -609,11 +618,25 @@
 	print RELOAD "kill -USR1 \`cat \$pidfile\`\n";
 	close(RELOAD);
 
+	# Pre install
+	open(PREINST, ">$config_directory/.pre-install");
```
iliajie, line 622: N: It just generates default scripts for pre and post install commands, and in this section does it for older systems to use your code from `.stop-init` and `.start-init` scripts.

setup.pl
```diff
@@ -674,11 +697,26 @@
 		print RELOADD "$systemctlcmd reload webmin\n";
 		close(RELOADD);
 
+		# Pre install
+		open(PREINSTT, ">$config_directory/.pre-install");
```
iliajie, line 701: N: This section generates pre and post install scripts for newer systems with `systemd` to use _SIGSTOP_ and _SIGCONT+SIGHUP_ respectively and only on the _main_ process. We can not leave it default to _all_ as it would freeze the whole control group, making updates from UI itself frozen too.

setup.sh
```diff
@@ -614,11 +614,10 @@ if [ "$noperlpath" = "" ]; then
 	echo ""
 fi
 
-# Re-generating main
-rm -f $config_dir/.stop-init $config_dir/.start-init $config_dir/.restart-init $config_dir/.restart-by-force-kill-init $config_dir/.reload-init
+# Re-generating main scripts
 echo "Creating start and stop init scripts.."
 # Start main
```
iliajie, line 619: N: Nothing to explain for `setup.sh` as the logic is nearly identical to `setup.pl`.

setup.pl
jcameron: I agree, but that change shouldn't be in this PR as it's unrelated.

iliajie: okay, i'll remove it and commit that directly.

makerpm.pl
```diff
@@ -210,19 +204,26 @@
...
-if [ "\$inetd" != "1" -a "\$startafter" = "1" ]; then
-	/etc/webmin/stop >/dev/null 2>&1 </dev/null
-	/etc/webmin/start >/dev/null 2>&1 </dev/null
-	if [ "\$?" != "0" ]; then
-		echo "warning: Webmin server cannot be restarted. It is advised to restart it manually by\nrunning \\"/etc/webmin/restart-by-force-kill\\" when upgrade process is finished"
+if [ "\$inetd" != "1" ]; then
+	if [ "\$1" == 1 ]; then
+		/etc/webmin/start >/dev/null 2>&1 </dev/null
+		if [ "\$?" != "0" ]; then
+			echo "error: Webmin server cannot be started. It is advised to start it manually\n       by running \\"webmin force-restart\\" command"
```
jcameron, line 220: Didn't we agree elsewhere not to recommend using the "webmin" command to control the process? Can this be replace by /etc/webmin/force-restart instead?

iliajie:
Jamie, we could get away with using `KillMode=process` (as a smallest possible change) but that is not recommend by systemd developers!

iliajie:
Although, this is a worse solution, as `systemd` will always log and complain about left-over processes:

```
Jul 02 21:52:03 ubuntu22-pro.local systemd[1]: webmin.service: Found left-over process 241809 (webmin.postinst) in control group while starting unit. Ignoring.
Jul 02 21:52:03 ubuntu22-pro.local systemd[1]: This usually indicates unclean termination of a previous run, or service implementation deficiencies.
Jul 02 21:52:03 ubuntu22-pro.local systemd[1]: webmin.service: Found left-over process 242904 (start) in control group while starting unit. Ignoring.

```

 Also, it won't solve the problem with another user restarting Webmin in parallel (like creating a domain in Virtualmin during Webmin upgrade).

I am strongly confident we need a proper solution as in this PR. This is dangerous to go half-way around, even though it may seem safer (as lesser code change).

iliajie:
Stale. Suppressed with https://github.com/webmin/webmin/pull/1677.

PR #1677, opened 2022-07-03
Reprise: Fix Webmin restarts on upgrades

iliajie:
As we added numerous small changes recently which intersect with yesterday's PR, I am closing that one, and adding all vital changes to this new PR, to have a clearer picture about what's happening.

**Summary:**

  1. Upgrading Webmin from 1.994 to 1.995 will run `systemctl stop webmin` command upon upgrade:
     * This *must not* kill Webmin, i.e. upgrade from UI will seem finished because Webmin 1.994 has _KillMode_ set to _none_;
     * However, upgrading further, i.e. 1.995 to 1.996 _will fail_ but only _if_ Webmin was restarted correctly when previously upgrading from 1.994 to 1.995! This is due to the fact that proper restart must change unit previous _KillMode_ that is was set to _none_ to _control-group_. If Webmin, when upgrading from 1.994 to 1.995, wasn't restarted correctly, you will see _false positive_ picture, where further upgrades from 1.995 to 1.996 happens "correctly" - don't be cheated by this.

  2. Proper restart of Webmin 1.994 will change the _KillMode_ to _control-group_, where further later _any_ Webmin upgrades will _kill_ Webmin upgrades from UI, because calling  `systemctl stop webmin` won't be forgiving with default _KillMode_ set to _control-group_;

  3. This PR address this issue delicately:
     * First of all, instead of completely killing the process, we will send _SIGSTOP_ signal _only_ on the main _PID_, which will suspend current Webmin process. Important to mention, that the signal _will not_ affect control group processes, meaning that package upgrade that is happening as a child process _will continue_. However, the main Webmin process will not be able to process any requests or spawn any new processes after the time when upgrade started;
     * After setup is finished we will send _SIGCONT_ on the main _PID_, un-pausing it and then _SIGHUP_ making Webmin to re-read configuration completely;
     * By pausing the main processes we _must_ make sure that Webmin update will not be interrupted in any way, for example while Webmin is upgrading, another user on the system must not be able to start a new child process, like creating a new domain in Virtualmin, which will try to restart Webmin afterwards, that most likely will cause issues;

-----

As the **proof of concept**, I have create development repos (_deb_/_rpm_) with both Webmin 1.995 that includes all current patches from _master_ and Webmin 1.996 with all patches from this _PR_.

The devel repo can be set up as easy as:

**Debian/Ubuntu**:
```
echo "deb [trusted=yes] https://builds.ilia.engineer ./" > /etc/apt/sources.list.d/webmin-devel.list
```

**RHEL/Fedora/CentOS/Alma/Rocky:**

```
echo -e "[webmin-devel-noarch]\nname=Webmin and modules development\nbaseurl=https://builds.ilia.engineer\nenabled=1\ngpgcheck=0" > /etc/yum.repos.d/webmin-devel.repo
```

I encourage advanced and capable Webmin users like @swelljoe @chris001 (and anyone reading this if willing) also participate and help testing to fix this bug as soon as possible.

I recommend anyone trying on a clean system starting with Webmin 1.994. Try different distros, let's say CentOS 7 and Ubuntu 22.04, as those have different `systemd` versions and the test will be more objective.

Install Webmin 1.994 and then try upgrading to 1.995 and then 1.996 using UI.

If you don't want to setup repo, you can download builds with these links:

**Debian/Ubuntu**:

```
https://builds.ilia.engineer/webmin_1.996_all.deb
https://builds.ilia.engineer/webmin_1.995_all.deb
https://builds.ilia.engineer/webmin_1.994_all.deb
```

**RHEL/Fedora/CentOS/Alma/Rocky:**

```
https://builds.ilia.engineer/webmin-1.996-1.noarch.rpm
https://builds.ilia.engineer/webmin-1.995-1.noarch.rpm
https://builds.ilia.engineer/webmin-1.994-1.noarch.rpm
```

### DIFF
--- a/makedebian.pl
+++ b/makedebian.pl
@@ -260,11 +260,17 @@ system("chmod 755 $preinstall_file");
 open(SCRIPT, ">$postinstall_file");
 print SCRIPT <<EOF;
 #!/bin/sh
+justinstalled=1
+if [ -e "/etc/$baseproduct" ]; then
+	justinstalled=0
+fi
 inetd=`grep "^inetd=" /etc/$baseproduct/miniserv.conf 2>/dev/null | sed -e 's/inetd=//g'`
-if [ "\$1" = "upgrade" -a "\$1" != "abort-upgrade" ]; then
+if [ "\$1" = "configure" ]; then
 	# Upgrading the package, so stop the old Webmin properly
 	if [ "\$inetd" != "1" ]; then
-		/etc/$baseproduct/stop >/dev/null 2>&1 </dev/null
+		if [ -e "/etc/$baseproduct/.pre-install" ]; then
+			/etc/$baseproduct/.pre-install >/dev/null 2>&1 </dev/null
+		fi
 	fi
 fi
 cd /usr/share/$baseproduct
@@ -309,10 +315,16 @@ fi
 rm -f /var/lock/subsys/$baseproduct
 
 if [ "$inetd" != "1" ]; then
-	/etc/$baseproduct/stop >/dev/null 2>&1 </dev/null
-	/etc/$baseproduct/start >/dev/null 2>&1 </dev/null
-	if [ "\$?" != "0" ]; then
-		echo "W: $ucproduct server cannot be restarted. It is advised to restart it manually by\nrunning \\"/etc/webmin/restart-by-force-kill\\" when upgrade process is finished"
+	if [ "\$justinstalled" = "1" ]; then
+		/etc/$baseproduct/start >/dev/null 2>&1 </dev/null
+		if [ "\$?" != "0" ]; then
+			echo "E: Webmin server cannot be started. It is advised to start it manually\n   by running \\"/etc/webmin/restart-by-force-kill\\" command"
+		fi
+	else
+		/etc/$baseproduct/.post-install >/dev/null 2>&1 </dev/null
+		if [ "\$?" != "0" ]; then
+			echo "W: Webmin server cannot be restarted. It is advised to restart it manually\n   by running \\"/etc/webmin/restart-by-force-kill\\" command when upgrade process is finished"
+		fi
 	fi
 fi
 
@@ -361,10 +373,10 @@ if [ "\$1" != "upgrade" -a "\$1" != "abort-upgrade" ]; then
 	if [ "\$?" = 0 ]; then
 		# Package is being removed, and no new version of webmin
 		# has taken it's place. Run uninstalls and stop the server
+		/etc/$baseproduct/stop >/dev/null 2>&1 </dev/null
 		if [ "$product" = "webmin" ]; then
 			(cd /usr/share/$baseproduct ; WEBMIN_CONFIG=/etc/$baseproduct WEBMIN_VAR=/var/$baseproduct LANG= /usr/share/$baseproduct/run-uninstalls.pl) >/dev/null 2>&1 </dev/null
 		fi
-		/etc/$baseproduct/stop >/dev/null 2>&1 </dev/null
 		/bin/true
 	fi
 fi
--- a/setup.pl
+++ b/setup.pl
@@ -700,13 +700,13 @@ else {
 		# Pre install
 		open(PREINSTT, ">$config_directory/.pre-install");
 		print PREINSTT "#!/bin/sh\n";
-		print PREINSTT "$systemctlcmd kill --signal=SIGSTOP --kill-who=main webmin\n";
+		#print PREINSTT "$systemctlcmd kill --signal=SIGSTOP --kill-who=main webmin\n";
 		close(PREINSTT);
 
 		# Post install
 		open(POSTINSTT, ">$config_directory/.post-install");
 		print POSTINSTT "#!/bin/sh\n";
-		print POSTINSTT "$systemctlcmd kill --signal=SIGCONT --kill-who=main webmin\n";
+		#print POSTINSTT "$systemctlcmd kill --signal=SIGCONT --kill-who=main webmin\n";
 		print POSTINSTT "$systemctlcmd kill --signal=SIGHUP --kill-who=main webmin\n";
 		close(POSTINSTT);
 
--- a/setup.pl
+++ b/setup.pl
@@ -143,7 +143,9 @@ if ($upgrading) {
 			system("$config_directory/stop.bat >/dev/null 2>&1");
 			}
 		else {
-			system("$config_directory/stop >/dev/null 2>&1");
+			if (-r "$config_directory/.pre-install") {
+				system("$config_directory/.pre-install >/dev/null 2>&1");
+				}
 			}
 		}
 
@@ -540,7 +542,7 @@ if ($os_type eq "windows") {
 	}
 else {
 	
-	# Re-generating main	
+	# Re-generating main scripts
 	
 	# Start main
 	open(START, ">$config_directory/.start-init");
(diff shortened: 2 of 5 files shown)


setup.pl
```diff
@@ -674,11 +697,26 @@
 		print RELOADD "$systemctlcmd reload webmin\n";
 		close(RELOADD);
 
+		# Pre install
+		open(PREINSTT, ">$config_directory/.pre-install");
+		print PREINSTT "#!/bin/sh\n";
+		print PREINSTT "$systemctlcmd kill --signal=SIGSTOP --kill-who=main webmin\n";
```
jcameron, line 703: Please remove this sending of STOP / CONT signals from this PR, unless it is 100% necessary to fix the restart issue. I understand why we want it in theory, but it should be a separate PR.

setup.sh
```diff
@@ -696,27 +701,34 @@ if [ -x "$systemctlcmd" ]; then
...
 	echo "$systemctlcmd stop webmin" >>$config_dir/restart-by-force-kill
 	echo "$config_dir/.stop-init --kill >/dev/null 2>&1" >>$config_dir/restart-by-force-kill
 	echo "$systemctlcmd start webmin" >>$config_dir/restart-by-force-kill
 	# Reload systemd
-	echo "#!/bin/sh" >>$config_dir/reload
+	echo "#!/bin/sh" >$config_dir/reload
 	echo "$systemctlcmd reload webmin" >>$config_dir/reload
+	# Pre-install on systemd
+	echo "#!/bin/sh" >$config_dir/.pre-install
+	echo "$systemctlcmd kill --signal=SIGSTOP --kill-who=main webmin" >>$config_dir/.pre-install
```
jcameron, line 722: Also remove SIGSTOP here

makedebian.pl
```diff
@@ -260,11 +260,17 @@
 open(SCRIPT, ">$postinstall_file");
 print SCRIPT <<EOF;
 #!/bin/sh
+justinstalled=1
```
jcameron, line 263: Is justinstalled set to 1 when Webmin is installed for the first time?

setup.sh
```diff
@@ -656,20 +655,26 @@ echo "    (ps axf | grep \"webmin\/miniserv\.pl\" | awk '{print \"kill -9 -- -\"
...
 echo "pidfile=\`grep \"^pidfile=\" $config_dir/miniserv.conf | sed -e 's/pidfile=//g'\`" >>$config_dir/.reload-init
 echo "kill -USR1 \`cat \$pidfile\`" >>$config_dir/.reload-init
-
-chmod 755 $config_dir/.stop-init $config_dir/.start-init $config_dir/.restart-init $config_dir/.restart-by-force-kill-init $config_dir/.reload-init
+# Pre install
+echo "#!/bin/sh" >$config_dir/.pre-install
+echo "$config_dir/.stop-init" >>$config_dir/.pre-install
+# Post install
+echo "#!/bin/sh" >$config_dir/.post-install
+echo "$config_dir/.start-init" >>$config_dir/.post-install
```
jcameron, line 675: Why does .post-install call .start-init here, when in setup.pl it sends a HUP signal below on line 726?

makedebian.pl
```diff
@@ -309,10 +315,16 @@
...
-	/etc/$baseproduct/start >/dev/null 2>&1 </dev/null
-	if [ "\$?" != "0" ]; then
-		echo "W: $ucproduct server cannot be restarted. It is advised to restart it manually by\nrunning \\"/etc/webmin/restart-by-force-kill\\" when upgrade process is finished"
+	if [ "\$justinstalled" = "1" ]; then
+		/etc/$baseproduct/start >/dev/null 2>&1 </dev/null
+		if [ "\$?" != "0" ]; then
+			echo "E: Webmin server cannot be started. It is advised to start it manually\n   by running \\"/etc/webmin/restart-by-force-kill\\" command"
+		fi
+	else
+		/etc/$baseproduct/.post-install >/dev/null 2>&1 </dev/null
```
jcameron, line 324: I agree with the idea of sending a HUP signal here instead of doing stop/start , but how does .post-install send a HUP signal in all cases? Don't forget that this has to work even for systems without systemd

setup.pl
iliajie: This is not strictly necessary, and most likely would work most of the times. However, removing it may be risky, as 
 the main process will be accessible (as explained earlier). Also, if we removed it the new logic won't replicate old behaviour, where Webmin process was stopped before running upgrade.

I don't disagree that it could be a separate PR, however I cannot agree, that this change should go to the release version after upcoming (like 1.997).

makedebian.pl
iliajie: Rather `justinstalled` is unset the moment after it's set, in case `/etc/webmin` exists.

setup.sh
iliajie: This code (620-675) is for the older systems only (without `systemd`), while code on lines 700-726 is only for new systems, therefor new systems (with `systemd`) will have `SIGHUP` line, while older systems will call initial start script.

This behavior replicates old logic, where at first on older systems you called `/etc/webmin/stop` (and now I call `SIGSTOP` on new systems with `systemd`) and after setup you called `/etc/webmin/start` (and I call `SIGCONT` and `SIGHUP` on newer systems).

That said, the code in files `/etc/webmin/.pre-install` and `/etc/webmin/.post-install` on old systems and new systems won't be the same.

makedebian.pl
iliajie: I know and I didn't forget. I have answered your question in the comment above, here - https://github.com/webmin/webmin/pull/1677#discussion_r912508947

makedebian.pl
jcameron: Ok, but what it's detecting is whether webmin is being installed for the first time?

setup.pl
jcameron: Let's make it a separate PR. Also, if another user does access Webmin during the upgrade process, will this actually cause the upgrade to fail?

makedebian.pl
iliajie: Yes, and it's serving just a simple purpose to know whether to run `/etc/webmin/start` or `/etc/webmin/.post-install` (please don't forget that `../start` and `../.post-install` files are different on `systemd` systems).

jcameron: Got it, makes sense

iliajie: .. i.e. in case of older systems, it wouldn't matter, and in practice `../start` would be called in both cases (install/upgrade) .. but on `systemd` it's different.

setup.pl
iliajie: > Let's make it a separate PR. 

.. ohh, okay .. let's ..

> Also, if another user does access Webmin during the upgrade process, will this actually cause the upgrade to fail?

If the process is on-pause or stopped another user won't even be able to access it. But even if it's not suspended, I don't think that just accessing Webmin would do any harm.

setup.sh
jcameron: Ok, I think I see. As I mentioned, the only change I'd recommend is remove the STOP/CONT signals.

setup.pl
iliajie: There you go - https://github.com/webmin/webmin/pull/1677/commits/6f5bb3947f3eee55c338d9902161c928a4cd8ee4.

setup.sh
iliajie: Removed for now.

makedebian.pl
jcameron: Ok I see now .. thanks!

jcameron:
Merging this now .. I'll run some tests, and let you know how it goes

iliajie:
Alright, thanks. Please consider this PR too - https://github.com/webmin/webmin/pull/1679.

PR #1679, opened 2022-07-03
Fix to send `STOP` and `CONT` signals in upgrade

iliajie:
Jamie, test this PR too. I believe it must go to upcoming release.

### DIFF
--- a/setup.pl
+++ b/setup.pl
@@ -700,13 +700,13 @@ else {
 		# Pre install
 		open(PREINSTT, ">$config_directory/.pre-install");
 		print PREINSTT "#!/bin/sh\n";
-		#print PREINSTT "$systemctlcmd kill --signal=SIGSTOP --kill-who=main webmin\n";
+		print PREINSTT "$systemctlcmd kill --signal=SIGSTOP --kill-who=main webmin\n";
 		close(PREINSTT);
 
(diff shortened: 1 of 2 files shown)


iliajie:
Reference with explanations - https://github.com/webmin/webmin/pull/1677#issue-1292328735.

jcameron:
Code-wise this looks fine, but I still don't quite understand what will happen during an upgrade if we don't SIGSTOP the miniserv process? Even if a sub-process gets spawned, will it somehow interrupt the upgrade? If so, how?

iliajie:
> Code-wise this looks fine, but I still don't quite understand what will happen during an upgrade if we don't SIGSTOP the miniserv process?

That would be fair if I asked the same question to you -- what will happen during the upgrade if you don't call `/etc/webmin/stop`. 🙂 Why not just call `SUGHUP` alone after Webmin upgrade is finished? 

I didn't try hard to come up with something to actually break it but I think it's _theoretically_ possible, and my educated guess is that the chances of breaking it are in direct ratio to the number of currently logged in admin users working in Webmin.

Speaking _practically_ nothing should ever happen unless some other admin calls _reboot_ or _poweroff_ during update.

I don't mind leaving this change behind and going to making new releases.

PR #1690, opened 2022-07-16
Fix BIND configs for modern distros

iliajie:
This is it. Jamie, have a look, please.


iliajie:
Will `bind8/config-debian-linux-10.0-ALL` cover Ubuntus?

iliajie:
I would also prefer to change to `systemctl` commands for other main modules used with Virtualmin?

bind8/config-AlmaLinux-6.0-ALL
```diff
@@ -42,15 +40,13 @@ keygen=dnssec-keygen
 signzone=dnssec-signzone
 checkconf=named-checkconf
 other_slaves=1
-restart_cmd=restart
 keygen=dnssec-keygen
 tmpl_dnssec=0
 dnssec_period=21
 tmpl_dnssec_dt=1
-dnssectools_conf=/etc/dnssec-tools/dnssec-tools.conf
```
jcameron, line 50: Why remove these lines?

iliajie: I assumed that this no longer relevant nowadays. Isn't it?

jcameron: Maybe, but let's fix one  thing at a time in this PR. I think the main issue is using the incorrect stop command?

iliajie: > but let's fix one thing at a time in this PR.

Duh, thanks for reminding me, Jamie! I'm about to get used to fixing one issue at a time!

iliajie: > I think the main issue is using the incorrect stop command?

Yes, _stop_, _start_ and _reload_ command for new distros should use `systemctl` command.

jcameron:
Thanks for the fix!

iliajie:
Do you want me to add familiar changes for all other major modules we use with Virtualmin?

jcameron:
Yes, as long as we can be sure that it's applied only for distro versions that use `systemd`

PR #1691, opened 2022-07-20
Fix more modules config files

iliajie:
Fix more modules config files


apache/config-AlmaLinux-7.0-ALL
```diff
@@ -1,4 +1,4 @@
-show_list=0
+show_list=1
```
jcameron, line 1: Why make this change?

iliajie:
Sorry for not putting this to a separate commit but it's just cosmetics. The point is to have Apache display hosts as a table, which you can type and search for a domain name. Also there is a button to open domain. I find it much more useful then icons, especially if there a lot of domains.

Do you mind changing it?

iliajie:
I have taken `show_list` out from this PR. Tell me if you're okay to switching to table layout. I will do it later.

iliajie:
I just realized I need to add CentOS Stream support for configs too!

apache/config-debian-linux-7.0-9.0
```diff
@@ -1,4 +1,4 @@
-show_list=0
+show_list=1
```
jcameron, line 1: Why change show_list in this PR?

iliajie: A mistake.. Let me clean it.

iliajie: Fixed.

PR #1697, opened 2022-07-22
Add `fdisk` GPT improvements

iliajie:
1. Do not use `fdisk` command in `Parted` mode
 2. Add `fdisk` GPT tags
 3. Parse `fdisk` output with GPT disk type
 4. Do not display selects and inputs if mounted (cannot be edited)
 5. Do not redirect even if one disk, as module should always display consistent, no confusing output
 6. Do not print elements if there is no actual data


jcameron:
Can you confirm that this still works fine even on older systems with only the `fdisk` command?

iliajie:
Well, this was _only_ tested with `fdisk`. Although, I didn't run actual commands on the disk to create or change partitions. 

Considering it worked for you in the past with `fdisk` and GPT partition tables?

iliajie:
**GPT with `parted`:**
image

image

**GPT with `fdisk`:**

image

image

**DOS (either tool):**
image
image

jcameron:
Can you also test this on some older Linux distros, like CentOS 6 ?

iliajie:
Yes, will do in a fee minutes.

iliajie:
I have MSDOS partition in my old CentOS 6 instance and it works just fine.

PR #1702, opened 2022-07-29
Fix to hide time selection for Linux systems

iliajie:
Fix to hide time format selection for Linux systems.

### DIFF
--- a/time/time-lib.pl
+++ b/time/time-lib.pl
@@ -263,34 +263,27 @@ return &has_command("hwclock") &&
        !&running_in_openvz() && !&running_in_zone();
 }
 
-# config_preserve()
-# Returns a list of config options that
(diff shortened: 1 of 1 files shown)


iliajie:
Alright, I have reverted the first change in this PR partially, as now I [added @commit/de3b01d](https://github.com/webmin/webmin/commit/de3b01d083eb7020c02573ec28ba15f8eca4e4de) an ability to preserve hidden options values, to make sure that the behavior of a module will not be changed.

jcameron:
Looks mostly reasonable, but see my comments on the other commit.

iliajie:
I commented back on that commit.

iliajie:
Are we merging this?

jcameron:
I still have questions about this config override API ..

PR #1711, opened 2022-08-05
Fix to properly transition from `non-systemd` to `systemd` instance

iliajie:
I have come to understanding that it's actually possible to transition/restart old Webmin installs (before Webmin 1.993), even those were started outside of `systemd` (with an old version of `/etc/webmin/start` script) to using a new `webmin-systemd` unit file seamlessly.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -618,6 +618,9 @@ if [ ! -f "$config_dir/.pre-install" ]; then
 	killmodenonesh=1
 fi
 
+# Test if we have systemd system
+systemctlcmd=`which systemctl 2>/dev/null`
+
 # Re-generating main scripts
 echo "Creating start and stop init scripts .."
 # Start main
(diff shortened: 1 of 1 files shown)


iliajie:
.. the solution is, in case of an upgrade from an older Webmin versions is to first _stop_ Webmin using old init script (release the socket) and the _start_ it using `systemctl` command, which will work without an error.

setup.sh
```diff
@@ -680,11 +680,6 @@ if [ "$killmodenonesh" = "1" ] && [ -x "$systemctlcmd" ]; then
 	if [ "$ancient_version" = "1" ]; then
 		echo "$config_dir/.stop-init" >>$config_dir/.reload-init
 		echo "$config_dir/start" >>$config_dir/.reload-init
-		# Reset this file after the call in setup.sh
```
iliajie: The test case to try:

```Bash
#!/bin/sh
echo Reloading Webmin server in /usr/share/webmin
pidfile=`grep "^pidfile=" /etc/webmin/miniserv.conf | sed -e 's/pidfile=//g'`
kill -USR1 `cat $pidfile`
/etc/webmin/.stop-init
/etc/webmin/start
echo "#!/bin/sh" >/etc/webmin/.reload-init
echo "echo Reloading Webmin server in /usr/share/webmin" >>/etc/webmin/.reload-init
echo "pidfile=\`grep \"^pidfile=\" /etc/webmin/miniserv.conf | sed -e 's/pidfile=//g'\`" >>/etc/webmin/.reload-init
echo "kill -USR1 \`cat \$pidfile\`" >>/etc/webmin/.reload-init
```

PR #1712, opened 2022-08-07
Add a better way to run conditional command

iliajie:
Add a better way to run conditional command as initially intended in https://github.com/webmin/webmin/pull/1711/commits/162de2bda0557452edaa34058b0727e4e64dd1e1.

### DIFF
--- a/setup.pl
+++ b/setup.pl
@@ -910,13 +910,12 @@ if (!$ENV{'nostart'}) {
 			$start_cmd_extra = "$config_directory/.reload-init-systemd";
 			if (-r $start_cmd_extra) {
 				$start_cmd .= " ; $start_cmd_extra";
-				$start_cmd_extra = -1;
 				}
 			}
 		print "Attempting to $action Webmin web server ..\n";
 		$ex = system($start_cmd);
 		unlink($start_cmd_extra)
-			if ($start_cmd_extra == -1);
+			if (-r $start_cmd_extra);
 		if ($ex) {
 			&errorexit("Failed to $action web server!");
 			}
--- a/setup.pl
+++ b/setup.pl
@@ -534,7 +534,7 @@ if (!$ENV{'noperlpath"'} && $os_type ne 'windows') {
 	print ".. done\n";
         print "\n";
 	}
-
+my $systemctlcmd = &has_command('systemctl');
 print "Creating start and stop scripts ..\n";
 if ($os_type eq "windows") {
 	open(START, ">>$config_directory/start.bat");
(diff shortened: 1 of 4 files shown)


setup.pl
```diff
@@ -894,8 +905,18 @@
...
+			$start_cmd_extra = "$config_directory/.reload-init-systemd";
+			if (-r $start_cmd_extra) {
+				$start_cmd .= " ; $start_cmd_extra";
+				$start_cmd_extra = -1;
+				}
+			}
 		print "Attempting to $action Webmin web server ..\n";
 		$ex = system($start_cmd);
+		unlink($start_cmd_extra)
+			if ($start_cmd_extra == -1);
```
jcameron, line 919: How does unlinking a file named -1 work?

iliajie: I just used the same variable name to set it to -1. It will be set to -1 only if related code is used. (which is set just a few lines above). It could be anything in place of -1.

iliajie: Duh, you're right!! I just realized that I unset the actual file name! I will fix that!

setup.sh
```diff
@@ -674,12 +674,15 @@ echo "#!/bin/sh" >$config_dir/.reload-init
...
 echo "kill -USR1 \`cat \$pidfile\`" >>$config_dir/.reload-init
+# Switch to systemd from init (intermediate)
 if [ "$killmodenonesh" = "1" ] && [ -x "$systemctlcmd" ]; then
 	current_version=`cat "$config_dir/version" 2>/dev/null`
 	ancient_version=`echo $current_version 1.994 | awk '{if ($1 < $2) print 1; else print 0}'`
 	if [ "$ancient_version" = "1" ]; then
-		echo "$config_dir/.stop-init" >>$config_dir/.reload-init
-		echo "$config_dir/start" >>$config_dir/.reload-init
+		echo "#!/bin/sh" >$config_dir/.reload-init-systemd
+		echo "$config_dir/.stop-init" >>$config_dir/.reload-init-systemd
```
jcameron, line 683: Doesn't this need to be `.stop-init --kill` so that it works if the PID file is missing?

setup.pl
iliajie: Fixed.

setup.sh
iliajie: Oh, no no! If we use `--kill` it will take down the whole process. It will be handled correctly either way for these rare cases, i.e. a user won't see a breakage at all. At least, great majority of users won't see an actual breakage at all. And I highly doubt that anyone who had Webmin actually finished to upgrade (files placed on the disk) and using Authentic Theme will ever see a breakage, even in case Webmin wasn't restarted correctly.

Transition from `init` to `systemd` is complicated because we don't know how older versions of Webmin (before 1.994) were started (using older `/etc/webmin/start` script or using `systemctl` command). In addition there is also to consider this odd situation with missing pid file.

We simply cannot and must not use `--kill`.

jcameron: What about the case where the PID file is missing though? Isn't --kill needed to handle that, because `.stop-init` contains : 

```
  if [ "$1" = "--kill" ]; then
    (ps axf | grep "webmin\/miniserv\.pl" | awk '{print "kill -9 -- -" $1}' | bash) 2>/dev/null
  fi
```

iliajie: This is meant to be only called from a console. We cannot call it from the process itself. Webmin like no other, restarts itself and this is where there great complexity is coming from.

PR #1715, opened 2022-08-11
Add support to install multiple versions of Webmin with `systemd`

iliajie:
Jamie, check this out, as requested earlier today.

### DIFF
--- a/init/atboot.pl
+++ b/init/atboot.pl
@@ -12,7 +12,7 @@ my $kill = &has_command("kill");
 
 if ($init_mode eq "osx") {
 	# Darwin System
-	&enable_at_boot("webmin", "Webmin administration server",
+	&enable_at_boot("$product", "Webmin administration server",
 			"$config_directory/.start-init >/dev/null 2>&1 </dev/null",
 			"$config_directory/.stop-init");
 	}
--- a/init/atboot.pl
+++ b/init/atboot.pl
@@ -12,7 +12,7 @@ my $kill = &has_command("kill");
 
 if ($init_mode eq "osx") {
 	# Darwin System
-	&enable_at_boot("$product", "Webmin administration server",
+	&enable_at_boot($product, "Webmin administration server",
 			"$config_directory/.start-init >/dev/null 2>&1 </dev/null",
 			"$config_directory/.stop-init");
 	}
--- a/setup.pl
+++ b/setup.pl
@@ -405,7 +405,6 @@ else {
 			'logfile' => "$var_dir/miniserv.log",
 			'errorlog' => "$var_dir/miniserv.error",
 			'pidfile' => "$var_dir/miniserv.pid",
-			'bootscript' => $bootscript,
 			'logtime' => 168,
 			'ppath' => $ppath,
 			'ssl' => $ssl,
--- a/setup.pl
+++ b/setup.pl
@@ -84,9 +84,13 @@ print "Webmin uses separate directories for configuration files and log files.\n
 print "Unless you want to run multiple versions of Webmin at the same time\n";
 print "you can just accept the defaults.\n";
 print "\n";
-print "Config file directory [/etc/webmin]: ";
+my $envetcdir = $ENV{'config_directory'} || "/etc/webmin";
+print "Config file directory [$envetcdir]: ";
 if ($ENV{'config_directory'}) {
 	$config_directory = $ENV{'config_directory'};
+	print "$envetcdir\n";
+	print ".. predefined\n";
+	$envetcdirexists = 1;
 	}
 else {
 	chop($config_directory = <STDIN>);
@@ -101,7 +105,7 @@ if (!-d $config_directory) {
 		&errorexit("Failed to create directory $config_directory");
 	}
 if (-r "$config_directory/config") {
-	print ".. found\n";
+	print ".. found\n" if (!$envetcdirexists);
 	$upgrading = 1;
 	}
 
--- a/setup.pl
+++ b/setup.pl
@@ -1012,16 +1012,15 @@ return @rv;
 sub make_dir_recursive_local
 {
 my ($dir, $mod) = @_;
-my @folders = split /\//, $dir;
-map {
-    if ($_) {
-        if (mkdir $_) {
-            chmod($mod, $_) if ($mod && -d $_);
-            }
-        chdir $_;}
-    else {
-        chdir '/';
+my @folders = split(/\//, $dir);
+my $folder_created;
+foreach my $folder (@folders) {
+    next if (!$folder);
+    $folder_created .= "/$folder";
+    if (mkdir($folder_created)) {
+        chmod($mod, $folder_created)
+            if ($mod && -d $folder_created);
         }
-} @folders;
+    }
 return -d $dir;
 }
--- a/setup.pl
+++ b/setup.pl
@@ -101,7 +101,7 @@ if ($config_directory !~ /^([a-z]:)?\//i) {
 	&errorexit("Config directory must be an absolute path");
 	}
 if (!-d $config_directory) {
-	mkdir($config_directory, 0755) ||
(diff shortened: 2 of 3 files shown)


init/atboot.pl
```diff
@@ -12,7 +12,7 @@
 
 if ($init_mode eq "osx") {
 	# Darwin System
-	&enable_at_boot("webmin", "Webmin administration server",
+	&enable_at_boot("$product", "Webmin administration server",
```
jcameron, line 15: This can be just $product instead of "$product"

setup.pl
```diff
@@ -387,6 +388,7 @@
 			'logfile' => "$var_dir/miniserv.log",
 			'errorlog' => "$var_dir/miniserv.error",
 			'pidfile' => "$var_dir/miniserv.pid",
+			'bootscript' => "$bootscript",
```
jcameron, line 391: No quotes needed here either

init/atboot.pl
iliajie: Then we need to test that user defined env variable to comply to a certain format .. what if that would be set to "webmin 1.999-4" (intentionally or by mistake?).

jcameron: What I meant was, in Perl there's no difference between `$foo` and `"$foo"`

iliajie: Oh, that's Perl, right. Removed.

setup.pl
```diff
@@ -121,8 +126,17 @@
...
 	$var_directory = $var_dir;
+	$ENV{'WEBMIN_VAR'} = $var_directory;
 	close(VAR);
 
+	# Get current bootscript name
+	if (-r "$config_directory/bootscript-name") {
+		open(BOOTVAR, "$config_directory/bootscript-name");
+		chop($bootscript = <BOOTVAR>);
+		close(BOOTVAR);
+		$bootscript ||= ($ENV{'bootscript'} || "webmin");
```
jcameron, line 137: Shouldn't this be moved outside the if statement, so $bootscript is always set?

setup.pl
```diff
@@ -991,3 +1010,19 @@ sub files_in_dir
...
+sub make_dir_recursive_local
+{
+my ($dir, $mod) = @_;
+my @folders = split /\//, $dir;
+map {
+    if ($_) {
+        if (mkdir $_) {
+            chmod($mod, $_) if ($mod && -d $_);
+            }
+        chdir $_;}
```
jcameron, line 1022: I don't think calling chdir in here is a good idea, as that's a global operation that could effect the rest of the script.

setup.sh
```diff
@@ -138,6 +160,14 @@ if [ "$upgrading" = 1 ]; then
 	# Get current var path
 	var_dir=`cat $config_dir/var-path`
 
+	# Get current bootscript
+	if [ -r "$config_dir/bootscript-name" ]; then
+		bootscript=`cat $config_dir/bootscript-name`
+		if [ "$bootscript" = "" ]; then
```
jcameron, line 166: Also this should be moved outside the if statement

setup.sh
```diff
@@ -230,13 +265,19 @@ else
 		exit 3
 	fi
 	if [ ! -d $var_dir ]; then
-		mkdir $var_dir
+		mkdir -p $var_dir
 		if [ $? != 0 ]; then
-			echo "ERROR: Failed to create directory $var_dir"
-			echo ""
-			exit 3
+			mkdir $var_dir
```
jcameron, line 270: Why is a second mkdir needed?

setup.sh
```diff
@@ -489,6 +531,7 @@ else
 	echo "logfile=$var_dir/miniserv.log" >> $cfile
 	echo "errorlog=$var_dir/miniserv.error" >> $cfile
 	echo "pidfile=$var_dir/miniserv.pid" >> $cfile
+	echo "bootscript=$bootscript" >> $cfile
```
jcameron, line 534: Does this need to be stored in miniserv.conf and bootscript-name ?

iliajie: The variable stored in`miniserv.conf` is not something we currently use. I thought just to store it, just in case.

setup.sh
iliajie: The second `mkdir` is needed as a presumption that some OS may not support the `-p` flag.

setup.pl
iliajie: Then we could store the current `chdir` and restore it afterwards.

setup.pl
iliajie: No, as this is a global variable. It will be stored either way.

setup.pl
jcameron: But why chdir at all, rather than using absolute paths?

setup.sh
jcameron: I checked a few OS man pages, and I couldn't see any that don't support -p.

setup.sh
jcameron: Don't bother, we can read it from that other file if needed.

setup.pl
iliajie: To create a path dir-by-dir, without using `system` and `mkdir -p`.

jcameron: I mean why not call mkdir on the full path?

setup.sh
iliajie: Alright, thank you for checking on that. I have fixed that.

setup.sh
iliajie: Okay, I'm fine with removing it!

setup.sh
iliajie: All of these variables are not block-scoped. These are visible globally!

setup.pl
iliajie: We cannot simply call `mkdir` on the full path as it won't create parents, this is not a pleasant user experience.

setup.sh
jcameron: But what if the `bootscript-name` file doesn't exist? `$bootscript` will never be set to `webmin`

setup.pl
jcameron: No I mean, instead of using `chdir` to enter each new directory, just always call `mkdir` on the full path.

setup.sh
iliajie: Oh, I see where confusion is coming from. It will. Please check [line 12](https://github.com/webmin/webmin/pull/1715/files/dcdf0c59ae16449e48d06acf8831fd1338d0cf46#diff-4209d788ad32c40cbda3c66b3de47eefb929308ca703bb77a6382625986add17R12-R15) on this PR.

setup.pl
iliajie: Oh, you mean, if the path is `/webmin/1.999-3/data/etc`, you want me to call it:

```
mkdir /webmin
mkdir /webmin/1.999-3
mkdir /webmin/1.999-3/data
mkdir /webmin/1.999-3/data/etc
```

.. without using `chdir` at all?

iliajie: Nice suggestion - fixed https://github.com/webmin/webmin/commit/82937d72963ff8309bd4914690dbf1f3b85aa168.

setup.sh
jcameron: Oh I see now ... yeah, that's better.

setup.pl
jcameron: Thanks! That works..

setup.pl
jcameron: But is $bootscript set to some default higher up if `bootscript-name` doesn't exist?

setup.sh
iliajie: > Oh I see now ... yeah, that's better.

It was already there, I didn't just add it. 🙂

setup.pl
iliajie: Yes, it's the same as in `setup.sh`, and already set on the line 12 of `setup.pl` ..

jcameron: Ok cool

PR #1716, opened 2022-08-13
Dev/redirect no message

iliajie:
This fix to redirect without showing a message if JavaScript is enabled, or show a message if there is no JavaScript on the browser support.

### DIFF
--- a/miniserv.pl
+++ b/miniserv.pl
@@ -1452,8 +1452,14 @@ elsif ($reqline !~ /^(\S+)\s+(.*)\s+HTTP\/1\..$/) {
 				"This web server is running in SSL mode. ".
 				"Try the URL <a href='$url'>$url</a> instead.</noscript>".
 				"<script>".
+				"var noscript = document.querySelector('noscript');".
+				"var redirect = '<tt>This web server is running in SSL mode.<br>".
+				"Redirecting to the URL ".
+				"<a style=\"color: #f12b2b\" href=\"https://".
+				"'+location.host+'\">https://'+location.host+'</a> instead ...</tt>';".
+				"noscript.insertAdjacentHTML('afterend', redirect);".
 				"if (location.protocol != 'https:') {".
-				"  location.protocol = 'https:';".
+				"  setTimeout(function(){location.protocol = 'https:'},4000);".
 				"}".
 				"</script>",
 				0, 1, 1);
--- a/miniserv.pl
+++ b/miniserv.pl
@@ -1399,12 +1399,19 @@ elsif ($reqline !~ /^(\S+)\s+(.*)\s+HTTP\/1\..$/) {
 			}
 		local $url = $wantport == 443 ? "https://$urlhost/"
 					      : "https://$urlhost:$wantport/";
+		local $jsurl = $config{'musthost'} ?
+				                   $url :
+				                   "https://'+location.host+'";
+		local $jsredir = $config{'musthost'} ?
+				                   "location.href='$url'" :
+				                   "location.protocol='https:'";
 		&http_error(200, "Document follows",
 			"This web server is running in SSL mode. ".
-			"Try the URL <a href='$url'>$url</a> instead.".
+			"Trying to redirect to <a href='$url'>$url</a> instead URL ...".
 			"<script>".
 			"if (location.protocol != 'https:') {".
-			"  location.protocol = 'https:';".
+			"  document.querySelector('a').href='".$jsurl."';document.querySelector('a').innerText='".$jsurl."';".
+			"".$jsredir."".
 			"}".
 			"</script>",
 			0, 1);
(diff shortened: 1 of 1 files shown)


jcameron:
What effect will this have in practice? I found that with the current code, the error message isn't even visible as it's immediately redirected away ..

iliajie:
It depends on the browser. The error message may be shown. Redirect may have a delay. Try it like 10 times in a row to see what I mean.

jcameron:
I'm worried though that if the JS does fail entirely, the user will just see an empty page and have no idea what to do ..

iliajie:
Yeah, I thought of that too. Although, I don't see how it could fail.. Nevertheless, maybe we should make it even more user friendly, and first print:

```
Redirecting to SSL version (https://example.com:10000) ..
```

.. and if that fails a user would be able to just click the link ..

jcameron:
> and if that fails a user would be able to just click the link

But that's basically how it works before this PR?

iliajie:
Close but it wanted you to click the link. And as you mentioned the host detection wasn't great.

I think we can just leave it as it is now on the PR. I don't think it JS part can fail, really.

iliajie:
Jamie, I have changed it a little more. It will print a nice message and redirect automatically:

image

..and if a user doesn't have Javascirpt enabled, it will show a message and will suggest clicking a link (as before):

image

jcameron:
Honestly I think this change is just getting too complex for the problem it's solving. Let's just leave the current code, which works fine even if occasionally the user sees the redirect message for a second.

jcameron:
Also, I wonder if there's an HTTP header that we can set which provides a hint that the browser should upgrade to HTTPS?

iliajie:
> Also, I wonder if there's an HTTP header that we can set which provides a hint that the browser should upgrade to HTTPS?

As far as I'm aware there are no ways for a browser to automatically redirect despite of headers sent. A browser could send a header to suggest a server to switch to `HTTPS` .. but this is not what we need.

> Honestly I think this change is just getting too complex for the problem it's solving. Let's just leave the current code, which works fine even if occasionally the user sees the redirect message for a second.

Well, the change is straight forward. It uses requested URL to print URL and redirect to `HTTPS` a bit more nicely. I have added additional change to make sure that redirect would still work even if the code above your initial `location.protocol != 'https:'` fails. Also, I believe that `hide_admin_url` and `ssl_redirect` shouldn't exist at all in server running in SSL mode (if `ssl=1`). This is a very dangerous practice.

---

❗ What we really _have_ do (separate from this PR or in it) is to make sure that we add **`Strict-Transport-Security`** header when SSL mode is enabled to make sure users won't become a subject of MITM by accidentally accessing the server using `http`! That header (which has expiration time (imho should be set to max possible)) won't allow a user to access non-ssl version of the website in SSL enabled mode, _**even if**_ you manually put `http://host:10000` .. also (with using this header), on initial redirect we need to reset all cookies!

My suggestion drop completely `hide_admin_url` and `ssl_redirect`, keep JS redirect (yours which is simpler or mine which is a bit more complex but it provides much better UX (and also resilient to failures with the latest patch)) and most importantly set `Strict-Transport-Security` header on first successful HTTPS connection, resetting all cookies that were sent using HTTP.

jcameron:
I don't see any security risk of users accessing Webmin in HTTP mode, because miniserv is unable to parse the request anyway, so it will never serve any content. 

For at least the 2.0 release, I'd like to keep the currently checked-in code which does a JS redirect in a simple way. And I agree that `hide_admin_url` and `ssl_redirect` can go .. I'll submit changes to do that.

iliajie:
> I don't see any security risk of users accessing Webmin in HTTP mode, because miniserv is unable to parse the request anyway, so it will never serve any content.

You are right, although the problem is not in `miniserv` serving a request. This is about the client connecting via HTTP and leaking secure data, like session cookie. Is highly susceptible to MITMA!

This patch in PR https://github.com/webmin/webmin/pull/1719 will solve this problem. After a user connects to SSL version of `miniserv` the first time, it sets `Strict-Transport-Security` header telling the browser to never send HTTP request. In other words, if you have logged in to Webmin using SSL and then a user tries to connect with `http://example.com:10000`, then the browser will never request `HTTP` variant of the `miniserv` at all, and `HTTPS` will be used straight forward.

You can test it by doing the following:

  1. Get Webmin SSL enabled instance with proper SSL certificate
  2. Open incognito tab in Chrome
  3. Paste `http://example.com:10000` and hit enter
  4. See the redirect with 200 document follows
  5. Try again and see what happens with set `Strict-Transport-Security` headers and without

We can continue a discussion on the new PR. I will keep this PR opened for the future to handle very first redirect more nicely.

jcameron:
Aren't we protected from clients sending cookies in the wrong mode via the `secure` tag that's added to cookies here : https://github.com/webmin/webmin/blob/master/miniserv.pl#L4263

iliajie:
I'm not sure about that, sorry.

iliajie:
Also, it (secure cookie) doesn't protect against various downgrade attacks. The topic `secure cookie vs hsts` is discussed widely.

I suggest having both is the right choice nowadays.

iliajie:
> Although, what if a user has Webmin proxied from HTTP to HTTPS?

I assume that would be a problem of a proxy and its configuration. Besides, it wouldn't work (as it didn't work before this change) and testing cookie would throw and error.

PR #1719, opened 2022-08-17
Fix to always force HSTS in SSL

iliajie:
This is important improvement. I will provide more details [here](https://github.com/webmin/webmin/pull/1716#issuecomment-1217543466).

### DIFF
--- a/web-lib-funcs.pl
+++ b/web-lib-funcs.pl
@@ -969,6 +969,11 @@ sub PrintHeader
 {
 my ($cs, $mt) = @_;
 $mt ||= "text/html";
+my %miniserv;
+&get_miniserv_config(\%miniserv);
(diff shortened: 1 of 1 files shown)


web-lib-funcs.pl
```diff
@@ -969,6 +969,11 @@ sub PrintHeader
 {
 my ($cs, $mt) = @_;
 $mt ||= "text/html";
+my %miniserv;
+&get_miniserv_config(\%miniserv);
+if (!$gconfig{'no_strict_transport_security'} && $miniserv{'ssl'}) {
```
jcameron, line 974: Instead of reading the miniserv conf, just check if `$ENV{'HTTPS'}` equals `'ON'`

iliajie: Alright! Fixed.

PR #1727, opened 2022-08-27
Added translation in Webmin log

unknown10777:
btw what's the meaning of `__norefs=1` ?
it's at the end of `webminlog/lang/en`


webminlog/lang/zh
```diff
@@ -1,51 +1,153 @@
...
+view_files=文件更改和命令执行
+view_raw=啃生日志
+view_rawname=参数名称
+view_rawvalue=参数值
+
+acl_mods=可以查看模块里的日志
+acl_all=所有
+acl_sel=下方选中的 ..
+acl_self=只有自己
+acl_users=Can view actions by users
```
unknown10777, line 119: I uh think I may forgot this. Will correct in next pr. sorry.

jcameron:
__norefs means there are no strings that refer to other strings via $ substitutions. 

Thanks for the patch!

PR #1735, opened 2022-09-06
Fix BIND list zone names error handling when adding slave ; fix hash check

iliajie:
This PR fixes:

  1. Error handling when zone on remote cannot be listed [Ref.](https://forum.virtualmin.com/t/licence-in-3-servers-are-missconfused/116681/15)
  2. Checks that `$_` is always a hash


jcameron:
When would list_zone_names ever return something that's not a hash?

iliajie:
Staled.

PR #1743, opened 2022-09-17
Fix to use correct URL name ; consider `musthost` option

iliajie:
This PR replaces previous PR. It is more simple and works in all browsers.

This PR:

   * Considers `musthost` option
   * Prints correct _hostname_ in UI to make it look consistent, otherwise it may say redirecting to `smtp.hostname.local` and eventually redirect to `window.location.host` (which may be completely different host)

### DIFF
--- a/miniserv.pl
+++ b/miniserv.pl
@@ -1407,7 +1407,7 @@ elsif ($reqline !~ /^(\S+)\s+(.*)\s+HTTP\/1\..$/) {
 				                   "location.protocol='https:'";
 		&http_error(200, "Document follows",
 			"This web server is running in SSL mode. ".
-			"Trying to redirect to <a href='$url'>$url</a> instead URL ...".
+			"Trying to redirect to <a href='$url'>$url</a> instead ...".
 			"<script>".
 			"if (location.protocol != 'https:') {".
 			"  document.querySelector('a').href='".$jsurl."';document.querySelector('a').innerText='".$jsurl."';".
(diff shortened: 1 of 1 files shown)


miniserv.pl
```diff
@@ -1399,12 +1399,19 @@ sub handle_request
...
+				                   "location.href='$url'" :
+				                   "location.protocol='https:'";
 		&http_error(200, "Document follows",
 			"This web server is running in SSL mode. ".
-			"Try the URL <a href='$url'>$url</a> instead.".
+			"Trying to redirect to <a href='$url'>$url</a> instead ...".
 			"<script>".
 			"if (location.protocol != 'https:') {".
-			"  location.protocol = 'https:';".
+			"  document.querySelector('a').href='".$jsurl."';document.querySelector('a').innerText='".$jsurl."';".
```
jcameron, line 1413: Why the need to update the URL using javascript when we just printed it above?

iliajie: Well, if the host in the URL is forced with `musthost` (which 99% of a time isn't), then it's all fine.

However, as you mentioned earlier yourself (that host is not always detected correctly) it may print a wrong host on the redirect message, but JavaScript using `location.protocol='https:` will redirect it to whatever is in initial query.

i.e. the message can be:

image


.. and the final redirect done by JavaScript may happen to the other URL, as Webmin cannot detect host correctly.

I found this very confusing and solved this problem with this PR. So, if we say that redirect is happening to `alma9-pro.synology:10000` we cannot eventually redirect to `mail.alma9-pro.synology:10000` ..

jcameron: Ok I see now ... thanks!

iliajie: Anything else needed to merge it?

PR #1750, opened 2022-10-05
Fix to create a target directory before writing a file

iliajie:


jcameron:
No, we don't want this - callers should create any needed directories themselves.

Issue #1752, opened 2022-10-07
LetsEncrypt DNS validation fails for dynamic zones


dsbibby:
Using virtualmin to request a LetsEncrypt SSL certificate for a virtual server whose BIND DNS zone is a dynamic zone.

This virtual server is internal facing only so Web based validation isn't possible. Equally wildcard certificates require DNS validation. However the dynamic zone is not updated correctly with the current implementation so DNS validation fails.

I have a working solution which I'll submit as a PR.

swelljoe:
Based on the discussion on the linked PR, it seems like this is resolved in newer Webmin/Virtualmin versions.

PR #1753, opened 2022-10-07
Fix LetsEncrypt dynamic dns zones

dsbibby:
This PR patches the LetsEncrypt auth-hook and cleanup-hook perl scripts to add in RFC2136 based dynamic zone updates to add and delete the required TXT record for dynamic DNS zones.

Tested with both dynamic and static BIND DNS zones successfully.

Closes: #1752

### DIFF
--- a/webmin/letsencrypt-cleanup.pl
+++ b/webmin/letsencrypt-cleanup.pl
@@ -67,4 +67,11 @@ else {
 	&virtual_server::reload_bind_records($d);
 	}
 
+# Clean up any dynamic record
+use Net::DNS;
+my $update = Net::DNS::Update->new( $dname );
+$update->push( update => rr_del('_acme-challenge.'.$dname.' TXT') );       my $resolver = Net::DNS::Resolver->new();
+$resolver->nameservers('127.0.0.1');
+my $reply = $resolver->send($update);
+
 &webmin_log("letsencryptcleanup", undef, $dname);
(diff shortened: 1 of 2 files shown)


webmin/letsencrypt-cleanup.pl
```diff
@@ -67,4 +67,11 @@
 	&virtual_server::reload_bind_records($d);
 	}
 
+# Clean up any dynamic record
+use Net::DNS;
```
jcameron, line 71: What if the Net::DNS perl module isn't installed?

dsbibby: Valid point - I'd assumed it was a dependency of webmin already as it was installed on my system, but having looked it's actually a dependency of virtualmin.

In which case I'm not sure yet what the right approach to this is. `nsupdate` could be used, but the same question goes for that. It might work by using webmin APIs to freeze, modify and thaw the zone. I'll have to do some research and experimentation.

jcameron: I'll merge this, and add some code to support the case where those deps are missing.

jcameron: Actually ... I think the freeze/thaw approach is better, because that's what Webmin normally does.

jcameron: Check out this patch : https://github.com/webmin/webmin/commit/94c64c2308328118de5a1cff45165413f82edab3

dsbibby: Amazing, just tested and that works by the look of it! Thank you! 👍

Neustradamus:
@dsbibby, @jcameron: What is the status of this PR?

jcameron:
@Neustradamus this shouldn't be needed in the latest Webmin release

PR #1762, opened 2022-10-19
Fix `systemd` unit lost properties

iliajie:
Howdy, Jamie! 😉 

As we run numerous of `systemctl show unit1 ... unit150` in one go for a better performance, the output of `systemctl show` command **doesn't start** with an _`Id`_, and as a result all lines that related to the unit before the _`Id`_ (i.e. collected in `if (lc($n) eq 'id') { ... }` code) will be lost. This is the reason why `$info` object never gotten set `ExecMainPID`, `ExecStart`, `ExecStop` and `ExecReload` keys, simply because they were all found in a dump of `@lines` before _`Id=`_ line.

I was looking a lot into `systemctl show --property=` command to force it return unit _`Id`_ at the top of the list of all properties but it's sorted alphabetically and seems just impossible.

I was thinking about the solution and came up to the most straight forward one is to manually print _**`Id`**_ based on the passed unit name, which always has to be the same considering that we're listing units using `systemctl` command before that.

I have also dropped `ExecStart`, `ExecStop` and `ExecReload` keys which apparently were never used, but would consume a lot of space in the object.

It also makes the whole code run faster for about 100-200 ms. Not much but not worse, considering we are displaying more units now (not loosing units anymore).

You can open browser's console and run `document.querySelectorAll('[action="mass_systemd.cgi"] [type="checkbox"]').length` to count the number of units before my yesterday's patches and with currently distributed Webmin version. You will see that due to a bug we were loosing a lot of units!

Tested on CentOS 7, Alma 9 and Ubuntu 22.04. All works.

I was going to push this change directly but because there is a lot to explain, I decided to make a PR.

If you have any questions please ask.


jcameron:
Do you mean that the first unit in the list doesn't start with `Id=` , but the rest do?

Or is the order of values in each unit just totally random?

iliajie:
Any unit.

jcameron:
I don't like the use of `echo` here, it seems hacky. Should we instead just run `systemctl show` on units one at a time, and parse each separately?

iliajie:
No, it doesn't matter if you call it one at a time or not. Also it will be significantly slower.

And it just won't work any way as `systemctl show` returns a list of properties with `Id` at the middle of the list. It will always makes us loose required `ExecMainPID` property which comes before `Id` property.

This is not much of a hack. We just manually add `Id` to the list of lines. There is just no other way (faster way) of doing it really.

iliajie:
But it could be done the way you suggest if we turn this list into a hash by calling it individually. But that will be around 2.5x slower. This page is already taking way too long to load.

I wouldn't worry about `echo` as long as `Id` and service name match. It's super fast and does what we want.

jcameron:
I reverted your original change in favor of this method of using the blank lines between units : https://github.com/webmin/webmin/commit/9350bd79cd2725ca5da225f658c89a9ad144e0c2

iliajie:
I don't know what else to try. Unless there is some undocumented param that can be passed to `systemctl show`, I don't know how else to make it work but running `systemctl show` for each unit, which is 25-40% slower unfortunately. I have updated my PR to include your latest changes but the problem remains in currently checked in code.

Have a look -- https://github.com/webmin/webmin/pull/1762/commits/5404142690c8f0050beb284ee868a93aa54ee67b.

jcameron:
I'm not seeing this loss of units on my system. Does it lose the same one all the time? Or is there some upper limit on how many it returns?

iliajie:
> I'm not seeing this loss of units on my system.

That is impossible, as it happens on any contemporary distro I tried - Alma 9, Debian 11 or Ubuntu 22.04.

The only way to make it work correctly with currently checked in code, is you run each command individually, i.e. replacing `100` with `1` in `while(@args < 100 && @units)`. It will be slow but it will help you to see what's wrong. You can count the number of returned units in UI by running in browser's console `document.querySelectorAll('[action="mass_systemd.cgi"] [type="checkbox"]').length` or simply make a dump of `@args` and diff it.

> Does it lose the same one all the time? Or is there some upper limit on how many it returns?

No upper limit. It always returns the same list/number of units with `100` set as a limit. If you set it to `50` or some other number the list changes a bit, although it never produces correct results!

jcameron:
Ok I added some code to diff the units we asked for and the units we got ... and you're right, they are not consistent!! Looking into a fix now..

jcameron:
I noticed that `systemctl show` with the `--property` flag causes some units to be skipped if they don't have the properties. But this patch helps : https://github.com/webmin/webmin/commit/532e1150ced28d338b548347e62a73505c42d8b7

iliajie:
> I noticed that `systemctl show` with the `--property` flag causes some units to be skipped if they don't have the properties. But this patch helps : [532e115](https://github.com/webmin/webmin/commit/532e1150ced28d338b548347e62a73505c42d8b7)

`systemctl show` uses `--all` flag for this. Although, this is not the source of the problem, nor the solution.

As the proof of concept that it still doesn't work correctly, try the following: 

  1. Apply currently checked in code and refresh `init/index.cgi` page and then count number of returned units
  2. Take your code and change `while(@args < 100 && @units) {` to `while(@args < 1 && @units) {` and then count the number of returned units

Is the returned number of units the same?

(to count units in UI I used `document.querySelectorAll('[action="mass_systemd.cgi"] [type="checkbox"]').length` in browser's console).

iliajie:
I started to look deeper for a reason why it may happen, that the output is like this, and at first I assumed that it's a glitch of `systemctl show` command (and therefor made initial patch) but now found the answer - `systemctl show` simply fails at some units, and all followed units are skipped, i.e.:

```
Failed to get properties: Unit name blockdev@.target is missing the instance name. 
```

iliajie:
Also, I have factored out the units types to a separate sub.

init/init-lib.pl
```diff
@@ -2178,7 +2181,7 @@ sub list_systemd_services
 	while(@args < 100 && @units) {
 		push(@args, shift(@units));
 		}
-	my $out = &backquote_command("systemctl show ".join(" ", @args)." 2>/dev/null");
+	my $out = &backquote_command("systemctl show --property=Id,Description,UnitFileState,ActiveState,SubState,ExecStart,ExecStop,ExecReload,ExecMainPID,FragmentPath ".join(" ", @args)." 2>/dev/null");
```
jcameron, line 2184: Are you sure that it's safe to request specific properties if units without them would be skipped?

iliajie: Yes, it's safe, as we need units to have the `Id` field at least. Also, there are no such units out of 300+ on Ubuntu which would be skipped with `--properties` code vs show all. And it should be much faster with `--properties`, as otherwise there is a lot of lines to iterate.

jcameron: Have you tested to see what happens if you request a property that some units don't have?

iliajie: Yes, nothing happens:

```
root@ubuntu22-pro:~# systemctl show --property=Id,Description,UnknownProp webmin.service
Id=webmin.service
Description=Webmin server daemon
```

.. only known properties will be returned, and others are ignored.

jcameron: Ok! Do I still need to merge this PR, or have all the fixes been submitted?

iliajie: Yes, you need to merge it. 😅

PR #1763, opened 2022-10-21
Some fixes in polish translation

7Adrian:


iliajie:
Thanks for the patch!

PR #1765, opened 2022-10-24
Interactive shell improvements

iliajie:
Jamie, hello.

I have started integrating the new interactive shell to Authentic and found few was to improve it first to the older themes:

   * Update to the latest Xterm.js 5.0.0
   * Drop .map files (non needed and saves few extra MB of disk space)
   * Add new Xterm.js module to auto-fit container size, to full screen:
     image

   * Drop rows and columns config options as terminal will fit the container automatically the best way possible


jcameron:
We need to keep the rows / columns options, or at lease columns - many users expect the terminal to be 80 characters wide.

iliajie:
Why would users expect that?

Or you mean at minimum?

jcameron:
I mean we should still have an option to have a fixed-width terminal.

iliajie:
Okay, I will make it optional then.

Can you explain why would any user wanted it to be fixed width, rather than stretched to the whole screen?

iliajie:
> I mean we should still have an option to have a fixed-width terminal.

Oh, I see now. Then we should find a way to pass it to `COLUMNS` env var.

iliajie:
Okay, I have added back a config option to set width and height of the terminal but also added an auto option.

Now let me see how we can set `COLUMNS` env reliably, depending on the screen size.

iliajie:
> I mean we should still have an option to have a fixed-width terminal.

Added ability to make it automatic as well, i.e. depending on the screen size.

iliajie:
Also fixed UTF-8 support. Can be validated by running `systemctl status webmin` with the code before and after the patch.

iliajie:
Also fixed rows (lines) to be browser dependent in automatic mode. Can be validated with opening `nano /etc/passwd` and controls now are at the bottom of the screen.

jcameron:
Can you split one of these up so I can merge them separately? It will make my development on the terminal module easier, because I can merge the obvious changes faster.

iliajie:
Yes, but which these commits you're referring to? Also, they are all not really touching any of your initial logic. And I will be finished in few minutes with this one (add support for spa themes).

iliajie:
Alright, I have finished. All commits were tested pretty well in both themes. I would like to kindly ask you to check/test all the commits and merge them before making further changes, so there would be no conflicts. Changes are cosmetics mostly.

Tomorrow I will integrate it to Authentic Theme to work in a standard mode (with left menu) and add it as a drop-down terminal too (if module `xterm` is available, and if not we will fall back to `shell` module).

Also, let's rename it to `Terminal` instead?

xterm/index.cgi
```diff
@@ -7,13 +7,72 @@
...
+	return;
+}
+
+# Set pixel to columns conversion
+my $cols_num_user = int($screen_width / (int($in{'f'}) || 9));
+
+# Set pixel to rows (lines) conversion
+my $rows_num_user = int($screen_height / (int($in{'l'}) || 18));
+
+# Adjust columns and rows offset
```
jcameron, line 46: Why this adjustment?

xterm/index.cgi
```diff
@@ -7,13 +7,72 @@
...
 require './xterm-lib.pl';
+&ReadParse();
 
+# Get Webmin current version for links serial
+my $wver = &get_webmin_version();
+$wver =~ s/\.//;
+
+# Build Xterm dependency links
+my $termlinks = 
+	{ 'css' => ['xterm.css?$wver'],
```
jcameron, line 18: Why include the version here?

xterm/index.cgi
```diff
@@ -7,13 +7,72 @@
...
+		 "<script src=\"$termlinks->{'js'}[1]\"></script>\n".
+		 "<script src=\"$termlinks->{'js'}[2]\"></script>\n"
 		);
 
+# Set column size depending on the browser window size
+my $screen_width = int($ENV{'HTTP_X_AGENT_WIDTH'}) || int($in{'w'});
+my $screen_height = int($ENV{'HTTP_X_AGENT_HEIGHT'}) || int($in{'h'});
+if (!$screen_width ||
+    !$screen_height) {
+	print "<script>location.href = location.pathname + '?w=' + window.innerWidth + '&h=' + window.innerHeight;</script>";
```
jcameron, line 36: What's the purpose of these w= and h= URL parameters?

xterm/index.cgi
```diff
@@ -52,26 +111,32 @@ $ENV{'SESSION_ID'} = $main::session_id;
...
-print &ui_table_row(undef, "<div id=terminal></div>", 2);
-print &ui_table_end();
-print "</center>\n";
 my $url = "wss://".$ENV{'HTTP_HOST'}.$wspath;
-my $rows = $config{'rows'} || 24;
-my $cols = $config{'cols'} || 80;
+print "<div id=\"terminal\" $termopts->{'ContainerStyle'}></div>\n";
+my $term_script = <<EOF;
+	var term = new Terminal($termopts->{'Options'}),
+		socket = new WebSocket('$url', 'binary'),
```
jcameron, line 118: This indent looks wrong

xterm/index.cgi
```diff
@@ -52,26 +111,32 @@ $ENV{'SESSION_ID'} = $main::session_id;
...
+		attachAddon = new AttachAddon.AttachAddon(socket);
+	term.loadAddon(attachAddon);
+	$termopts->{'FitAddonLoad'}
+	term.open(document.getElementById('terminal'));
+	$termopts->{'FitAddonAdjust'}
+	term.focus();
+EOF
+
+# Return inline script data depending on type
+my $term_script_data =
```
jcameron, line 128: What does this do?

xterm/index.cgi
iliajie: This is the way to determine the size on the user's browser window and adjust the the terminal accordingly.

xterm/index.cgi
iliajie: This is to tell the browser, not to use cached CSS and JS files, which potentially can change with each new Webmin version.

xterm/index.cgi
iliajie: These return browser's current window width and height.

xterm/index.cgi
iliajie: Agreed. Fixed.

xterm/index.cgi
iliajie: Themes that are designed as SPA won't be able to process module load in a standard way, as there is no regular page load process happening. The module instead will return JSON object with instructions from the server.

xterm/index.cgi
iliajie: Oh, right, you're asking about the adjustments. Well, this is because we have a static container for the terminal. Static container would have, let's say 400px width. But calculated width would depend as each tiny block inside the terminal takes 9x18 pixels of WxH by default. This is to accommodate those 9 pixels in width and 18 pixels in height. Cockpit's terminal is also using Xterm.js and it seems very well implemented. I will look into a better (more precise) ways to do it for us as well. It's good already as it is but not yet perfect.

iliajie: Yas, queen! 🥳  I figured, and will submit the change after full rework and final tests.

xterm/index.cgi
swelljoe: Surely perlcritic would complain about it being empty.

xterm/index.cgi
jcameron: Ok, makes sense!

xterm/index.cgi
jcameron: Will this case the page to be reloaded though? And do we really need this, or can we use width=100% ?

iliajie: Yes, it will but only if headers (with width and height) aren't sent, i.e. in old themes.

We need absolute values, not relative, so `%` won't work.

xterm/index.cgi
swelljoe: Oh, I see it continues on the next line. Sorry. I'm just confused.

xterm/index.cgi
jcameron: Why not use the `redirect` sub instead of setting `location.href` ?

iliajie: > Why not use the `redirect` sub instead of setting `location.href` ?

The trick here is to get container size, which depends on the user screen/resolution/scaling. Authentic Theme won't need it, as container size will be sent with headers. Older themes will have to do quick reload (it's un-noticeable though and URL will be cleared afterward to support page resize).

iliajie:
Alright, all is finished. What was improved:

   1. Drop fit addon, as we always pass cols and rows (even in auto mode
   2. Fix terminal fit by columns and lines (rows) (text with `nano /etc/passwd`)
   3. Add support for proper terminal resize on full page reloads
   4. Add CSS only based pre-loader
   5. Refactor and improve code

xterm/index.cgi
jcameron: What's special about this module that the theme needs to get a JSON object?

iliajie: This module preloads `xterm` _js_ and _css_. Authentic theme when run in a SPA mode (default, with the nav menu) doesn't load things from the `head` like this. Also, as was in development it's easier for me like this, as I will have a simpler and more deep control over the module on the theme side.

jcameron: I see, that makes sense

jcameron:
So I noticed that when using the authentic theme, the shell opens in the top-level frame rather than just on the right ... is that expected?

iliajie:
For now, yes. But infact it shouldn't be loading at all, as I haven't integrated it yet to run in SPA mode.

Also, it's expected for any module if you call it directly (i.e. module directory in URL) to load without left menu.

jcameron:
What I mean is, under the older themes it opens in the right frame : 

image

iliajie:
It will be the same in Authentic when I integrate it!

jcameron:
Ok cool! Wasn't sure if this was intentional or not ..

iliajie:
Okay, it Xterm support in SPA mode is fully implemented on Authentic side. It supports connection cleaning, reconnection and many other goodies.

The last step is to integrate it as a drop-down terminal. This is also isn't trivial. I will update this ticket when it's finished.

For now, I suggest applying all latest patches from Webmin and Authentic repos and giving it a try. What should be tried:

  1. Different resolutions and scaling modes wile opening `nano filename`
  2. Restarting Webmin in the background
  3. Multiple tabs

iliajie:
Alright, dropdown terminal in UI was implemented!

Give it a try. Dropdown terminal also supports reconnections.

And it looks indistinguishable from old _Command Shell_ implementation.

PR #1770, opened 2022-10-30
Add an ability for a master admin to switch to specific user

iliajie:


xterm/index.cgi
```diff
@@ -181,6 +181,18 @@ my $user = $access{'user'};
 if ($user eq "*") {
 	$user = $remote_user;
 	}
+
+# Switch to given user
+if (&webmin_user_is_admin()) {
```
jcameron, line 186: Instead just check if `$user eq "root"`

xterm/index.cgi
```diff
@@ -181,6 +181,18 @@ my $user = $access{'user'};
...
+	my $username = $in{'user'};
+	if ($username) {
+	my @uinfo = getpwnam($username);
```
jcameron, line 189: Better to call error if the provided username doesn't exist

xterm/index.cgi
iliajie: No, no .. why? I would prefer to use a specialized sub for it.

jcameron: In this specific case, if the user can access the shell as `root`, then have the power to run as any user. Checking `webmin_user_is_admin` may have a different meaning.

xterm/index.cgi
iliajie: Fixed.

xterm/index.cgi
iliajie: Got it. Fixed.

PR #1771, opened 2022-10-31
Add special flavors to a new Terminal module

iliajie:
This PR fixes/improves:

  1. Fixes sent init commands (such as `clear` and other) via socket won't be stored in `bash` history
  2. Fixes initial command line prompt shown with 150 ms delay to make sure initial commands won't be reflected in UI
  3. Adds (optional) colors to `ls` command
  3. Adds (optional) colors to `bash` prompt depending on the user

For _root_ capable users (magenta user name and blue host):

image

For regular users (green user and blue host):

image

This will make our Terminal look distinctive and improves UX.

### DIFF
--- a/xterm/config
+++ b/xterm/config
@@ -1,1 +1,2 @@
 base_port=555
+flavors=1
--- a/xterm/config
+++ b/xterm/config
@@ -1,2 +1,2 @@
 base_port=555
-flavors=1
+noflavors=0
--- a/xterm/config
+++ b/xterm/config
@@ -1,2 +1,2 @@
 base_port=555
-noflavors=0
+flavors=1
--- a/xterm/config
+++ b/xterm/config
@@ -1,1 +1,2 @@
 base_port=555
+flavors=0
--- a/xterm/config
+++ b/xterm/config
@@ -1,2 +1,1 @@
 base_port=555
-flavors=1
--- a/xterm/index.cgi
+++ b/xterm/index.cgi
@@ -201,7 +201,7 @@ if ($user eq "root" && $in{'user'}) {
 
 # Terminal flavors
 my (@cmds, $term_flavors);
-if (!$config{'noflavors'}) {
+if ($config{'flavors'}) {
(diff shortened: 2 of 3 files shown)


xterm/index.cgi
```diff
@@ -197,6 +233,7 @@ if (!-r $shellserver_cmd) {
 	}
 defined(getpwnam($user)) || &error(&text('index_euser', &html_escape($user)));
 my $tmpdir = &tempname_dir();
+$ENV{'HISTCONTROL'} = 'ignoredups:ignorespace';
```
jcameron, line 236: What does this do?

iliajie: `HISTCONTROL` is evn variable  that controls how commands are saved in the history list.

iliajie: i.e. we send commands though the socket directly, like `'clear'` and `'exit'`, and a few other, so we will use a `space + command`, like `' clear'` or `'  exit'` to make sure it won't be saved to the history list.

xterm/config
```diff
@@ -1 +1,2 @@
 base_port=555
+noflavors=0
```
jcameron, line 2: This should be off by default, at least initially. We don't know if the user's shell even supports aliases or the PS1 var, and also I want to get a basic working version of this module out to users before we add a lot of fancy stuff.

xterm/index.cgi
jcameron: Ah, thanks!

xterm/config
iliajie: `bash` does support all of this. If not, it will be just ignored. But alright, I reverted that one. Let's keep it off by default for now for regular users.

xterm/config
```diff
@@ -1 +1,2 @@
 base_port=555
+flavors=1
```
jcameron, line 2: Shouldn't this be flavors=0

iliajie: If you want to keep it off for root capable users too, then yes? But I thought we could keep it off only for domain owners?

iliajie: Okay, fixed it to be off by default.

PR #1773, opened 2022-11-03
Drop obsolete DNSSEC-Tools

iliajie:


jcameron:
How are they obsolete?

iliajie:
> How are they obsolete?

Because BIND 9.9 (starting 2012) can handle DNSSEC on its own.

PR #1774, opened 2022-11-04
Add automatic option for command prompt color customization

iliajie:


xterm/index.cgi
```diff
@@ -198,10 +198,14 @@ if ($user eq "root" && $in{'user'}) {
 		&error(&text('index_euser', &html_escape($in{'user'})));
 	$user = $in{'user'};
 	}
+defined(getpwnam($user)) || &error(&text('index_euser', &html_escape($user)));
+my @uinfo = getpwnam($user);
```
jcameron, line 202: Can we skip calling getpwnam twice?

iliajie: Yeah, like that - https://github.com/webmin/webmin/pull/1774/commits/19070f33833fe238b3a364e4074ff8b20e08c7f5

PR #1775, opened 2022-11-07
Add first steps to support Terminal resize

iliajie:
I have tried to find escape sequence to resize terminal, and I think I did, although it doesn't work.

This PR executes both JavaScript and Perl codes, just for demonstration purposes. I was hoping to find a method on `$shellfh` from `IO::Pty` but it also didn't work. I am pretty stuck here.


xterm/index.cgi
```diff
@@ -271,11 +273,34 @@ my $term_script = <<EOF;
...
 		term.open(termcont);
+		fitAddon.fit();
 		term.focus();
+		
+		// On resize event triggered by fit()
+		term.onResize(function(e) {
+			// \033[8;(40);(100)t
+			// socket.send('\\\\033\\\\[8;('+e.rows+');('+e.cols+')t');
+			term.write('\\033[8;'+e.rows+';'+e.cols+'t')
+			socket.send(JSON.stringify(e));
```
jcameron, line 288: This seems risky ... what if the some valid JSON gets sent as part of the normal terminal operation?

iliajie: This is not a production ready code. If that's the path to make resizes work, I will make it very specific type of JSON object to work with. Now, it's just for testing purposes.

PR #1776, opened 2022-11-08
Fix to check for resize escape sequence explicitly

iliajie:


xterm/index.cgi
```diff
@@ -282,9 +282,7 @@ my $term_script = <<EOF;
 		
 		// On resize event triggered by fit()
 		term.onResize(function(e) {
-			// \033[8;(40);(100)t
-			//socket.send('\\\\033\\\\[8;('+e.rows+');('+e.cols+')t');
-			socket.send('___RESIZE___ '+e.rows+' '+e.cols);
+			socket.send('\\\\033[8;('+e.rows+');('+e.cols+')t');
```
jcameron, line 285: Why this escape sequence? Is it interpreted by the shell?

iliajie: No, we will interpret on shell server.

iliajie: It's somewhat expected sequence though (used in PR). `___RESIZE___` theoretically could be used for something else. Where as this one is expected to do resize only. I realize that we don't send it to the terminal, and we send it over the socket to the shell. Yet I find it safer, as anyone calling it would only expect to do a resize on the terminal, and nothing else.

xterm/shellserver.pl
```diff
@@ -116,7 +117,13 @@
 				exit(0);
 				}
 			if ($wsconn) {
-				$wsconn->send_utf8($buf);
+				# Do not print commands to display, if start from two
+				# consecutive spaces and ends with return char. This
+				# regex only catches commands sent in one shot using
+				# socket.send('  command\r') call, i.e. it won't work
+				# if command is pasted or typed in a normal way
+				$buf =~ s/^[\s]{2}\S+.*[\r\n]+[\e]\[\?2004l[\n]*//;
```
jcameron, line 125: This seems a bit hacky - I think we should just pass all output back, because this could trigger if something in the shell outputs a line that starts with two spaces.

xterm/index.cgi
jcameron: Ok, makes sense

xterm/shellserver.pl
iliajie: > because this could trigger if something in the shell outputs a line that starts with two spaces.

It should also match a new line and escape char, which is impossible.

Alright, I made it more resilient to impossible output - https://github.com/webmin/webmin/pull/1776/commits/163480b8591cdc4ccabccef4ff112d778afd1ced

iliajie: We could technically remove it, it's not critical but it would be nice to have a way to suppress output for the commands sent directly.

iliajie: Alright, I will drop it for now to move on. We can discuss it later.

As everything else is alright, I will merge it then.

jcameron: Thanks!

PR #1782, opened 2022-11-10
Add support to source user profile if any

iliajie:


jcameron:
This is the wrong solution - instead we should just run the shell as a login shell so that .bash_profile is read, like so : https://github.com/webmin/webmin/commit/ad70a474d910a4509e17a8786ed2923c5e711d50

iliajie:
> This is the wrong solution

Apologies! Agreed, and I mentioned this too in our yesterday's [discussion](https://github.com/webmin/webmin/discussions/1781).

> instead we should just run the shell as a login shell so that .bash_profile is read, like so :

Yes, that makes it a login shell now. Great, thanks!

PR #1784, opened 2022-11-11
Add support to respect currently user defined `PS1`

iliajie:


jcameron:
Why do it by sending a command to the shell, instead of just setting `$ENV{'PS1'}` before launching the shell process?

iliajie:
> Why do it by sending a command to the shell, instead of just setting `$ENV{'PS1'}` before launching the shell process?

Because those would be overwritten by anything set upon login. Also, we cannot just set it. We need to know if user uses known default `PS1` or uses a custom `PS1`. If a user uses custom, we should respect user preferences and not touch `PS1` at all, unless a user forces it in options (i.e. using `Yes`, and not default `Automatic`).

iliajie:
> Why not instead make use of the `--init-file` flag to `bash` , which lets you define commands to be run before the interactive shell starts?

That would work too, if we knew whether a user does or does not customized his own `PS1`. As I understand this, it's only possible to do after we have the shell. Remember we cannot set our `PS1` until we learn about what user has defined.

The other solution (probably less reliable) would be to read `~/.bashrc`, `~/.bash_profile` and `.profile` using `read_file_lines` and see if any `PS1` was set in return. Although, this approach would skip any custom sourced user files within mentioned files. However, we could parse and later read any embedded files starting from `. filename` or `source filename` .. as I think of this now, it's very doable too.

> This whole approach seems unreliable to me, as it depends on knowing what the OS-level default settings for PS1 are, which might change between distros.

It seems quite a common practice for us to rely on OS level defaults. Besides if we fail to test it, it won't impose our `PS1` rather on opposite, it will do nothing. Also, those defaults haven't changed for decades.

jcameron:
The `--init-file` flag specifies a script to run instead of `.bashrc` . So we could in theory set this to a temporary script like : 

```
source $HOME/.bashrc
if [ "$PS1" = "" ]; then
  PS1=....
fi
```

iliajie:
Yes, that perhaps would work, _**but**_ unfortunately this won't solve our problem, as shells set `PS1` to default ugly one, like `[\u@\h \W]\$`. This is why we need to either test current shell with `echo $PS1` or as I'm thinking of it now, it would be better to read all user bash shell initialization files, and check for `PS1`.

I have much more confidence in doing that rather than using `&wait_for()` on an interactive context. This is the only part of my PR which I wasn't very confident of, as unfortunately `eof()` doesn't work with interactive content ..

Do you know the way to reliably and _simply_ read `$shellfh` into a string?

Also, let me make another PR which we could discuss to and pick best with reading `bash` shell initialization files.

jcameron:
Reading all the shell initialization files is risky, because we don't know if PS1 is set inside an `if` statement or via some complex substitution..

iliajie:
> Reading all the shell initialization files is risky, because we don't know if PS1 is set inside an `if` statement or via some complex substitution..

Not really risky at all. Check [this PR](https://github.com/webmin/webmin/pull/1785). I will close this for now.

PR #1785, opened 2022-11-11
Further Terminal improvements (done right) 

iliajie:
Jamie, please have a look. This is really the enhanced version. 

  1. It abstracts the code to make it possible to add more flavors to other shells if needed
  2. It checks on known `bash` shell init files (including possible user sourced files) and checks for `PS1`, and respects it if set
  3. It adds support to define on the module config custom environmental vars, like:
  ```
  var1=val1
  var2=val2
  ```
  4. It adds support to define on module config commands to run after shell login, like:
   ```
   alias ls='ls -sa'
   alias lsetc='ls -lsa /etc'
   ```

### DIFF
--- a/xterm/config
+++ b/xterm/config
@@ -1,2 +1,2 @@
 base_port=555
-flavors=2
+flavors=1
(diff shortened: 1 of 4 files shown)


iliajie:
Suppressed by https://github.com/webmin/webmin/pull/1786.

PR #1788, opened 2022-11-13
Fix to take out general `websocket` subs to API

iliajie:


jcameron:
There's a lot going on in this PR ... why move these methods up to the general API? They are specific to the xterm module.

iliajie:
> There's a lot going on in this PR

Doesn't change the logic.

>  why move these methods up to the general API? They are specific to the xterm module.

They are websockets specific and any other module using websockets would do be able to re-use them.

jcameron:
OK, if and when other modules need this functionality, we can move it. We don't even know yet what kinds of paths they'll use. Also, I prefer not to add to the global API until it's needed, as it increases the startup time of other modules.

iliajie:
Alright, fare point. Let me see with that another PR for the port..

PR #1789, opened 2022-11-13
Add minor clean ups

iliajie:
* [Fix to place cleanup miniserv to the lib file](https://github.com/webmin/webmin/commit/e6795f28f9e75d241aa503bea83ba72c080d34c2)
 * [Fix to save miniserv websocket after we have the pid from PTY](https://github.com/webmin/webmin/commit/4907aeb2ddfa0e795407d15815474bca690feede)
 * [Fix to clean up others sockets if the shell server](https://github.com/webmin/webmin/commit/b62fb504ac17413af010b515dfad956fc92bc312)
 * [Fix to adjust functions names](https://github.com/webmin/webmin/commit/dbb09b527849bed91993d6c241a3b9bcaced706b)

### DIFF
--- a/xterm/shellserver.pl
+++ b/xterm/shellserver.pl
@@ -75,10 +75,6 @@ else {
 	print STDERR "Running shell $shcmd with pid $pid\n";
 	}
 
-# Save new websocket info to miniserv.conf
-# only after we get the pty pid
-&save_miniserv_websocket($port);
-
 # Detach from controlling terminal
 if (fork()) {
 	exit(0);
@@ -86,6 +82,7 @@ if (fork()) {
 untie(*STDIN);
 close(STDIN);
 
+# Clean up when socket is terminated
 $SIG{'ALRM'} = sub {
 	&remove_miniserv_websocket($port);
 	die "timeout waiting for connection";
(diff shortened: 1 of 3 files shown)


xterm/shellserver.pl
```diff
@@ -68,13 +68,17 @@
...
+	&remove_miniserv_websocket($port);
 	die "Failed to run shell with $shcmd\n";
 	}
 else {
 	print STDERR "Running shell $shcmd with pid $pid\n";
 	}
 
+# Save new websocket info to miniserv.conf
+# only after we get the pty pid
+&save_miniserv_websocket($port);
```
jcameron, line 80: We can't move this down, because the port allocation and updating miniserv.conf has to be done while the file is locked, to prevent a collision between multiple shells launched at the same time. That's why the current code has it all in one block.

xterm/xterm-lib.pl
```diff
@@ -5,10 +5,47 @@
...
 our %access = &get_module_acl();
 
-# cleanup_old_websockets([&skip-ports])
-# Called by scheduled status collection to remove any websockets in
-# miniserv.conf that are no longer used
-sub cleanup_old_websockets
+# save_miniserv_websocket(port)
+# Save new websocket info
+# to miniserv.conf file
+sub save_miniserv_websocket {
```
jcameron, line 11: Please fix the { at the end of the line

xterm/shellserver.pl
iliajie: Alright, fixed.

PR #1790, opened 2022-11-14
Fix connecting to external IPv6 LDAP server

krouma:
If the provided hostname couldn't be translated to IPv4 address or wasn't IPv4 address, Webmin threw error:
"The LDAP server $1 does not exist".


iliajie:
Thanks for the patch!

PR #1792, opened 2022-11-21
Add API to get buffer size for file transfers

iliajie:


filemin/download.cgi
```diff
@@ -25,7 +25,7 @@ print "Content-Disposition: attachment; filename=\"$name$ext\"\n";
 print "Content-Length: $size\n\n";
 open (FILE, "< $file") or die "can't open $file: $!";
 binmode FILE;
-local $/ = \2048000;
+local $/ = \&get_buffer_size_binary();
```
jcameron, line 28: Why the \ here?

jcameron: Why does it need to be a number reference and not just a number?

iliajie: It was mentioned somewhere on the docs:

<code>
Setting $/ to a reference to an integer, scalar containing an integer, or scalar that's convertible to an integer will attempt to read records instead of lines, with the maximum record size being the referenced integer.
</code>
<br><br>
I am not sure why it should be a reference to an integer. Do you have a clue why?

jcameron: It also says "scalar containing an integer" , so there's no need for the \

jcameron: I don't get why it needs to be a **reference** to an integer. Can you test this?

iliajie: I actually did test it!

jcameron: Also, doesn't \& create a reference to the function, not the integer?

iliajie: No, as we use `&func()` rather than `&func` .. and that (our case) will get the function return.

We need to set it to bytes, so we got to use `\`.

PR #1796, opened 2022-11-25
Fix to factor out Webmin Websockets subs

iliajie:


jcameron:
Why do this? Is there another module we'd be calling these from?

iliajie:
Yeah, I'm think of the theme using a socket connection to pull the stats.

jcameron:
Ok, makes sense..

PR #1805, opened 2022-12-05
Add support to set TERM variable and shell character encoding

iliajie:
This PR adds ability to set `TERM` variable and shell character encoding from module config:

image


xterm/config.info
```diff
@@ -1,3 +1,5 @@
+xterm=Set <tt>TERM</tt> environmental variable to,4,0-xterm&#45;256color,1-xterm&#45;16color,2-xterm,3-vt102,3-vt100,4-vt52,5-rxvt,6-nsterm,7-dtterm,8-ansi
```
jcameron, line 1: Why not just make the values the actual terminal code instead of numbers?

iliajie: I wanted to have an explicit check on what is going to `TERM` variable, rather than just passing it directly.

jcameron: Thanks for fixing this!

PR #1806, opened 2022-12-05
Update ru

Divarion-D:


iliajie:
Thanks, Danil.

Although, you should also remove human translated strings from `webmin/lang/ru.auto` file.

iliajie:
Great, thanks!

PR #1808, opened 2022-12-14
Add support to test if parent session exists

iliajie:


xterm/index.cgi
```diff
@@ -179,7 +179,10 @@ if (!-r $shellserver_cmd) {
 	}
 my $tmpdir = &tempname_dir();
 $ENV{'SESSION_ID'} = $main::session_id;
-&system_logged($shellserver_cmd." ".quotemeta($port)." ".quotemeta($user).
+&foreign_require('acl');
+my $parentsessionid = &acl::hash_session_id($main::session_id);
```
jcameron, line 183: Passing in the session ID as a command-line parameter is a bad idea, because it will be visible to anyone on the system running `ps` and could be used to hijack another user's session!  Also, it's available via the `$main::session_id` variable anyway.

xterm/websockets-lib-funcs.pl
```diff
@@ -97,4 +97,17 @@ sub cleanup_miniserv_websockets
 &unlock_file(&get_miniserv_config_file());
 }
 
+sub check_miniserv_parent_session_websocket
+{
+my ($sid) = @_;
+&foreign_require('acl');
+my %miniserv;
+&get_miniserv_config(\%miniserv);
+&acl::open_session_db(\%miniserv);
```
jcameron, line 106: You'll need to close the session DB, or else I don't think updates done by `miniserv` are guaranteed to be visible..

xterm/index.cgi
iliajie: I wasn't passing session id as is, as the session id was hashed.

xterm/websockets-lib-funcs.pl
iliajie: Got it. Thanks.

jcameron:
Actually we can drop this PR, because I fixed it in miniserv

Issue #1833, opened 2023-02-05
Bacula Restore not showing subfolders


thjansen72:
When I try to restore via Bacula restore, I can't select subfolders.
![image](https://user-images.githubusercontent.com/20270638/216818060-f06bf5ff-392f-4db8-920c-432018ceaeaa.png)

Restore the complete folder is working.

I use the actual Webim on Linux server 5.15.84-v8+ #1613 SMP PREEMPT aarch64 GNU/Linux.

jcameron:
@ctrlbru you re-wrote this file chooser in https://github.com/webmin/webmin/commit/7259d0199d21d844009f12692b3a4e73cf71524e  .... any ideas?

ctrlbru:
I can replicate the issue, I'll check it

ctrlbru:
The reason is 8de2847fe527edd3445dc6a0bf078b4a3d03ab0c, where a reference to JSON:PP was moved to bacula-bacup-lib.pl, but the require in list.cgi is not in a BEGIN block. I verified that with a BEGIN block it works, but I'm no perl expert so I can't evaluate other side effects. Let me know if you need me to prepare a pull request for this.

The fix is this line in list.cgi:

`BEGIN { require './bacula-backup-lib.pl'; }`

jcameron:
Sure, a PR for this would be great!

thjansen72:
Working fine! Thank you!

PR #1835, opened 2023-02-06
Fix for file chooser not displayed correctly in bacula module

ctrlbru:
This PR fixes the file chooser not displaying correctly in the bacula module when restoring.

List.cgi needs JSON::PP which is now included via bacula-backup-lib.pl. Without a BEGIN block around the require, JSON::PP was not visible (soon enough?).

Closes: #1833


jcameron:
Thanks for the fix!

thjansen72:
Working fine! Thank you!

PR #1836, opened 2023-02-07
fix displaying perviously selected users

pzindyaev:
Looks like a typo, but it was preventing displaying previously selected users if you try to edit valid users list from Samba module
Groups selector doesn't have that typo


jcameron:
Thanks for the patch!

PR #1844, opened 2023-02-21
Fix setup-repos.sh perm issues

ItzMiracleOwO:
Lets check for root perms before executing to prevent issues.

Preventing issue like this
image


iliajie:
Thanks for the patch!

PR #1846, opened 2023-02-24
Add set ACLs support

iliajie:
Jamie, this is the first part yet. I will ping you when it's all ready.


jcameron:
Ok let me know when this is ready to review..

iliajie:
It is ready.

filemin/setfacl.cgi
```diff
@@ -0,0 +1,55 @@
...
+    }
+my $cmd = &has_command('setfacl');
+error($text{'acls_error'}) if (!$cmd);
+my $types;
+$types = join(',',@types) if (@types);
+$types .= " $extra" if ($extra);
+my $args = "$action $types $recursive";
+$args =~ s/\s+/ /g;
+$args = &trim($args);
+$args =~ s/[\`\$\;\/\'\"\?\%\&\#\*\(\)\+]//g;
```
jcameron, line 43: Why not use quotemeta instead of this?

filemin/filemin-lib.pl
```diff
@@ -173,6 +173,26 @@ sub print_interface {
     %allowed_for_edit = map { $_ => 1} @allowed_for_edit;
     my %tinfo = &get_theme_info($current_theme);
 
+    # User and group lists for acls
+    if (&has_command('setfacl')) {
+        &foreign_require('useradmin');
+        my @allulist = &useradmin::list_users();
+        my @ulist = &useradmin::list_allowed_users(\%access, \@allulist);
+        @ulist = sort map { $_->{'user'} } @ulist;
```
jcameron, line 181: Is this getting the list of all the users that the current user has permissions to edit? That could be different from the set of users ACLs can be set on. Also, how will this work in Usermin?

filemin/setfacl.cgi
iliajie: Well, it's just doesn't work with `quotemeta`. The command just fails.

filemin/filemin-lib.pl
iliajie: > Is this getting the list of all the users that the current user has permissions to edit That could be different from the set of users ACLs can be set on.

Right, I think we should get all users and groups then instead. I will fix that.

> Also, how will this work in Usermin?

Duh. I forgot that it won't work in Usermin. What is the best way to extract the list of users and groups?

jcameron: Best to just keep it simple and allow selection of any user. There's a `ui_user_textbox` function in ui-lib.pl for this.

iliajie: Well, the best would be is to have a list of users and groups returned in an array. We cannot use `ui_user_textbox` because it generates a modal in Authentic, but choose user functionality if we use it as suggested, would be called from another modal .. It won't look good ..

iliajie: Please factor out `get_users_list`.

filemin/setfacl.cgi
jcameron: Ok but what's the purpose of this regular expression?

filemin/filemin-lib.pl
jcameron: It's not a good idea to create <select> list for choosing a user though, as some systems might have 10s of thousands of users..

filemin/setfacl.cgi
iliajie: To make sure that line injection won't be possible.

filemin/filemin-lib.pl
iliajie: It's not a problem for a select. This is why we want a select because it can have a filter. Also, even with old and standard selects when it's opened, typing will do a fast-search!

jcameron: But the HTML for the select might be huge! That's why the ui_user_textbox function uses a popup..

iliajie: While this is actually true I only noticed a slow down with a select containing 100k entries.

Even if we assume that a system has 10k users + 10k groups (which sounds very unrealistic), in this case we could test it, and if a system has lets say 2k+ users, then we just return a simple input field.

In case of 10K users, the mentioned earlier `ui_user_textbox` would most probably be laggy and slow too, and much slower than a select element. But the advantage that it's not loaded for each page load .. even though, I agree theoretically it is a better solution but practically, for the vast majority of users it won't be noticeable.

Also, speaking theoretically we should not print all modals on the main File Manager page load and pull it in the background. This would solve the problem too. But that would require a large re-work.

Meanwhile, I suggest we factor out users and groups list functions and test if the list is larger than _N_ entries we just return a text box for users and/or groups.

iliajie: I have added a test for large amount of users/groups. The only thing that we should do (outside of these changes) for this PR to work is to factor out `get_users_list` and `get_groups_list` subs.

jcameron: I still think we should use the `ui_user_textbox` function in this case, and if need be change it to a different kind of selector. Or maybe make it behave differently when used modally like this?

iliajie: I could do it most probably. But this is raising complexity.

What is the percentage of users that would have 500 users, 1000 users and 10000 respectively?

jcameron: There are definitely Webmin users with 10,000 or even 100,000 users, often stored in LDAP. Which by the way is why you can't rely on the useradmin module to find the list of users on the system, because it's quite possible they are stored externally and not in `/etc/passwd`

iliajie: Well, alright. Then it has to be done the way it would work in all case scenarios.

Okay .. but could you probably make a simple and available for both Webmin and Usermin functions for getting a list of users and groups without HTML? Perhaps, I could come up with a better solution rather than using our old popup?

jcameron: The Perl API has functions for this already .. you can get all users with code like : 

```
setpwent();
while(my $u = getpwent()) {
  push(@users, $u);
  }
endpwent();
```

iliajie: You're right! I just realized that we cannot really even _try_ to read a list of users and groups on the regular page load. It may be very slow if there are that many users ..

I have added a full support on Authentic theme side to work with user/group popup select, with multiple modals being opened at once. It will work flawlessly!

jcameron: Excellent!

filemin/setfacl.cgi
jcameron: So why not use the standard Perl `quotemeta` function? It will do all the escaping needed in this situation ...

iliajie: > It will do all the escaping needed in this situation

Yeah, fixed [here](https://github.com/webmin/webmin/pull/1846/commits/31eb2180944af65acda9313e830d03444e1e01ae). However, you need to avoid running `quotemeta` on spaces in the params' line, otherwise (for some reason) it fails.

jcameron: You have to call quotemeta on each param separately

PR #1849, opened 2023-03-02
Fix to properly display used space (consider reserved)

iliajie:
https://forum.virtualmin.com/t/disk-usage-accuracy/119276/9?u=ilia

### DIFF
--- a/mount/linux-lib.pl
+++ b/mount/linux-lib.pl
@@ -905,8 +905,8 @@ if ($_[0] eq "proc" || $_[0] eq "swap" ||
 &clean_language();
 local $out = &backquote_command("df -k ".quotemeta($_[1]), 1);
 &reset_environment();
-if ($out =~ /Mounted on\n\S+\s+(\S+)\s+\S+\s+(\S+)/) {
-	return ($1, $2);
+if ($out =~ /Mounted on\n\S+\s+(?<total>\S+)\s+(?<used>\S+)\s+(?<free>\S+)\s+(?<percent>\d+)/) {
+	return ("$+{total}", "$+{free}", "$+{used}", "$+{percent}");
 	}
 return ( );
 }
@@ -919,8 +919,8 @@ if (&get_mounted($_[1], "*") < 0) { return (); }
 &clean_language();
 local $out = &backquote_command("df -i $_[1]", 1);
 &reset_environment();
-if ($out =~ /Mounted on\n\S+\s+(\S+)\s+\S+\s+(\S+)/) {
-	return ($1, $2);
+if ($out =~ /Mounted on\n\S+\s+(?<total>\S+)\s+(?<used>\S+)\s+(?<free>\S+)\s+(?<percent>\d+)/) {
+	return ("$+{total}", "$+{free}", "$+{used}", "$+{percent}");
 	}
 return ( );
 }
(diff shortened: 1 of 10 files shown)


mount/linux-lib.pl
```diff
@@ -905,8 +905,8 @@ sub disk_space
 &clean_language();
 local $out = &backquote_command("df -k ".quotemeta($_[1]), 1);
 &reset_environment();
-if ($out =~ /Mounted on\n\S+\s+(\S+)\s+\S+\s+(\S+)/) {
-	return ($1, $2);
+if ($out =~ /Mounted on\n\S+\s+(?<total>\S+)\s+(?<used>\S+)\s+(?<free>\S+)\s+(?<percent>\d+)/) {
+	return ("$+{total}", "$+{free}", "$+{used}", "$+{percent}");
```
jcameron, line 909: Can we not just compute the percent rather than fetching it from the `df` command?

iliajie: We are only fetching it from `df`. And compute if unavailable.

jcameron: But why not make this regexp simpler and just always compute it?

iliajie: Oh, I see. Well, we cannot change the returned params order! This is made not to always compute percent or not. This is made to make sure that we always first return `total` and `free`, and a new additional params `used` and `percent` .. this is what caller expects. There are other non Linux libs and they only (for now) return old type of data (i.e. `total` and `free`).

In other words, we need to preserve the return order of the params but it's impossible to do using regex in any other way because you were initially skipping on `used` with this old regex `\S+\s+(\S+)\s+\S+\s+(\S+)/`..

jcameron: Oh I see ... in that case, let's keep it the way it is then. I assume that all Linux versions use the same format for the `df` command.

iliajie: Alright, fixed on the last commit.

mount/osf1-lib.pl
```diff
@@ -262,9 +262,9 @@ sub disk_space
 if ($_[0] eq "fd" || $_[0] eq "proc" || $_[0] eq "swap" || $_[0] eq "autofs") {
 	return ();
 	}
-if (&backquote_command("df -k ".quotemeta($_[1]), 1) =~
-    /Mounted on\n\S+\s+(\S+)\s+\S+\s+(\S+)/) {
-	return ($1, $2);
+my $out = &backquote_command("df -k ".quotemeta($_[1]), 1);
```
jcameron, line 265: Are you sure this OS has the same `df` output as Linux?

mount/netbsd-lib.pl
```diff
@@ -231,9 +231,9 @@ sub disk_space
 {
 if (&get_mounted($_[1], "*") < 0) { return (); }
 if ($_[0] eq "proc" || $_[0] eq "swap") { return (); }
-if (&backquote_command("df -k ".quotemeta($_[1]), 1) =~
-    /Mounted on\n\S+\s+(\S+)\s+\S+\s+(\S+)/) {
-	return ($1, $2);
+my $out = &backquote_command("df -k ".quotemeta($_[1]), 1);
```
jcameron, line 234: Are you sure this OS has the same `df` output as Linux?

mount/osf1-lib.pl
iliajie: Just checked. The output is not the same, as it also includes `inodes`.  However, this regex will work as well.

mount/netbsd-lib.pl
iliajie: > Are you sure this OS has the same df output as Linux?

Yes, this regex also works on `df` output in NetBSD!

PR #1850, opened 2023-03-02
Add support for loading given modules info in background only on initial Dashboard load

iliajie:
This patch will allow Authentic theme not to process any possible available packages updates on the Dashboard load making it much faster. It will only use currently collected data from `system-status` module on initial Dashboard load. It will especially be noticeable if `package-updates` cache is not present.

Also, I believe there is some other bug as well, probably in `system-status` module. So, after background collection was run, the Dashboard load jumps 2x of time.

That is unrelated to this PR, I think but could you also double check why after background collection was run the Dashboard load doubles in time? And by the way -- restarting Webmin fixes it!

### DIFF
--- a/system-status/system-status-lib.pl
+++ b/system-status/system-status-lib.pl
@@ -69,7 +69,7 @@ if (&foreign_installed("package-updates") && $config{'collect_pkgs'}) {
 	my $poss_current = !$poss_collect_blocked ? 2 : undef;
 	my @poss = &package_updates::list_possible_updates(undef, $poss_collect_blocked);
 	$info->{'poss'} = \@poss;
-	$info->{'reboot'} = &package_updates::check_reboot_required();
+	$info->{'reboot'} = &package_updates::check_reboot_required($poss_collect_blocked);
 	}
 
 # CPU and drive temps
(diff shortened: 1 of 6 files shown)


system-status/system-status-lib.pl
```diff
@@ -64,7 +65,9 @@ sub collect_system_info
 # Available package updates
 if (&foreign_installed("package-updates") && $config{'collect_pkgs'}) {
 	&foreign_require("package-updates");
-	my @poss = &package_updates::list_possible_updates(2, 1);
+	my $poss_collect_blocked = grep(/^\Qpackage-updates\E$/, @{$modskip});
```
jcameron, line 68: Better to use `&indexof` here instead of `grep`

web-lib-funcs.pl
```diff
@@ -12179,20 +12179,22 @@ =head2 list_combined_system_info(&data, &in)
...
 	&foreign_require($m->{'dir'}, "system_info.pl");
 	foreach my $i (&foreign_call($m->{'dir'}, "list_system_info",
-				     $data, $in)) {
+				     $data, $in, $modskip)) {
 		$i->{'module'} = $m->{'dir'};
 		push(@rv, $i);
 		}
 	}
-if (&foreign_available("webmin")) {
+if (!grep(/^webmin$/, @{$modskip}) && &foreign_available("webmin")) {
```
jcameron, line 12197: Also use `&indexof` here

iliajie: Alright, both are done.

PR #1853, opened 2023-03-09
Can I install litespeed server on webmin?

TiydoSoft:
Hi,
Is there any way to install litespeed server on webmin?

Thank you!


iliajie:
You can install LiteSpeed web server and configure it manually or using File Manager by editing config files directly. There is no Webmin module do it. At least not yet.

PR #1868, opened 2023-03-21
Fix wrong condition when asking first

bisbastuner:
The second test was wrong, it exits if the user asks with a capital Y. Fix by exiting only if the answer from the user is different from a lowercase y AND a capital Y


iliajie:
Thanks for the patch.

PR #1887, opened 2023-04-16
Add support for Terminal to work in Usermin

iliajie:
Connected https://github.com/webmin/usermin/issues/98

### DIFF
--- a/xterm/index.cgi
+++ b/xterm/index.cgi
@@ -23,7 +23,12 @@ foreach my $modname (@modnames) {
 				&text('index_suse', "<tt>$modname</tt>",
 					'https://software.opensuse.org/download/package?package=perl-IO-Tty&project=devel%3Alanguages%3Aperl');
 			}
-		print $missinglink ."<p>\n";
+		if (&get_product_name() eq 'usermin') {
+			print &text('index_missing', $modname) ."<p>\n";
+			}
+		else {
+			print $missinglink ."<p>\n";
+			}
 		&ui_print_footer("/", $text{'index'});
 		exit;
 		}
@@ -179,6 +184,8 @@ my @uinfo = getpwnam($user);
 my $dir = $in{'dir'};
 
 # Launch the shell server on the allocated port
+setvar('cron_no_switch_to_remote_user', 1)
+	if (&get_product_name() eq 'usermin');
 &foreign_require("cron");
 my $shellserver_cmd = "$module_config_directory/shellserver.pl";
 if (!-r $shellserver_cmd) {
--- a/cron/cron-lib.pl
+++ b/cron/cron-lib.pl
@@ -19,7 +19,8 @@ use WebminCore;
 $env_support = $config{'vixie_cron'};
 if ($module_info{'usermin'}) {
 	$single_user = $remote_user;
-	&switch_to_remote_user();
+	&switch_to_remote_user()
+		if (!getvar('cron_no_switch_to_remote_user'));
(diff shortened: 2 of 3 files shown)


xterm/index.cgi
```diff
@@ -179,6 +184,8 @@ my @uinfo = getpwnam($user);
 my $dir = $in{'dir'};
 
 # Launch the shell server on the allocated port
+setvar('cron_no_switch_to_remote_user', 1)
```
jcameron, line 187: Why is this necessary? Seems like we can instead create the `shellserver.pl` script in the user's home directory if running in Usermin..

iliajie: We have to leave the main process running as `root`, because we need to initiate a socket connection, make changes to `miniserv.conf` and write other files, which a regular user not meant to do. However, the shell process will be switched to run as user.

I have checked our code multiple times, and I don't see any ways for privilege escalation. Do you?

jcameron: In that case, maybe we can avoid using the cron module at all, so we don't need to set this global variable which feels a bit hacky?

iliajie: > In that case, maybe we can avoid using the cron module at all, so we don't need to set this global variable which feels a bit hacky?

Yeah, it is possible. Check this patch https://github.com/webmin/webmin/pull/1887/commits/03d245384d06cfd2f2d6c85a4904c926cf448eb0.

swelljoe: Why would a user terminal ever need to (or be allowed to) change `miniserv.conf`? That sounds wildly dangerous.

iliajie: > think actually this is the better patch : https://github.com/webmin/webmin/commit/db8d712d83f26eb6252aca6723b6cb97073fdea8
It's like your original code, but solves the problem more generally..

Jamie, it seems risky, because we may not switch to user process when we expect to do so, and when `$ENV{'FOREIGN_MODULE_NAME'}` is not set. It could potentially introduce a security issue.

I think the simpler and safer solution would be not to `foreign_require` Cron module at all, and just put `create_wrapper` to `websockets-lib-funcs.pl` file (the way I did this in the latest commits).

> Why would a user terminal ever need to (or be allowed to) change `miniserv.conf`? That sounds wildly dangerous.

I think Jamie designed it that way because `miniserv` code always has access to `miniserv.conf` file (for convenience).

> While we're at it, does that mean a user terminal in Webmin can change miniserv.conf? That also sounds scary as hell.

User cannot change `miniserv.conf` voluntarily in any way, parent `root` process does it. What's the risk?

jcameron: The `xterm` module need to change `miniserv.conf` in order to setup a websockets proxy. But it drops permissions before opening the shell ..

Also, we could just move `create_wrapper` up to `web-lib-funcs.pl` , because really it has nothing to do with the Cron module. That would be even cleaner..

iliajie: Alright, I can do it. Do you still want to keep your latest patch, i.e. https://github.com/webmin/webmin/commit/db8d712d83f26eb6252aca6723b6cb97073fdea8?

iliajie: > Also, we could just move `create_wrapper` up to `web-lib-funcs.pl` , because really it has nothing to do with the Cron module. That would be even cleaner.

Done here https://github.com/webmin/webmin/pull/1887/commits/d154935e8f026dde5f402f530d7d2869f1dc9a51.

Issue #552, opened 2017-05-08
Logrotate configuration. Create missing log files.


chris001:
Couldn't Virtualmin be fixed to just work, by `touch`ing these missing files by itself, so that you don't have to manually go to the command line two times, and enter the touch command for each file ?

```
Failed to save enabled features : Current logrotate configuration is not valid :

error: stat of /var/log/virtualmin/server6.mydomain.com_access_log failed: No such file or directory

. This must be fixed, or log file rotation will not be done.
```

```
Failed to save enabled features : Current logrotate configuration is not valid :

error: stat of /var/log/virtualmin/server6.mydomain.com_error_log failed: No such file or directory

. This must be fixed, or log file rotation will not be done.
```

jcameron:
Yesss ... but they should never be missing in the first place, unless there is some bug in which an old incorrect logrotate config file was left around.

Is this a deleted domain, or does it still exist on your system?

chris001:
> Is this a deleted domain, or does it still exist on your system?
Good question.  If I remember correctly... while troubleshooting a DNS issue, I deleted the virtual server, which was empty anyway, in order to see if recreating the virtual server would fix the issue.  It was the only virtual server in virtualmin.  Virtualmin bugged out and gave an error and didn't want to delete the last virtual server.  So maybe it left the logrotate config file around.

jcameron:
Right, that seems like the real cause. The fix in this case is to delete the entries for those missing logs from logrotate.conf

PR #1893, opened 2023-04-25
Fix to work-around the bug with time offsets #552

iliaross:


jcameron:
This doesn't seem like a good idea! We should always store Unix times in GMT, and if there's any change needed for display purposes that should happen in make_date.

iliajie:
> This doesn't seem like a good idea! We should always store Unix times in GMT, and if there's any change needed for display purposes that should happen in make_date.

The problem is not in `make_date` though, right?

jcameron:
Isn't it? The time() function returns a Unix time integer that should be the same regardless of which timezone the system is in, unless the user's machine time is set wrong.

iliajie:
> Isn't it? The time() function returns a Unix time integer that should be the same regardless of which timezone the system is in, unless the user's machine time is set wrong.

Well, this is super confusing. I now can also see that `time()` returns the same timestamp regardless of where it's called. Before it was't the case whatsoever! I will have to take another look.

Also, the incorrect time is displayed in **Recent Logins** accordion only, and both timestamps are run through `make_date` sub, and the issue happens only for the given time zone. Perhaps, when run via CLI it changes something? I will have to check again!

Issue #1894, opened 2023-04-25
It appears that "-G" could be passed as an argument to 'pdbedit'


farglabs:
https://github.com/webmin/webmin/blob/3f322bee84016589770812f30aad5ebfea009c46/samba/samba-lib.pl#L649

This would throw an error.

martelrotschy:
It appears that the `-G` worked the same as `-g` in an older version of pdbedit (even though the docs  never included G as a valid option).

I confirmed that this is broken with pdbedit version 4.15.13-Ubuntu, but worked on version 4.13.14-Ubuntu.

Since it was never in the docs, I am not going to open an issue for samba-common utils.

PR #1895, opened 2023-04-26
Fixes webmin/webmin#1894

martelrotschy:
Invalid option for pdbedit in samba-lib.pl that broke due to a recent update of samba-common


jcameron:
Thanks for the patch!

PR #1902, opened 2023-05-10
Update Apache HTTP Server Support to V. 2.4 from Expired V. 2.2 on FreeBSD and Solaris Unix

MSSEsq:
FreeBSD and Solaris Unix provide different ports and packages for Apache HTTP Server and Modules for versions 2.4 and 2.2. Support by Webmin of Apache 2.4 has lagged on these systems due to the expired versions 2.2 still being referenced in the original versions of the files edited in this fork. It seems that the names of the configuration files in the Webmin apache directory determine which OS versions they apply to. I was not able to find any files in the Webmin project that reference these configuration files by name, so it appears that no other files require editing to enable support of Apache 2.4.


jcameron:
Do all FreeBSD 11+ and Solaris 11+ systems use these new paths for Apache by default? I'm worried that if we change the default paths, it will break systems that are still using Apache 2.2..

jcameron:
Thanks for the patch!

PR #1904, opened 2023-05-10
Add support for paginated tables in Users and Groups module

iliajie:
Jamie, check this small PR with an example on how we can replace **_old_** search forms, i.e.:

image

.. with the **_new_** pagination table, e.g.:

image

I like paginated table more than the old search form, because search form didn't list any data, which made it more difficult to list through the users and groups.

Please let me know if you have any questions. I have made a nightly Webmin built to include these changes!

PR #1905, opened 2023-05-12
Add ability to pass browser's client height to the server

iliajie:
This is a take out of the [other PR](https://github.com/webmin/webmin/pull/1904#issuecomment-1545236094) for a better review.


chris001:
Excellent, looks like solid progress towards responsive web design - [only CSS is needed](https://levelup.gitconnected.com/how-to-create-responsive-website-only-with-css-df4e8a1e8e4e).  CSS detecting the size and resolution of user device screen then dynamically calculate page layout and default font size.

iliajie:
> Excellent, looks like solid progress towards responsive web design - [only CSS is needed](https://levelup.gitconnected.com/how-to-create-responsive-website-only-with-css-df4e8a1e8e4e).

Don't we do it already?

chris001:
> Don't we do it already?

The page layout is responsive, however, if [the font sizes would automatically recalculate for optimum readability](https://www.smashingmagazine.com/2022/01/modern-fluid-typography-css-clamp/), this would be golden.

iliajie:
We will look into it in time, in a major UI upgrade.

PR #1908, opened 2023-05-17
Add API to check if given URL redirects somewhere

iliajie:
Jamie, check this PR, which we discussed a while a ago. It adds `get_http_redirect()` API to test if given URL redirects somewhere. It works correctly with "one-level" redirects. I am not sure how to make it follow all redirects cleanly.

The given redirect as `https://google.com` will return:

```
....
  'host' => 'google.com',
  'page' => '/',
  'port' => '443',
  'proto' => 'https',
  'redir' => {
    'host' => 'www.google.com',
    'path' => '',
    'port' => '443',
    'proto' => 'https',
    'url' => 'https://www.google.com'
  },
  'url' => 'https://google.com/'
}
```


web-lib-funcs.pl
```diff
@@ -12997,6 +12997,126 @@ sub get_default_system_locale
...
+
+# Original request
+my %rs;
+$rs{'url'} = $url;
+$rs{'proto'} = $proto;
+$rs{'host'} = $host;
+$rs{'port'} = $port;
+$rs{'page'} = $page;
+eval "use LWP::UserAgent";
+if (!$@) {
```
jcameron, line 13039: Why not just always use webmin's built-in http download code if you're going to have to support that case anyway?

iliajie: > Well, at first Webmin code in this PR needs to be able to handle multiple redirects. Also, I just thought that using `LWP::UserAgent` is more straight forward, as it is very well supported and many users use it..

Although, I think it doesn't work the same with the local connections. Perhaps, it's indeed better to rely only on Webmin download code?

jcameron: Best to use only the Webmin code to keep things simpler

web-lib-funcs.pl
```diff
@@ -12997,6 +12997,142 @@ sub get_default_system_locale
...
+	my @headers;
+	push(@headers, [ "Host", $host ]);
+	push(@headers, [ "User-agent", "Webmin" ]);
+	push(@headers, [ "Accept-language", "en" ]);
+
+	# Actually download it
+	$main::download_timed_out = undef;
+	local $SIG{ALRM} = \&download_timeout;
+	alarm($timeout);
+	my $h = &make_http_connection($host, $port, $ssl, "GET", $page, \@headers);
```
jcameron, line 13087: If $h is an error message string rather than a ref, we should fail the check right here.

iliajie:
> Best to use only the Webmin code to keep things simpler

Alright, fixed. Although, how do we reach the final redirect?

> If $h is an error message string rather than a ref, we should fail the check right here.

Redirect may never work, why failing?

jcameron:
What I mean is, how does this function indicate that it was unable to fetch the URL? It should have some way of returning an error status ..

iliajie:
> What I mean is, how does this function indicate that it was unable to fetch the URL? It should have some way of returning an error status

Yes, correct! We are doing this already. Here is the return on error:

```
{
  'error' => 'Failed to lookup IP address for googles.com',
  'host' => 'googles.com',
  'page' => '/about',
  'port' => '80',
  'proto' => 'http',
  'redir' => {},
  'url' => 'http://googles.com'
}
```

iliajie:
I will look into implementing a redirect that can do multiple hops..

iliajie:
Jamie, alright, support for multi hops in redirect is now supported. Tested and working for me!

Give it a try:

```Perl
my $url = 'https://google.com';
my $rs = &get_http_redirect($url, '/about');

var_dump("Redirect host: $rs->{'redir'}->{'host'} ; redirect path: $rs->{'redir'}->{'path'}");
# Redirect host: about.google ; redirect path: 

var_dump($rs);
# {
#   'handle' => {
#     'buffer' => '<HTML><HEAD><meta http-equiv="content-type" content="text/html;charset=utf-8">
# <TITLE>301 Moved</TITLE></HEAD><BODY>
# <H1>301 Moved</H1>
# The document has moved
# <A HREF="https://about.google/">here</A>. 
# </BODY></HTML> 
# ',
#     'fh' => '1684401381127137',
#     'ssl_con' => '93855143229168',
#     'ssl_ctx' => '93855143216576'
#   },
#   'host' => 'www.google.com',
#   'path' => '/about',
#   'port' => '443',
#   'proto' => 'https',
#   'redir' => {
#     'hops' => 2,
#     'host' => 'about.google',
#     'path' => '',
#     'port' => '443',
#     'proto' => 'https',
#     'resolved' => {
#       'ipv4' => '216.239.32.29',
#       'ipv6' => undef
#     },
#     'url' => 'https://about.google'
#   },
#   'url' => 'https://www.google.com'
# }
```

iliajie:
Jamie, here is the final version. The returned data will be as follows and will also include details of each hop, e.g.:

```Perl
{
  'hops' => [
    {
      'host' => 'google.com',
      'path' => '/about',
      'port' => '443',
      'proto' => 'https',
      'redir' => {
        'host' => 'www.google.com',
        'path' => '/about',
        'port' => '443',
        'proto' => 'https',
        'resolved' => {
          'ipv4' => '172.217.22.68'
        },
        'url' => 'https://www.google.com/about'
      },
      'response' => '<HTML><HEAD><meta http-equiv="content-type" content="text/html;charset=utf-8"><TITLE>301 Moved</TITLE></HEAD><BODY><H1>301 Moved</H1>The document has moved<A HREF="https://www.google.com/about">here</A>. </BODY></HTML>',
      'url' => 'https://google.com'
    },
    {
      'host' => 'www.google.com',
      'path' => '/about',
      'port' => '443',
      'proto' => 'https',
      'redir' => {
        'host' => 'about.google',
        'path' => '',
        'port' => '443',
        'proto' => 'https',
        'resolved' => {
          'ipv4' => '216.239.32.29'
        },
        'url' => 'https://about.google'
      },
      'response' => '<HTML><HEAD><meta http-equiv="content-type" content="text/html;charset=utf-8"><TITLE>301 Moved</TITLE></HEAD><BODY><H1>301 Moved</H1>The document has moved<A HREF="https://about.google/">here</A>. </BODY></HTML> ',
      'url' => 'https://www.google.com'
    }
  ],
  'host' => 'about.google',
  'path' => '',
  'port' => '443',
  'proto' => 'https',
  'resolved' => {
    'ipv4' => '216.239.32.29'
  },
  'url' => 'https://about.google'
}
```

The error would look like this:

```Perl
{
  'error' => 'Failed to lookup IP address for webmin.con'
}
```

The usage is simple. Using `google.com` as an example:

  1. Pass to the sub the URL like `https://google.com`, e.g. `get_http_redirect('https://google.com')` and in the end get `$rs->{'host'}` with `www.google.com` redirect result with 1 hop
  2. Pass to the sub the URL like `https://google.com` and page like `/about`, e.g. `get_http_redirect('https://google.com', '/about')` and in the end get `$rs->{'host'}` with `about.google` redirect  result with 2 hops

web-lib-funcs.pl
```diff
@@ -12997,6 +12997,146 @@ sub get_default_system_locale
...
+	$rs{'response'} = $h->{'buffer'};
+	}
+else {
+	$rs{'error'} = $h
+	}
+if (ref($h)) {
+	if ($h->{'buffer'} =~ /has\s+moved\s+<a\s+href=['"](.*?)['"]/mi) {
+		my $rurl = $1;
+		$rurl =~ s/\/$//;
+		$rs{'redir'}->{'url'} = $rurl;
```
jcameron, line 13088: Won't this generate a perl warning if the 'redir' key isn't set?

web-lib-funcs.pl
```diff
@@ -12997,6 +12997,146 @@ sub get_default_system_locale
...
+		$port ||= $uport;
+		$rs{'redir'}->{'proto'} = $proto;
+		$rs{'redir'}->{'host'} = $host;
+		$rs{'redir'}->{'port'} = $port;
+		$rs{'redir'}->{'path'} = $path;
+		}
+	}
+
+# Finally test if redirected URL can be resolved
+my $redir_host = $rs{'redir'}->{'host'};
```
jcameron, line 13103: Also here

web-lib-funcs.pl
```diff
@@ -12997,6 +12997,146 @@ sub get_default_system_locale
...
+if (ref($h)) {
+	&write_http_connection($h, "\r\n");
+	}
+&complete_http_download($h, undef, \$error, undef, undef, $host, $port,
+			undef, $ssl, 1, $timeout);
+if (ref($h)) {
+	$rs{'response'} = $h->{'buffer'};
+	}
+else {
+	$rs{'error'} = $h
```
jcameron, line 13082: Couldn't you just return at this point?

web-lib-funcs.pl
```diff
@@ -12997,6 +12997,146 @@ sub get_default_system_locale
...
+			my $rport = $redirect{'redir'}->{'port'};
+			$rport = undef if ($rport =~ /80|443/);
+			$rport = ":$rport" if ($rport);
+			my $rpath = $redirect{'redir'}->{'path'};
+			$rpath = undef if ($rpath eq "/");
+			return &get_http_redirect("$proto://$redir_host$rport", $rpath, $timeout);
+			}
+		}
+	}
+%redirect = %redirects if (keys %redirects);
```
jcameron, line 13126: What's going on with this code?

web-lib-funcs.pl
```diff
@@ -12997,6 +12997,146 @@ sub get_default_system_locale
...
+my %original_redirect = %redirect;
+%redirect = %{$redirect{'redir'}};
+if ($original_redirect{'error'}) {
```
jcameron, line 13129: Doesn't the function above return early if there's an error?

iliajie:
> Doesn't the function above return early if there's an error?

Oh, right. We don't need this anymore!

> What's going on with this code?

We're looking for the last known redirect.

jcameron:
> We're looking for the last known redirect.

The code is kind of confusing ... what are the %redirect and %redirects hashes used for?

iliajie:
> what are the %redirect and %redirects hashes used for?

`%redirect` hash is used to store the data about currently running process and `%redirects` stores the data about last successful redirect.

jcameron:
OK I see now - I would have preferred better variable names, but I'll allow it!

iliajie:
> I would have preferred better variable names, but I'll allow it!

Thanks, I'll consider using a better variable names!

Issue #1907, opened 2023-05-15
Upload and download module should not show HTML files


jcameron:
Seems potentially risky if the root user shows an un-trusted HTML file in the browser..

However, I did some tests and this would only happen if the "Show in browser if possible?" option is set to Yes. @iliajie does that match your experience?

![image](https://github.com/webmin/webmin/assets/102706/a09a75a7-4cc3-4047-b54f-2130ca0d7396)

iliajie:
Also, it's worth mentioning that detection of file type can be tricky. For example `&guess_mime_type()` sub and `file` command produce slightly different results on the file with mixed content.

For example, if we create `text.txt` file and place in it:

```
hello world
```

It will be recognized as `text/plain` in both cases. However, if we add to that file:

```
<script>alert(1234)</script>
```

Then `&guess_mime_type()` would still call that file `text/plain`, while `file` command output would return `HTML document, ASCII text`.

Although, luckily (and correctly) a browser would also treat that kind of file as plain text.

jcameron:
Surely all that matters is what MIME type we set in the response header?

iliajie:
> Surely all that matters is what MIME type we set in the response header?

Oh, actually, right! There is a better solution, I think we can always force `text/plain` header for any kind of text files. What do you think of this patch?

https://github.com/webmin/webmin/commit/5e9f137adfc28226121a351b2f78cc8db73630a1

jcameron:
No I don't like that approach, the regexp is too broad as it would match all kinds of MIME types that include the word "text". Please undo it, and I'll work on a safer solution, perhaps using the `filter_javascript` function.

jcameron:
I think this is a better fix : https://github.com/webmin/webmin/commit/ab44aa802343f2ae4b2a95852436e12a48603b0c

jcameron:
Oh god, I didn't know that SVG files could contain javascript!!

That patch won't entirely work, because filtering has to be done on the entire file at once in case some JS crosses the buffer boundary. But otherwise it's good ...

jcameron:
Yeah, `read_file_contents` would work just as well.

I'm not sure what other types are potentially dangerous to display..

chris001:
[...]

PR #1910, opened 2023-05-20
Add ability to read potentially dangerous files safely or force download

iliajie:
This PR is made on request and in continuation of [this](https://github.com/webmin/webmin/issues/1907#issuecomment-1555642036) discussion.


updown/fetch.cgi
```diff
@@ -80,14 +80,65 @@ if ($ENV{'PATH_INFO'}) {
...
+		my @st = stat($file);
+		my $fsize = $st[7];
+
+		# Get and analyze the file contents first
+		my $fdata = "";
+		my $dangertypes = $type =~ /html|xml|pdf/i;
+		my $htmltype    = $type =~ /html/i ? 1 : 0;
+		my $pdftype     = $type =~ /pdf/i ? 'pdf' : 0;
+		my $bsize =
+		    $dangertypes ? $fsize : &get_buffer_size_binary();
```
iliajie, line 93: >> [jcameron](https://github.com/jcameron) [3 hours ago](https://github.com/webmin/webmin/commit/3de93296a93c3843c7f6bc9f31c51996bd1de57b#r114170267)
>> It's not possible to read an entire file with the read() function .. instead use read_file_contents

---

> It's not possible to read an entire file with the read() function .. instead use read_file_contents

I started checking on it, to double check if it is possible to slurp the entire file using `read()` function .. and it appears to be possible .. read function in the third argument expects a _length_ of bytes to be read. If we pass in _length_ the actual file size, we read its full content in one go, isn't it right?

Also, [this article](https://www.perl.com/article/21/2013/4/21/Read-an-entire-file-into-a-string/) points out that it is also possible. Could you please double check?

iliajie:
>> [jcameron](https://github.com/jcameron) commented [3 hours ago](https://github.com/webmin/webmin/issues/1907#issuecomment-1555642036) 
Can you do this change as a PR so we can discuss it before committing? I think this solution is just too complex and potentially risky, and I really don't like the idea of inserting a message into the HTML!
Better to just do the simplest and safest thing, which is to only show file types in the browser if we are 100% sure they are safe (text, GIF, PNG and JPG). Actually, I suspect that most users of the file manager and upload/download modules just want to fetch a file to their PC, rather than viewing it in the browser.

---

> Can you do this change as a PR so we can discuss it before committing?

Alright, done in this [PR #1910](https://github.com/webmin/webmin/pull/1910).

> I think this solution is just too complex 

It looks straight forward to me. Sorry.

> and potentially risky

I wonder why is this risky though?

> and I really don't like the idea of inserting a message into the HTML!

Well, I wouldn't mind to adjust the message, however if we're displaying a file by modifying, we really have to tell the user, I think. Because it may look just fine and safe while pre-viewing it with Webmin (simply because we silently filtered out all JavaScript), and unintentionally we could trick the user in thinking that opening the same file after the download on the local computer is just as safe.

> Better to just do the simplest and safest thing, which is to only show file types in the browser if we are 100% sure they are safe (text, GIF, PNG and JPG).

A user most likely would want to have text files displayed nicely in the browser too.

> Actually, I suspect that most users of the file manager and upload/download modules just want to fetch a file to their PC, rather than viewing it in the browser.

Probably, however we have had this feature for decades removing it now would mostly affect someone negatively.

jcameron:
> I wonder why is this risky though?

Because if we promise a file is safe by displaying it in the browser, we have to be 100% certain that some new HTML tag that comes along isn't going to introduce a security risk. Filtering via regexp is just inherently risky.

> A user most likely would want to have text files displayed nicely in the browser too.

Yes, we should definitely allow this.

> Well, I wouldn't mind to adjust the message, however if we're displaying a file by modifying, we really have to tell the user

Yeah, but editing arbitrary HTML is pretty tricky .. we might end up corrupting it or making it un-displayable by inserting a message.

> Probably, however we have had this feature for decades removing it now would mostly affect someone negatively.

True, but as the recent vulnerability report informed us, there is a risk and we should fix it.

iliajie:
> Yeah, but editing arbitrary HTML is pretty tricky .. we might end up corrupting it or making it un-displayable by inserting a message.

Agreed! Fixed [here](https://github.com/webmin/webmin/pull/1910/commits/c4866735ba254120f1fb7a0a599cd6ba155af207). It will always download a file with JavaScript in it.

> Because if we promise a file is safe by displaying it in the browser, we have to be 100% certain that some new HTML tag that comes along isn't going to introduce a security risk.

Yes, I know! This is never-ending debate, been happening for decades, yet we still have `filter_javascript` function, expecting that it's better than doing no filtering at all. The best solution would be is to use parsers (like HTML parse) instead of using _regex_. Although, we cannot do it that easily in Webmin.

Also, I'd be interested to see not in theory but in practice the implementation, where our _regexes_ would fail and malicious JavaScript code would still be executed.

> True, but as the recent vulnerability report informed us, there is a risk and we should fix it.

Yes, I expect all my patches to fix the issue. If you always wish to download HTML and XML files, we can do that too. However, it will affect usability with normal (safe HTML and XML, SVG) files.

jcameron:
Thanks, that's an improvement. There is always a small risk of missing some JS though if it's encoded in a way we don't expect..

PR #1913, opened 2023-05-25
Fix for failure to re-start DNS BIND on slave servers

johnheenan:
See https://github.com/virtualmin/virtualmin-gpl/issues/565


jcameron:
Thanks! I will include this in the next release ..

PR #1915, opened 2023-05-29
Add support for properly displaying HTML emails

iliaross:
https://github.com/webmin/usermin/issues/99#issuecomment-1567141814


jcameron:
Thanks for this patch! Why is the change to `disable_html_images` needed though, and what does it do?

iliajie:
> Thanks for this patch! Why is the change to `disable_html_images` needed though, and what does it do?

Instead of completely removing images from HTML emails, possibly breaking layout, it is preferable to retain an element in its place and display it as a pink square, e.g.:

|Before|After|
|--|--|
|image|image
|

mailboxes/folders-lib.pl
```diff
@@ -2494,14 +2494,15 @@ sub safe_uidl
...
+my ($h2, $lynx);
 if (($h2 = &has_command("html2text")) || ($lynx = &has_command("lynx"))) {
 	# Can use a commonly available external program
 	local $temp = &transname().".html";
 	open(TEMP, ">", $temp);
-	print TEMP $_[0];
+	print TEMP $html;
 	close(TEMP);
-	open(OUT, ($lynx ? "$lynx -dump $temp" : "$h2 $temp")." 2>/dev/null |");
+	open(OUT, ($lynx ? "$lynx --display_charset=utf-8 -dump $temp" : "$h2 $temp")." 2>/dev/null |");
```
jcameron, line 2505: Unrelated to this PR, but we should really remove this ancient use of `lynx` to do HTML conversion!

iliajie: > `lynx` actually is really good at doing the job! 

.. better than `html2text` and also better than `HTML::TreeBuilder` and `HTML::FormatText`..

PR #1917, opened 2023-06-01
Add support to load email images using the server (for privacy)

iliajie:
Jamie, this is something we discussed with you yesterday. It's implemented in this PR.

It is now possible for emails to fetch images using server (not client in the browser).

Related Usermin [PR](https://github.com/webmin/usermin/pull/101).


XHR.cgi
```diff
@@ -0,0 +1,17 @@
+#!/usr/local/bin/perl
+# A caller for loading XHR related routines
+use strict;
+
+our ($trust_unknown_referers, $root_directory);
+
+$trust_unknown_referers = 1;
```
jcameron, line 7: Is it safe to turn off referer checks?

mailboxes/folders-lib.pl
```diff
@@ -2798,17 +2798,31 @@ sub disable_html_images
...
+		my $imgcont = $allimg;
+		$imgcont =~ s/src=/data-presrc=/g;
+		$imgcont =~ s/\Q$img\E/$imgid/g;
+		$newhtml .= $before.$imgcont;
+		$masked_img++;
+		# Try switching to remote user to correctly
+		# set temporary directory for super-users
+		&switch_to_remote_user();
+		# Save image URL reference
+		&write_file_contents(&tempname($imgid), $img);
```
jcameron, line 2813: What's going to clean up these temp files?

web-lib-funcs.pl
```diff
@@ -266,6 +266,27 @@ sub html_escape
 return $tmp;
 }
 
+=head2 html_unescape(string)
```
jcameron, line 269: Aren't there way more entities (like the numeric ones) that covered here?

iliajie: > Aren't there way more entities (like the numeric ones) that covered here?

Yeap! This patch https://github.com/webmin/webmin/pull/1917/commits/8d66b62fcede2c174409ffb0a0a46ce3bcd6a67c will address it.

XHR.cgi
iliajie: > Is it safe to turn off referer checks?

Yes, considering what this script does. However, it is not really necessary. I did it for testing purposes. Fix in this commit https://github.com/webmin/webmin/pull/1917/commits/4ee13f0e2059ee0b33fd171961d69b26cbdcbe1a.

mailboxes/folders-lib.pl
iliajie: > What's going to clean up these temp files?

Check this code https://github.com/webmin/webmin/pull/1917/files#diff-f6a5dfbecf6813723404d2314ff7df493f8fe396bcc853566d6d2571795392d7R26-R29.

jcameron: Can you explain to me what this code is doing and how it's used?

iliajie: > Can you explain to me what this code is doing and how it's used?

Yes, with pleasure. I will explain the whole process.

As we read an email and later filter images using `disable_html_images`, and if the mode is set to the new default `view_images=3` then:

  1. If email contains images we are not only extract URLs as before but also write a random temporary file with the given pattern (_hex_ string) in user's home, e.g. `/home/user1/.tmp/webmin/eacadbaaebfe2aafceb8bfff4cb0fcbdcefb316d`
  2. This temporary file has an actual image URL
  3. In place `src=` attribute name of an ``, is replaced with `data-presrc` and attribute value (link like `https://host/image.png`) is replaced with random _hex_ string we generated earlier
  4. We keep doing it further for each found image and we send it as part of HTML that is passed to the `iframe`
  5. Later, when the page with an email is loaded the [JavaScript code](https://github.com/webmin/webmin/pull/1917/files#diff-f89a4d333717390003e161a6a16387b29c9dd6b60f16160cd6ccf7da0c4ec493R2928-R2945) kicks in (all that is described happens within a split second)
  6. This code finds all images in an iframe with `data-presrc` attribute and makes a call to a new XHR API with the given, previously stored _hex_ string to find the temporary file that contains image URL
  7. When the file is found and read we delete it
  8. As we now have image pure URL we jump into downloading the linked file and storing its content in a temporary (trans) file
  9. As we have the image content written to the temp (trans) file we check on its MIME type to be added to the `Content-type:` header
  10. Finally we stream file's content back to the client
  11. Already mentioned JavaScript code receives blob data from the server and uses  `URL.createObjectURL`  to create a link.

That's all.

jcameron: Ok I see ... but instead of using a temp file referenced by a random hex ID, why not just pass the URL to the server and have xhr.cgi download that URL and stream the image back? That would avoid the need to create a temp file..

iliajie: You're right! Funny, I'm not sure why I used temp files at the first place. I will fix it as soon as I get back online.

iliajie: Jamie, check the latest commit. It should fix it and use no temp file.

XHR-lib.pl
```diff
@@ -0,0 +1,49 @@
...
+                # image URL (LinkedIn and other)
+                $url = &html_unescape($url);
+                my ($host, $port, $page, $ssl) = &parse_http_url($url);
+                my $img;
+                &http_download($host, $port, $page, \$img, undef, undef, $ssl, undef, undef, 10);
+                # Get MIME content type
+                eval "use File::MimeInfo";
+                my $img_tmp = &transname();
+                &write_file_contents($img_tmp, $img);
+                my $mime_type = mimetype($img_tmp);
```
jcameron, line 37: Can't we get the MIME type back as one of the headers from http_download? That seems safer than guessing it, and we then don't have to write to a temp file..

iliajie: I'll check, not sure off hand.

iliajie: Although, do you mean response headers? Can we really rely on them?

iliajie: > Can't we get the MIME type back as one of the headers from http_download? That seems safer than guessing it, and we then don't have to write to a temp file..

Jamie, check my two last commits. Does it look good to you?

jcameron: Looks better, but why not just always use the returned MIME type? It's guaranteed to be set..

XHR-lib.pl
```diff
@@ -0,0 +1,48 @@
...
+
+use strict;
+
+our (%in, %gconfig, $root_directory, $remote_user, $current_theme);
+
+sub xhr
+{
+
+my %data    = ();
+my $output  = sub {
```
jcameron, line 12: Why does this need to be a sub if it's only called once?

XHR-lib.pl
```diff
@@ -0,0 +1,48 @@
...
+    my ($data) = @_;
+    print "x-no-links: 1\n";
+    print_json($data);
+    };
+
+# Fetch actions
+if ($in{'action'} eq "fetch") {
```
jcameron, line 19: What happens if none of these if statements match? Shouldn't we call `error` ?

web-lib-funcs.pl
```diff
@@ -266,6 +266,31 @@ sub html_escape
...
+
+Converts HTML entities to the corresponding character
+
+=cut
+sub html_unescape
+{
+my ($str) = @_;
+if (!defined $str) {
+    return ''; # empty string
```
jcameron, line 278: Indent looks messed up?

web-lib-funcs.pl
```diff
@@ -2769,13 +2796,13 @@ sub http_download
 	else { &error(&html_escape($h)); }
 	}
 &complete_http_download($h, $dest, $error, $cbfunc, $osdn, $host, $port,
-			$headers, $ssl, $nocache, $timeout);
+			$headers, $ssl, $nocache, $timeout, \$$response_headers);
```
jcameron, line 2799: Shouldn't this just be $response_headers ?

web-lib-funcs.pl
```diff
@@ -10091,7 +10120,7 @@ =head2 guess_mime_type(filename, [default])
 sub guess_mime_type
 {
 my ($file, $def) = @_;
-if ($file =~ /\.([A-Za-z0-9\-]+)$/) {
+if ($file =~ /\.([A-Za-z0-9\-]+)(?:[?].*|)$/) {
```
jcameron, line 10123: What case does this handle?

XHR-lib.pl
iliajie: > Why does this need to be a sub if it's only called once?

This is meant to be called more than once (as code grows), similarly to my `xhr-lib.pl` in the theme.

XHR-lib.pl
iliajie: > What happens if none of these if statements match? Shouldn't we call `error` ?

Probably, yes! It would be a better design. I'll add error response in JSON..

web-lib-funcs.pl
iliajie: > What case does this handle?

This covers a case with an image having a tracker, i.e. `https://example.com/image1.png?uid=1234` or something like that.

web-lib-funcs.pl
iliajie: > Indent looks messed up?

Yes, it is. Fixed here https://github.com/webmin/webmin/commit/3d67e17cef426759eebc9a545727acad18eeeb8c.

web-lib-funcs.pl
```diff
@@ -2809,6 +2837,7 @@ sub complete_http_download
 	$header{lc($1)} = $2;
 	push(@headers, [ lc($1), $2 ]);
 	}
+$$response_headers = \%header;
```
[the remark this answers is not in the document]
iliajie: > Don't do this if $response_headers isn't set

Alright, fixed here https://github.com/webmin/webmin/pull/1917/commits/9a69913ccbc01c1cd86c1e91b08abda36872ad4d.

web-lib-funcs.pl
jcameron: Isn't response_headers a reference to the hash we want to fill in?

jcameron: Thanks!

web-lib-funcs.pl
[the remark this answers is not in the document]
jcameron: Commented on that commit

web-lib-funcs.pl
jcameron: I think this should be stripped off by the caller to guess_mime_type, not inside the function.

web-lib-funcs.pl
```diff
@@ -2809,6 +2838,8 @@ sub complete_http_download
 	$header{lc($1)} = $2;
 	push(@headers, [ lc($1), $2 ]);
 	}
+$$response_headers = \%header
```
jcameron, line 2841: Instead of using $$ , can't we just pass $response_headers around and then do `%$response_headers = %header` ?

iliajie: > Instead of using $$ , can't we just pass $response_headers around

Yes, check this patch https://github.com/webmin/webmin/commit/72c5e391212c26830ee7dd89b12e9e0db1a746a9.

> and then do `%$response_headers = %header` ?

No, it won't work like this!

XHR.cgi
```diff
@@ -0,0 +1,15 @@
...
+
+our ($root_directory);
+
+BEGIN { push(@INC, "."); };
+use WebminCore;
+
+&init_config();
+&ReadParse();
+&switch_to_remote_user();
+do "$root_directory/XHR-lib.pl";
```
jcameron, line 13: Shouldn't there be some kind of access control on this CGI, so that it can only be called by users with access to the `mailbox` or `mailboxes` modules? Otherwise any Virtualmin user could potentially use (or abuse) it to fetch URLs through the server, which is a potential DOS attack risk.

Seems like we could move it into `mailboxes` , and also remove XHR-lib.pl and put the one `xhr` function into XHR.cgi 

Also, why is it XHR.cgi instead of xhr.cgi to be consistent with the rest of Webmin?

web-lib-funcs.pl
jcameron: Thanks!

web-lib-funcs.pl
jcameron: It seems wrong to me for a function that's supposed to guess the MIME type of a filename to have special handling for URLs.

iliajie: No problem, I've reverted it.

XHR.cgi
iliajie: > Also, why is it XHR.cgi instead of xhr.cgi to be consistent with the rest of Webmin?

It conflicts with the Authentic theme, which is already using `xhr.cgi`.

> Shouldn't there be some kind of access control on this CGI, so that it can only be called by users with access to the mailbox or mailboxes modules? 

I thought `&init_config();` would take care of it?

> Seems like we could move it into mailboxes , and also remove XHR-lib.pl and put the one xhr function into XHR.cgi

I was hoping to use it universally for other modules too..

jcameron: So we have both XHR.cgi and xhr.cgi ? That's a little confusing ... I'd suggest renaming one.

Also, init_config in a top-level CGI script won't check for any module access, because it's outside of modules.

PR #1939, opened 2023-06-30
Add ability to downgrade database collation

iliajie:


mysql/mysql-lib.pl
```diff
@@ -1412,6 +1412,28 @@ sub list_collation_orders
...
+# fix_collation(file)
+# Fixes unsupported collations on restore, by replacing
+# unsuported with the closest supported variant
+sub fix_collation
+{
+my ($file) = @_;
+my ($version, $variant) = &get_remote_mysql_variant();
+if ($variant eq 'mariadb') {
+	my $tfile = &transname();
+	open(IN, '<' . $file) or die $!;
```
jcameron, line 1424: Is this safe to read and write to the same file? Seems like we should use a temp file ..

iliajie: > Is this safe to read and write to the same file? Seems like we should use a temp file ..

You're right. But this is what I already do .. ?

jcameron: Oh right I misread the filenames! Looks good ..

mysql/mysql-lib.pl
```diff
@@ -1034,7 +1034,7 @@ sub execute_sql_file
 			 : "";
 local $temp = &transname();
 &open_tempfile(TEMP, ">$temp");
-&print_tempfile(TEMP, "source ".$file.";\n");
+&print_tempfile(TEMP, "source ".&fix_collation($file).";\n");
```
jcameron, line 1037: You will need to make sure that this temp file returned by fix_collation has the same owner as the original file, because sometimes the `mysql` command gets run as the domain owner.

iliajie: Alright, how about that https://github.com/webmin/webmin/pull/1939/commits/b49d4f8147a5427d5ad1f3e9c98a39a75613d58e.

PR #1951, opened 2023-07-12
Fix to make errors less obscure

iliajie:


iliajie:
This PR will improve how error message will appear in general to make it less obscure. For example, the error about missing `at` module we've been discussing recently, will be displayed as follows, despite `error_stack` enabled or not, e.g.:

image

I will also need to make a few changes on the Authentic theme side ..

iliajie:
Actually we need to change it a bit, because it prints that technical information in standard error message, e.g.:

image

This patch will print it in a bubble instead:

image

PR #1953, opened 2023-07-16
Add PHP module improvements

iliaross:
1. Display tooltip showing default value for the given PHP option [#555](https://github.com/virtualmin/virtualmin-gpl/issues/555)
2. Fix not to display removed PHP options

### DIFF
--- a/phpini/edit_vars.cgi
+++ b/phpini/edit_vars.cgi
@@ -11,18 +11,19 @@ $conf = &get_config_as_user($in{'file'});
 print &ui_form_start("save_vars.cgi", "post");
 print &ui_hidden("file", $in{'file'}),"\n";
 print &ui_table_start($text{'vars_header'}, "width=100%", 4);
+if (php_version_test_maximum('7.4')) {
+	print &ui_table_row(&opt_help($text{'vars_magic'}, 'magic_quotes_gpc'),
+				&onoff_radio("magic_quotes_gpc"));
+	print &ui_table_row(&opt_help($text{'vars_runtime'}, 'magic_quotes_runtime'),
+				&onoff_radio("magic_quotes_runtime"));
+}
 
-print &ui_table_row(&opt_help($text{'vars_magic'}, 'magic_quotes_gpc'),
-		    &onoff_radio("magic_quotes_gpc"));
-
-print &ui_table_row(&opt_help($text{'vars_runtime'}, 'magic_quotes_runtime'),
-		    &onoff_radio("magic_quotes_runtime"));
-
-print &ui_table_row(&opt_help($text{'vars_register'}, 'register_globals'),
-		    &onoff_radio("register_globals"));
-
-print &ui_table_row(&opt_help($text{'vars_long'}, 'register_long_arrays'),
-		    &onoff_radio("register_long_arrays"));
+if (php_version_test_maximum('5.6')) {
+	print &ui_table_row(&opt_help($text{'vars_register'}, 'register_globals'),
+				&onoff_radio("register_globals"));
+	print &ui_table_row(&opt_help($text{'vars_long'}, 'register_long_arrays'),
+				&onoff_radio("register_long_arrays"));
+	}
 
 print &ui_table_row(&opt_help($text{'vars_args'}, 'register_argc_argv'),
 		    &onoff_radio("register_argc_argv"));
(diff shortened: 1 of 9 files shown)


phpini/phpini-lib.pl
```diff
@@ -305,18 +314,114 @@ sub get_php_ini_binary
...
+# -----------------------------------------------------------------------------
+#   caller: php_version_test_against(['7.4.34', '<='], 'php7.4');
+#   return: 0 because version 7.4.34 is not lower or equal than intalled (7.4.33)
+sub php_version_test_against
+{
+my ($version, $file) = @_;
+my $curr_php = &get_php_binary_version($file);
+return undef if (!$curr_php);
+my $cmp = '>=';
+if (ref($version) eq 'ARRAY') {
```
jcameron, line 390: Why not just make this function take two args, since it's always called with a comparator in this PR?

phpini/edit_vars.cgi
```diff
@@ -11,20 +11,21 @@ $conf = &get_config_as_user($in{'file'});
...
-		    &onoff_radio("magic_quotes_runtime"));
-
-print &ui_table_row($text{'vars_register'},
-		    &onoff_radio("register_globals"));
-
-print &ui_table_row($text{'vars_long'},
-		    &onoff_radio("register_long_arrays"));
-
-print &ui_table_row($text{'vars_args'},
+if (php_version_test_maximum('7.4')) {
```
jcameron, line 14: Don't you need to pass the file into php_version_test_maximum ?

phpini/edit_vars.cgi
```diff
@@ -11,20 +11,21 @@ $conf = &get_config_as_user($in{'file'});
...
+	print &ui_table_row(&opt_help($text{'vars_magic'}, 'magic_quotes_gpc'),
+				&onoff_radio("magic_quotes_gpc"));
+	print &ui_table_row(&opt_help($text{'vars_runtime'}, 'magic_quotes_runtime'),
+				&onoff_radio("magic_quotes_runtime"));
+}
+
+if (php_version_test_maximum('5.6')) {
```
jcameron, line 21: Also here

phpini/phpini-lib.pl
iliajie: As this is an API function, and not meant to be used directly, I just wanted to make it more flexible ..

phpini/edit_vars.cgi
iliajie: No, because it will be caught automatically, from `$in{'file'}` if not set.

phpini/edit_vars.cgi
iliajie: Same, here.

phpini/edit_vars.cgi
jcameron: It's a lot cleaner to pass `$in{'file'}` into this function as a parameter, rather than accessing %in from inside the function.

phpini/phpini-lib.pl
jcameron: Right but if we're not using that flexibility yet, it's not worth the complexity of adding it

phpini/edit_vars.cgi
iliajie: > It's a lot cleaner to pass `$in{'file'}` into this function as a parameter, rather than accessing %in from inside the function.

Jamie, yes maybe, and we always expect external callers to pass file parameter. However, doing it inside of `phpini` module will clutter a lot of `phpini`-_pages_, as we will also have to pass it to `opt_help` sub _everywhere_. It will look ugly and cluttered ..

jcameron: Oh it's needed by `opt_help` as well? Hmm .. in that case, I suppose we have no choice, even though I don't really like using global variables.

iliajie: > I don't really like using global variables.

Yeah, I also try to avoid using globals anytime possible.

phpini/phpini-lib.pl
iliajie: > Right but if we're not using that flexibility yet, it's not worth the complexity of adding it

Do you want me to fix it and pass two params instead?

jcameron: Yes please!

iliajie: Alright, fixed.

PR #1957, opened 2023-07-17
Add ability to present active interfaces the way network manager wants

iliajie:
https://github.com/webmin/webmin/commit/07a46ff56199d4c27ffc3c7da5163a30adae3ccb#commitcomment-121615422


jcameron:
This seems overly complex! Wouldn't it be better to fix active_interfaces to properly detect which IP is the primary one, perhaps based on the `dynamic` tag?

iliajie:
> Wouldn't it be better to fix active_interfaces to properly detect which IP is the primary one, perhaps based on the dynamic tag?

Well, it seems that we cannot 100% rely on `ip addr` output accuracy with NetworkManager. The order of `ip addr` command output is not straight forward with NetworkManager. We won't be able to get the proper order of interfaces. NetworkManager should be the source of truth, I believe.

Please have a close look into my code. It's very simple infact, and most importantly doing the job right!

iliajie:
I have reverted "overly complicated" code and made [changes](https://github.com/webmin/webmin/pull/1957/files) over your code instead.

jcameron:
Thanks, that seems better!

iliajie:
> Thanks, that seems better!

Cool!

PR #1987, opened 2023-08-19
Add support for `webmin main` repo

iliaross:
https://forum.virtualmin.com/t/warning-use-the-name-of-the-stable-distribution-instead-of-stable/122043/13?u=ilia

### DIFF
--- a/setup-repos.sh
+++ b/setup-repos.sh
@@ -166,7 +166,7 @@ deb)
   echo "  .. done"
   # Create repo file
   echo "  Setting up Webmin repository .."
-  echo "deb [signed-by=/usr/share/keyrings/$repoid_debian_like-$webmin_key_suffix.gpg] $webmin_download/download/newkey/repository stable contrib" >$debian_repo_file
+  echo "deb [signed-by=/usr/share/keyrings/$repoid_debian_like-$webmin_key_suffix.gpg] $webmin_download/download/repo webmin main" >$debian_repo_file
   echo "  .. done"
   # Clean meta
   echo "  Cleaning repository metadata .."
--- a/webmin/fix_repo.cgi
+++ b/webmin/fix_repo.cgi
@@ -44,8 +44,8 @@ foreach my $repo ($webmin_apt_repo_file, $global_apt_repo_file) {
 			}
 	};
 	foreach my $l (@$lref) {
-		if ($l =~ /^\s*deb\s+.*?((http|https):\/\/download.webmin.com\/download\/repository)\s+sarge\s+contrib/) {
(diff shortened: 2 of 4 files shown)


jcameron:
Does this change mean we'll need to create a new repo on download.webmin.com ?

iliajie:
> Does this change mean we'll need to create a new repo on download.webmin.com ?

Yeah, check this thread out:

https://forum.virtualmin.com/t/warning-use-the-name-of-the-stable-distribution-instead-of-stable/122043/20?u=ilia

jcameron:
I really don't want to maintain another repo to fix a cosmetic issue. Because users can take a long time to upgrade, anytime we add a new repo we still have to maintain the old one for potentially years..

iliajie:
Jamie, but I already described the process. You don't need to support multiple repos, right? Because you do it only once! The installed Webmin version 2.102 will switch any user to the _very_ new repo. All you will need to support is the new repo. You do it _this_ time for all 3 repos. And later then support only the latest (i.e `webmin main `) repo!

I'd really want to see repos done immaculately! Besides it's easy to do!

jcameron:
Doesn't that mean that users on versions older than 2.102 won't be able to upgrade beyond that release until they fix the repo, unless we maintain new versions in all 3 repos?

iliajie:
> Doesn't that mean that users on versions older than 2.102 won't be able to upgrade beyond that release until they fix the repo, unless we maintain new versions in all 3 repos?

No. And why do you think that you'd need to support all three repos?

A repo 1,2 and 3 will get Webmin 2.102 (once). Webmin 2.102, when installed will switch all those old repos to the new one. And in the future, repos 1 and 2 cab be abandoned and only the new repo 3, i.e. ` webmin main` would need to be supported ... right ... ?

jcameron:
Yes, but only if we can assume that users of repos 1 and 2 will click the button to fix their local repositories. Otherwise they'll be stuck on the old version..

iliajie:
> Yes, but only if we can assume that users of repos 1 and 2 will click the button to fix their local repositories. Otherwise they'll be stuck on the old version..

Considering that a user cannot dismiss this warning, we could expect with a high probability that Webmin repos will be fixed eventually..

jcameron:
Let's hold off on this change until we get a new webmin release out ..

iliajie:
Alright, I wouldn't mind ..

iliajie:
Are we moving on with this... ?

swelljoe:
I assume this PR is mooted by the repo refactor that has already been done?

PR #1989, opened 2023-08-25
Fix not to return short hostname unless set

iliaross:


jcameron:
I don't understand this change - what bug is it fixing?

iliajie:
> I don't understand this change - what bug is it fixing?

I had a hostname `host.alma9-pro.virtualmin.dev` and the returned result for the host name was incorrect, e.g. `alma9-pro.virtualmin.dev`

iliajie:
Although, I think it's rather a work-around though. I'm no longer confident about keeping this patch.

PR #1990, opened 2023-08-25
Add hostname detection using `hostnamectl` command

iliajie:


jcameron:
Should this happen before checking the various hostname files?

The current code checks files first because it's quicker than shelling out to a separate command....

iliajie:
I think it should happen before reading any files.

jcameron:
But why? This will slow down the get_hostname function which is called on every page view..

iliajie:
I'm pretty confident that it won't slow things down. I'm 
confident that running `hostnamectl` command isn't notably slower than reading a file. We could make speed tests though.

However, I wouldn't, as the main reason for this change is not about the speed, and if it's 1 or 2 ms faster or slower. The reason for this change is that `systemd` systems should get the hostname by using `hostnamectl` command rather than reading files directly. It's akin to changing _iptables_ directly when _firewalld_ is installed.

jcameron:
I guarantee that it's slower, because running a separate command involves forking a new process which may read many libraries and other files. But if this is a question of correctness because the files aren't in sync with `hostnamectl` , then this makes sense.

iliajie:
> But if this is a question of correctness because the files aren't in sync with `hostnamectl` , then this makes sense.

Yeah. It is!



> I guarantee that it's slower, because running a separate command involves forking a new process which may read many libraries and other files.

Makes sense! However, I'd like to run an actual tests to see the numbers ..

iliajie:
> I guarantee that it's slower, because running a separate command involves forking a new process which may read many

This is very true! However, for our case application it's only true if we don't cache the results. So, calling `get_system_hostname()` in a script to only get `hostnamectl --static` takes about 9 ms, while reading a file is about 7 ms. It's a small but still a difference. However, if cache is disabled, and we run `hostnamectl --static` x 1000 in a loop, then the slow down is _tremendous_ for command execution while reading a file is still super fast!

jcameron:
Ok, let's go with it then. This seems like the safer option ...

iliajie:
> Ok, let's go with it then. This seems like the safer option ...

Not sure what you mean but there is no problem using `hostnamectl` command ..

jcameron:
No, I mean it's good to use the `hostnamectl` command - the small performance cost is worth the gain in correctness.

iliajie:
> No, I mean it's good to use the `hostnamectl` command - the small performance cost is worth the gain in correctness.

Oh, alright!

PR #2002, opened 2023-09-15
Fix to save protocols to global config file

iliajie:
Jamie, hey I made an attempts to use `/usr/share/dovecot` included files only in read-only mode.

This PR works, but only partially.

If we save Networking and Protocols page the first time (when there is no `protocols` directives in main config file yet) and with IMAP only, it adds `protocols = imap` directive. All is good at this point and if you re-visit the page, the correct IMAP protocol is selected. However, if you save it the second time, instead of overwriting `protocols = imap` directive, it adds a new line to the Dovecot global config, but what's worse, it removes the line above.

It really seems to me as a bug in `dovecot::save_directive` sub, in a block where we _add some directives_. Back in time (long time ago) some people reported that some last `}` in the block was deleted. If I keep saving and saving Networking and Protocols page with this PR, the code is just keep deleting each line above the first directive, instead of replacing it, e.g.:

image

I wouldn't expect this happening. What is wrong here, could you explain?

Ref.: https://github.com/virtualmin/virtualmin-gpl/issues/639#issuecomment-1720679637

### DIFF
--- a/dovecot/dovecot-lib.pl
+++ b/dovecot/dovecot-lib.pl
@@ -33,18 +33,17 @@ return &get_config_file();
 # Returns a list of dovecot config entries
 sub get_config
 {
-my ($excl_tpl) = @_;
 if (!@get_config_cache) {
-	@get_config_cache = &read_config_file(&get_config_file(), undef, $excl_tpl);
+	@get_config_cache = &read_config_file(&get_config_file());
 	}
 return \@get_config_cache;
 }
 
-# read_config_file(filename, [&include-parent-rv], [exclude-templates])
+# read_config_file(filename, [&include-parent-rv])
 # Convert a file into a list od directives
 sub read_config_file
 {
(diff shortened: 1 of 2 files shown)


jcameron:
Seems like this won't solve the general problem of some config options being defined in files under /usr , only potentially the `protocols` line.

iliajie:
> I wouldn't expect this happening. What is wrong here, could you explain?

Duh, I forgot to increment the line number on next. Fix is here https://github.com/webmin/webmin/pull/2002/commits/3f0ab3fb92a35fad5bddd395e08ed37d43d9f3cf.

It all works now as expected.

iliajie:
Alright, Jamie, I simplified the code https://github.com/webmin/webmin/pull/2002/files.

jcameron:
Thanks, this looks pretty good! I suppose there's a small risk that ALL the configs are under /opt, but we can fix that pretty easily..

PR #2003, opened 2023-09-17
Add die handler that stores the error from eval message

iliaross:


jcameron:
Will this still re-throw the exception so that it's caught by eval?

iliajie:
Yes, it does. Don't you see it happening in your tests?

web-lib.pl
```diff
@@ -38,7 +38,7 @@ sub unique
 # eval message in a global variable
 $SIG{__DIE__} = sub {
 	# Don't meddle with the natural functioning of eval.
-    $main::error_last_eval = $_[0] if (!$^S);
+	$main::error_last_eval = $_[0] if (!$^S);
```
jcameron, line 41: What is `$^S` ?

iliajie: > What is `$^S` ?

https://perldoc.perl.org/perlvar#$%5ES

`$^S` is a special Perl variable. It provides the status of an ongoing `eval`. Here's what `$^S` can hold:
 - `undef` if the code is running outside an `eval` block.
 - `true` if the code is being executed inside an `eval` block and the `eval` is parsing (i.e., it hasn't actually begun executing the block yet).
 - `false` if the code is running inside an `eval` block and the `eval` is executing.

The purpose of `if !$^S;` is to make the `__DIE__` handler triggered during the compilation phase of an `eval`.

jcameron:
Thanks, patch looks good!

PR #2013, opened 2023-09-29
Add theme switcher using hotkeys

iliajie:
Jamie, hi!

It has been a long time since I wanted to add this feature to switch themes in a quick and straight forward way (for variety of reasons).

I added the subroutine to load JavaScript for switching themes using hotkeys.

Hotkeys are:
	To activate theme switch mode:
	Use: `Ctrl+Alt+T` (or `Control+Option+T` on Mac).

	Immediately after, within 1 second, select your desired theme by pressing:
	- `Shift + A`: Authentic theme
	- `Shift + G`: Gray theme
	- `Shift + L`: Legacy theme.

If there are any objections, we may consider enabling it only in debug mode or with a specific option or even not include new `switch_theme.cgi` at all. Although, I don't find it to be problematic to just have it enabled.

Try applying the patches from this PR and see how convinient it is.


jcameron:
Do people really switch themes that often that we need a hotkey for it? Honestly I don't think we need this feature, since I expect 99.9% of users never change away from the Authentic theme..

iliajie:
Agreed. This is why I made it only available in debug mode or if error stack is enabled. It is made primarily for us.

web-lib-funcs.pl
```diff
@@ -4449,10 +4450,16 @@ =head2 get_webmin_version
 =cut
 sub get_webmin_version
```
jcameron, line 4451: What does this change do?

switch_theme.cgi
```diff
@@ -0,0 +1,47 @@
+#!/usr/local/bin/perl
+# switch_theme.cgi
+# Change the theme for the current user if allowed
+
+BEGIN { push(@INC, "."); };
+use WebminCore;
+
+&init_config();
+&ReadParse();
+my $err = sub {
```
jcameron, line 10: Why not just call error() ?

switch_theme.cgi
```diff
@@ -0,0 +1,47 @@
...
+    !&foreign_available('acl')) {
+    &$err("You are not allowed to change themes!"); 
+    }
+# Check if the theme is known
+&$err("Theme identification is not known!")
+    if (!$in{'theme'} || $in{'theme'} !~ /^[123]$/);
+# Check if the remote user is known
+&$err("Remote user is not known!") if (!$remote_user);
+# Define the theme
+my $themes = {
```
jcameron, line 34: Wouldn't it be simpler to pass in the theme directory instead of a number when switching?

switch_theme.cgi
```diff
@@ -0,0 +1,47 @@
...
+    '1' => 'authentic-theme',
+    '2' => 'gray-theme',
+    '3' => '' };
+my $theme = $themes->{$in{'theme'}};
+# Change the theme
```
jcameron, line 39: Better to use the acl module's modify_user API for this, as it will work with other backends like LDAP

switch_theme.cgi
iliajie: It won't look good in this case application.

web-lib-funcs.pl
iliajie: It removes dots from the version.

switch_theme.cgi
iliajie: No, we need to perform strict checking.

jcameron: Do you mean preventing use of any other themes?

switch_theme.cgi
jcameron: Why won't it look good?

web-lib-funcs.pl
jcameron: But what if existing callers expect to get the dots in the version?

switch_theme.cgi
iliajie: > Why won't it look good?

Because we're not in the module. Those extra buttons inside of a large container .. It's much cleaner just to print the message and exit.

switch_theme.cgi
iliajie: > Do you mean preventing use of any other themes?

Yes. We just want to allow use of the hardcoded  themes.

switch_theme.cgi
iliajie: We don't modify a user. We modify local copies of Webmin `config` and `miniserv.conf` files.

jcameron: Yes, but the `acl::modify_user` function will also update those files using an API that will work even if we change the underlying format for storing the theme setting.

iliajie: > Yes, but the `acl::modify_user` function will also update those files using an API that will work even if we change the underlying format for storing the theme setting.

But can theme parameter used by Webmin be stored somewhere else? What would be the practical purpose of using that sub, if we only need to update the local version of `config` and `miniserv.conf` files? Unless your suggestion is more of a theoretical one?

jcameron: Yes, we support storing users in LDAP, MySQL or PostgreSQL. Only the acl::modify_user function will update the correct data store for the user, which might not be miniserv.conf

iliajie: > Yes, we support storing users in LDAP, MySQL or PostgreSQL. Only the acl::modify_user function will update the correct data store for the user, which might not be miniserv.conf

Yes, but does this data include `preroot_` from `miniserv.conf` file?

jcameron: Yes when a user is in LDAP or MySQL, there is no preroot_ line in miniserv.conf

iliajie: > Yes when a user is in LDAP or MySQL, there is no preroot_ line in miniserv.conf

Cool! Thanks!

I have changed the code to use ACL `modify_user` insted!

ui-lib.pl
```diff
@@ -2280,6 +2280,31 @@ sub ui_radio_selector_javascript
...
+	- `Shift + A`: Authentic theme
+	- `Shift + G`: Gray theme
+	- `Shift + L`: Legacy theme.
+
+=cut
+
+sub ui_switch_theme_javascript
+{
+return &theme_ui_switch_theme_javascript(@_) if (defined(&theme_ui_switch_theme_javascript));
+return "" if (!$gconfig{'error_stack'} && !$gconfig{'debug_enabled'} && !$gconfig{'debug_theme_switcher'});
```
jcameron, line 2302: Let's not enable this just because `error_stack` is turned on ... it should have a separate setting. A lot of users enable error_stack.

iliajie: Alright, I removed enabling on `error_stack` option.

jcameron:
Thanks! Merging now ...

iliajie:
Great! Thanks!

web-lib-funcs.pl
jcameron: So it looks like this actually caused a problem in the end :-( I think you should undo this, and make the caller remove dots from the version if needed..

iliajie: Sadly, but this was because the theme called this sub incorrectly, passing `1` as a parameter.

I made it happen in the caller https://github.com/webmin/webmin/commit/d7de842f450ff4c25f79e76d255965ecb7fdf474.

Issue #2022, opened 2023-10-10
since last upgrade error about systemd on sysvinit systems


aqueos:
hi,

 since the last version we got plenty of error messages like

System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to connect to bus: Host is down

on sysvinit system on debian 10.

[~]: webmin --version
2.103

  distro => {
    codename => "buster",
    description => "Debian GNU/Linux 10 (buster)",
    id => "Debian",
    release => {
      full => "10.13",
      major => "10",
      minor => "13"
    }
  },

i think there is an issue in the detection about the init system used by the system.

best regards,
Ghislain.

iliajie:
Hello,

Does your system have `hostnamectl` command? If so, what is the output of `hostnamectl --static` on your system?

aqueos:
hi,

 it returns this message:

[~]:  hostnamectl --static 
System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to connect to bus: Host is down

best regards,
Ghislain.

iliajie:
> System has not been booted with systemd as init system (PID 1). Can't operate.

Very interesting. Why do you have _systemd_ install if you're using _sysvinit_?

Can you run `echo $?` right after calling `hostnamectl --static` and share its output if any?

Also, does `/etc/hostname` contain the correct and expected hostname?

aqueos:
here it is :  

```
[~]: hostnamectl --static; echo $?
System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to connect to bus: Host is down
1

```

well we have the basic systemd packages because some software require parts of systemd  to run like systemd-tmpfiles.

all the php fpm package of debian sury need systemd-tmpfiles so systemd must be installed installed but not active.

The hostname is set to the fqdn on all machines and is the rigth one.

best regards,
Ghislain.

iliajie:
Got it!

Alright, if you feel comfortable you can try this [latest](https://builds.webmin.dev/webmin-latest.deb) devel build of Webmin and let me know if that works for you.

aqueos:
dont laugh but on debian 10 i got:

```
dpkg -i webmin-latest.deb 
dpkg-deb: error: archive 'webmin-latest.deb' uses unknown compression for member 'control.tar.zst', giving up
dpkg: error processing archive webmin-latest.deb (--install):
 dpkg-deb --control subprocess returned error exit status 2
Errors were encountered while processing:
 webmin-latest.deb

```

[...]

PR #2026, opened 2023-10-11
Fix Webmin installation issues

iliajie:
Ref.: https://github.com/webmin/webmin/issues/2022#issuecomment-1755745916

### DIFF
--- a/init/updateboot.pl
+++ b/init/updateboot.pl
@@ -31,5 +31,9 @@ if ($product) {
 		&flush_file_lines($temp);
 		copy_source_dest($temp, "$systemd_root/$product.service");
 		system("systemctl daemon-reload >/dev/null 2>&1");
-		};
+		}
+	elsif (-d "/etc/init.d") {
+		copy_source_dest("$root_directory/webmin-init", "/etc/init.d/$product");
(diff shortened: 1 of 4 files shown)


iliajie:
I have been running today more deep tests with clean installs. CentOS 6 installation still ends up without `/etc/init.d/webmin` file. What can be wrong?

jcameron:
Where in setup.sh should this be happening? Seems like it should be here, but that block is empty : https://github.com/webmin/webmin/blob/master/setup.sh#L799

iliajie:
> Where in setup.sh should this be happening? Seems like it should be here, but that block is empty : https://github.com/webmin/webmin/blob/master/setup.sh#L799

Hm, indeed. I thought we called `init/updateboot.pl` for this purpose? We're calling it, right?

jcameron:
Isn't it atboot.pl that does it?  Also, is this failing for an RPM or tar.gz install?

iliajie:
Alright, it's all fixed on the latest commits.

Although, I think we also need to fix a problem when Webmin and Usermin doesn't open default 10000 and 20000 firewall ports upon installation ..

jcameron:
Please open a separate ticket for that, because it should happen if a supported firewall is in use..

PR #2044, opened 2023-11-06
Add API and support to preserve module params across config page

iliaross:


jcameron:
Is the purpose of this so that when there's a link from a module's index.cgi with URL params to the config page, that those params will be preserved when the config is saved and it redirects back to index.cgi ?

iliajie:
Yes, to redirect back to module index with initial (the same) params as before going to module config.

config-lib.pl
```diff
@@ -407,5 +407,89 @@ sub save_module_preferences
...
+# print_config_posted_params()
+# Prints hidden fields for all the parameters in the referrer URL
+sub print_config_posted_params
+{
+my $env_referer = $ENV{'HTTP_REFERER'};
+my $params_referer = &read_config_params($env_referer);
+my $params_url = &read_config_params($ENV{'REQUEST_URI'});
+my $params = { %$params_referer, %$params_url };
+my @params;
+if ($params) {
```
jcameron, line 438: $params is always going to be set since it's a hash ref

config-lib.pl
```diff
@@ -407,5 +407,89 @@ sub save_module_preferences
...
+return \%params;
+}
+
+# print_config_posted_params()
+# Prints hidden fields for all the parameters in the referrer URL
+sub print_config_posted_params
+{
+my $env_referer = $ENV{'HTTP_REFERER'};
+my $params_referer = &read_config_params($env_referer);
+my $params_url = &read_config_params($ENV{'REQUEST_URI'});
```
jcameron, line 435: Why do we also need to include the params from the current URL?

config-lib.pl
iliajie: Right! Fixed.

config-lib.pl
```diff
@@ -407,5 +407,89 @@ sub save_module_preferences
...
+my $params = { %$params_referer, %$params_url };
+my @params;
+if ($params) {
+	foreach my $param (keys %{$params}) {
+		my $param_prefix = "_cparam-";
+		if ($param =~ /section|module|mode|section_next|nnext|nprev|xnavigation/ ||
```
jcameron, line 441: What are all these params that are being filtered out?

jcameron:
Ok I see where you are going with this, and it makes sense. But there seems to be a lot of extra complexity beyond what is needed to get the URL params from the referrer..

config-lib.pl
iliajie: This is for section config Next and Previous buttons work when Save and Next is clicked.

config-lib.pl
iliajie: These are standard params that are used in config page forms.

iliajie:
> Ok I see where you are going with this, and it makes sense. But there seems to be a lot of extra complexity beyond what is needed to get the URL params from the referrer..

Yes, because we need to make it work with:

  1. Main **Save** form button and also with **Save and Next** button
  2. Sections in config, i.e. **Previous** and **Next** buttons
  3. Return to module index page (not a form but link at the very bottom of the page)

config-lib.pl
jcameron: Wouldn't it make more sense to use all the params from the referrer URL, and only params starting with _cparam- from the config page? Instead of combining them and then doing this filtering, which is fragile if another param ever gets added.

iliajie: Not really. Because we don't know the names of the params that are there in the module index before going to module's config page. The original params from index page passed in the refer are prefixed with `_cparam-` to distinguish from the standard config parameters. Moreover, later params starting with `_cparam-` has to be resubmitted across pages in the sections lister (Previous and Next buttons). 

Honestly, this is a very simple feature to have but very difficult to implement, as it involves a lot of variables, like normal link button to return to module index has to work. Forms should work when submitted and returned back to index restoring original params but also, when sections are used params needs to be passed and separated correctly, so index page doesn't receive the params in the end that weren't there at the first place.

So, no, I don't think we can simplify it in any way and have the same functionality in place.

iliajie: Probably, as I'm thinking now of it, I could separate form submits from links, and referer and requests in separate functions. But that would require a rework. It won't give any benefit in usability though.

jcameron: Maybe a better approach is for modules to have a way to pass the URL params to index.cgi into config.cgi, and then have it pass them along with the Next/Prev buttons and footer links? Relying on the referer seems tricky to me..

iliajie:
Well, I'm not sure. Referer must be always set.

The only way as I see it now is to pass it to module config link as URL params.

PR #2048, opened 2023-11-16
fix update-from-repo.sh -release:lastest: get version from raw.githubusercontent.com

gnadelwartz:
update-from-repo.sh -release:lastest was no more working because of changes on githup, changed to get  version from raw.githubusercontent.com instead of browsing HTML output.

now update to latest release should work for webmin and usermin again

Kay


iliajie:
Danke, Kay!

Issue #1097, opened 2019-08-25
[feature suggestion] support ntftables


n8v8R:
When searching for it found https://sourceforge.net/p/webadmin/discussion/600155/thread/8cf48bd5/

which seems to have gone stale/unheeded.

Whilst `ipt` might be around for a while longer my nodes since long are utilizing `nft`. It would be handy if support for it could be added.

jcameron:
This isn't on our radar yet, sorry.

chris001:
Here's the cause of the bug/issue, and solution.
On debian 10 Buster, `iptables` aliases to the new `nftables` binaries, causing failures.  
The popular `webmin` module for `csf` also fails, because it depends on the old `iptables`, while debian 10 buster only has the new `iptables` (aliased to `nftables`).
Luckily `nftables` comes with a compatibility layer with all the classic old `iptables` commands.
https://forum.configserver.com/viewtopic.php?f=6&t=11372&p=30603&hilit=nftables
To use `iptables-legacy`, first detect `/etc/debian_version` :
```
# update-alternatives --set iptables /usr/sbin/iptables-legacy
# update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy
# update-alternatives --set arptables /usr/sbin/arptables-legacy
# update-alternatives --set ebtables /usr/sbin/ebtables-legacy
```
To switch back:
```
# update-alternatives --set iptables /usr/sbin/iptables-nft
# update-alternatives --set ip6tables /usr/sbin/ip6tables-nft
# update-alternatives --set arptables /usr/sbin/arptables-nft
# update-alternatives --set ebtables /usr/sbin/ebtables-nft
```
Source: https://wiki.debian.org/nftables#Current_status

jcameron:
Good to know. We haven't started working on any native support for `nftables` yet.

larsen0815:
Any idea when this might be available?

Daniel15:
Sorry for the bump but is there any ETA on this? iptables is legacy now and I'm not sure how much longer Debian will support it for.

iliarostovtsev:
@jcameron Jamie, perhaps it would be easier (at least for now) to just use a translate command, which will easily convert Iptables syntax rules to Nftables, for example:


```
iptables-translate -A INPUT -p tcp --dport 22 -m conntrack --ctstate NEW,ESTABLISHED -j ACCEPT
```

.. would return an equivalent rule for Nftables:

```
[...]

PR #2064, opened 2023-12-20
Simple nftables support via the use of update-alternatives

TheBeardedQuack:
Simply support for nftables (which is a drop-in replacement command line wise), simply by preferring to call `iptables-save|restore` rather over `iptables-legacy-save|restore`

Should be a simple resolve on #1097 for many people, and should work oob for any distros that set this alias up by default.

I'm unsure why the restore command was being run after a call to cd, is this a preventative measure for directing to the filepath? If so save should've had it also. If not, the `cd` should be removed from both, or a comment should be added to specify exactly why the move is required.

### DIFF
--- a/firewall/firewall-lib.pl
+++ b/firewall/firewall-lib.pl
@@ -407,13 +407,18 @@ local $out = &backquote_logged("cd / && $rcmd <$ipvx_save 2>&1");
 return $? ? "<pre>$out</pre>" : undef;
 }
 
+sub iptables_save_command
+{
+return &has_command("ip${ipvx}tables-legacy-save") ||
+       &has_command("ip${ipvx}tables-save");
+}
+
 # iptables_save()
 # Saves the active firewall rules, and returns any error
 sub iptables_save
 {
-local $scmd = &has_command("ip${ipvx}tables-legacy-save") ||
-	      "ip${ipvx}tables-save";
-local $out = &backquote_logged("$scmd >$ipvx_save 2>&1");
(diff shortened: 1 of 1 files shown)


TheBeardedQuack:
Setup system for use with nftables.

- Check if update-alternatives exist `update-alternatives --list iptables`
  - If not, create alternatives for `/usr/sbin/iptables-legacy` and `/usr/sbin/iptable-nft`
  - Repeat for `iptables-<ver>-restore` and `iptables-<ver>-save`, but set these up as slaves of the `iptables` variants (for more details see docs for update-alternatives). This will allow all commands up update when switched over.
- Disable the ip_tables and ip6_tables kernel modules
  - Modify `/etc/modules.d/blacklist.conf`
  - Add instruction `blacklist ip_tables` to the end of the file
  - Add instruction `blacklist ip6_tables` to the end of the file
- Reboot the system for kernel module config to be applied.

I'm currently testing this on my own home server (acting as primary router) and it seems to be working with the above setup plus the code changes. The "Apply Configuration" button is calling the correct commands and the updates are immediately effective.

Without disabling the ip_tables modules, the kernel seems to prefer this over nftables and it'll always take priority. Once the modules are blacklisted, the legacy iptables can be flushed and you only rely on nftables.

jcameron:
The use of `cd` is intentional, in case Webmin is being run from an NFS-mounted directory which could be broken by a firewall rule update.

jcameron:
Thanks for the patch!

PR #2072, opened 2024-01-03
Fixing Webmin Terminal WebSocket connection error for Reverse Proxy use cases

LeiCraft:
I am using the Webmin panel with its own domain (webmin.mydomain.tld) instead of accessing it through mydomain.tld:10000. However, I have encountered an issue where the terminal stops working. Upon further investigation, I observed that the terminal stops functioning when the hostname of the referer is different from the http_host. This behavior is typical when using a reverse proxy.

### DIFF
--- a/xterm/index.cgi
+++ b/xterm/index.cgi
@@ -206,7 +206,27 @@ if ($http_host_conf) {
 			}
 		$http_host_conf =~ s/[\/]+$//g;
 		}
-my $http_host = $http_host_conf || "$ws_proto://$ENV{'HTTP_HOST'}";
+
+# Get the hostname from the HTTP_REFERER
+my $referer_host;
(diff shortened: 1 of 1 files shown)


LeiCraft:
I am using the Webmin panel with its own domain (webmin.mydomain.tld) instead of accessing it through mydomain.tld:10000. However, I have encountered an issue where the terminal stops working. Upon further investigation, I observed that the terminal stops functioning when the hostname of the referer is different from the http_host. This behavior is typical when using a reverse proxy.

iliajie:
> This behavior is typical when using a reverse proxy.

Could you share your reverse proxy configuration? Is it Nginx?

LeiCraft:
The problem no longer occurs for me after one of the last updates

Issue #2102, opened 2024-03-08
Unhide Reboot Now


SlowFox72:
Maybe I'm overlooking something, but I do not get the message Reboot Now after kernelupdates.
I while ago I have hid the Reboot Now Alert, but I would like it back if possible.

Thanks in advance....

jcameron:
There is still a message on the dashboard if you need to reboot after a kernel upgrade, assuming that Webmin can detect it.

Which Linux distribution are you running there?

SlowFox72:
At the moment I'm running : Ubuntu Linux 22.04.4 LTS
and I notices when there was a Kernel Update from Linux 5.15.0-97-generic to Linux 5.15.0-100-generic  that I didn't get the "Reboot Now" option.

Thank in advance

jcameron:
Does the file `/var/run/reboot-required` exist on your system?

Also, does `/var/webmin/modules/webmin/postpone-reboot-required` exist?

SlowFox72:
Hi @jcameron 

I have checked in both folders `reboot-required` doesn't exist neither does `postpone-reboot-required' 

I have rebooted the server after the Kernel Update.

iliajie:
The `/var/run/reboot-required` file in Ubuntu is typically created by package post-installation scripts when a package update requires a reboot. I am not sure how we can help that.

Though, what is the output on your Ubuntu system of the following command:

```
needrestart
```

... and right afterwards, without running anything else, what is output of this command:

```
echo $?
```

SlowFox72:
Here are the results:

![image](https://github.com/webmin/webmin/assets/122017906/9a53a375-f253-4c3f-8398-20dca1691f01)

iliajie:
The command should have been `$?`. Try running it again - important to run one command right after another:

```
needrestart
echo $?
```

SlowFox72:
Sorry for the typo
![image](https://github.com/webmin/webmin/assets/122017906/397ceeea-b073-474a-b87f-e04358879a62)

jcameron:
Looks like it Ubuntu doesn't think a restart is needed?

iliajie:
My Ubuntu 22.04 creates `/var/run/reboot-required` file which we check. But we also could utilize `needrestart`, however I don't see it return correct exit code, i.e. even though a restart is needed it still returns `0`. Still testing though ..

SlowFox72:
Just to add : 

[...]

PR #2103, opened 2024-03-09
Fix to consider using `needrestart` to test if reboot is required

iliajie:
This PR address the issue in this issue report https://github.com/webmin/webmin/issues/2102.


jcameron:
Unless `needreboot` is doing some other check internally?

PR #2111, opened 2024-03-31
Capitalize SMTP auth command for broader compatibility.

mpkossen:
Some SMTP servers only accept a capitalized auth command. This patch capitalized the command before it is sent.

### DIFF
--- a/mailboxes/boxes-lib.pl
+++ b/mailboxes/boxes-lib.pl
@@ -965,7 +965,7 @@ elsif ($sm) {
 						'pass' => $pass } );
 		&error("Failed to create Authen::SASL object") if (!$sasl);
 		local $conn = $sasl->client_new("smtp", &get_system_hostname());
-		local $arv = &smtp_command($h, uc "auth $auth\r\n", 1);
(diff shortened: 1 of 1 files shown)


mailboxes/boxes-lib.pl
```diff
@@ -965,7 +965,7 @@ sub send_mail
 						'pass' => $pass } );
 		&error("Failed to create Authen::SASL object") if (!$sasl);
 		local $conn = $sasl->client_new("smtp", &get_system_hostname());
-		local $arv = &smtp_command($h, "auth $auth\r\n", 1);
+		local $arv = &smtp_command($h, uc "auth $auth\r\n", 1);
```
jcameron, line 968: Better to write this as : 

```
local $arv = &smtp_command($h, "AUTH ".uc($auth)."\r\n", 1);
```

mpkossen: Thanks for the feedback. Changed in the next commit.

jcameron:
Thanks for the fix!

PR #2117, opened 2024-04-09
Add EOL API

iliajie:
Hey Jamie,

As discussed [here](https://forum.virtualmin.com/t/indicator-for-eol-of-os-in-virtualmin-dashboard/125856/11?u=ilia) I created an API to return an object of EOL related data for the current system.

Returned EOL object looks like this:

```
{
  '_eol' => {
    'daymonth' => 'Wednesday, June 10',
    'month' => 'June',
    'short' => '06/10/2026',
    'timestamp' => 1781038800,
    'year' => '2026'
  },
  '_eol_in' => {
    'days' => 791,
    'months' => 25,
    'weeks' => 113,
    'years' => 2
  },
  '_eol_sec' => {
    'daymonth' => 'Saturday, June 10',
    'month' => 'June',
    'short' => '06/10/2028',
    'timestamp' => 1844197200,
    'year' => '2028'
  },
  '_eol_sec_in' => {
    'days' => 1522,
    'months' => 50,
    'weeks' => 217,
    'years' => 4
  },
  '_os' => 'Debian Linux',
  'codename' => 'Bookworm',
  'cycle' => '12',
  'eol' => '2026-06-10',
  'extendedSupport' => '2028-06-10',
  'latest' => '12.5',
  'latestReleaseDate' => '2024-02-10',
  'link' => 'https://www.debian.org/News/2023/20230610',
  'lts' => 1,
  'releaseDate' => '2023-06-10'
}
```

We can later use it in alert message, and table row to be show on the dashboard depending on configurable options.

The EOL data is cached for 5 days by default.

Have a look please!


webmin/os-eol-lib.pl
```diff
@@ -0,0 +1,176 @@
+# os-eol-lib.pl
+# Functions for managing OS end-of-life data
+
+use Time::Local;
+
+# eol_oses_list()
+# Returns a list of OSes for which EOL data is available
+sub eol_oses_list
+{
+return ('almalinux', 'amazon-linux', 'centos-stream', 'centos',
```
jcameron, line 10: Are these the OS names used by the EOL service we're calling?

webmin/os-eol-lib.pl
```diff
@@ -0,0 +1,176 @@
...
+sub eol_fetch_os_data
+{
+my $os = &eol_get_os();
+return undef if (!$os);
+my ($fetch, $error);
+my $eol_write_cache = sub {
+        my $data = shift;
+        &write_file_contents("$module_var_directory/eolcache", $data);
+};
+&http_download('endoflife.date', 443, "/api/$os.json", \$fetch, \$error, undef, 1);
```
jcameron, line 42: Add a timeout to this call

webmin/os-eol-lib.pl
```diff
@@ -0,0 +1,176 @@
...
+if ($error) {
+        &error_stderr("Could not fetch current OS EOL data: " . $error);
+        $eol_write_cache->('[]');
```
jcameron, line 45: Why clear out the cache if EOL data can't be read? Wouldn't it be better to preserve the cache file?

webmin/os-eol-lib.pl
```diff
@@ -0,0 +1,176 @@
...
+        return undef;
+        }
+my $fetch_json;
+eval { $fetch_json = &convert_from_json($fetch); };
+if ($@) {
+        &error_stderr("Could not parse fetched OS EOL data: $@");
+        $eol_write_cache->('[]');
```
jcameron, line 52: Same here

webmin/os-eol-lib.pl
```diff
@@ -0,0 +1,176 @@
...
+
+# eol_get_os_data()
+# Returns the current OS EOL data from the cache or fetches it
+# if it is not available
+sub eol_get_os_data
+{
+my $eol_cache_file = "$module_var_directory/eolcache";
+my $eol_cache = &read_file_contents($eol_cache_file);
+if ($eol_cache) {
+        # Check if the cache is still valid (5 days)
```
jcameron, line 67: Seems like there's no need to ever not use the cache, since EOL data changes really slowly

webmin/os-eol-lib.pl
```diff
@@ -0,0 +1,176 @@
...
+$month -= 1;
+my $eol_timestamp = timelocal(0, 0, 0, $day, $month, $year);
+my $eol_date = &make_date($eol_timestamp, { '_' => 1 });
+$eol_data->{'_eol'} =
+        { daymonth => $eol_date->{'complete_short'},
+          month => $eol_date->{'monthfull'},
+          year => $eol_date->{'year'},
+          short => $eol_date->{'short'},
+          timestamp => $eol_timestamp };
+$eol_data->{'_eol_in'} =
```
jcameron, line 122: Why not just return a timestamp and have the caller of this API format it as they want?

webmin/os-eol-lib.pl
iliajie: Yes.

webmin/os-eol-lib.pl
iliajie: The default timeout is 60 seconds. I thought it's fine?

webmin/os-eol-lib.pl
iliajie: It caches the file, so it will only try again in 5 days.

webmin/os-eol-lib.pl
iliajie: We can increase it to 30 days.

webmin/os-eol-lib.pl
iliajie: It's better call it once and make object just ready.

webmin/os-eol-lib.pl
iliajie: Okay, 5 seconds is more than enough.

webmin/os-eol-lib.pl
iliajie: Yes, fixed! It will preserve the file with data if it exists, otherwise creates an empty one.

iliajie:
Alright, all fixed!

webmin/os-eol-lib.pl
jcameron: Why not forever though? Presumably EOL dates don't get any earlier.

webmin/os-eol-lib.pl
jcameron: Yeah 5 sewconds is better

webmin/os-eol-lib.pl
jcameron: An internal API like this shouldn't contain a lot of presentation logic - it should be up to the caller to format the dates for display.

webmin/os-eol-lib.pl
iliajie: I have completely reworked how EOL data is fetched and stored. We will simply build EOL data upon Webmin build time in `makedist.pl` file, and distribute it with Webmin package. It shouldn't really happen on user side, as Webmin updates happen more often than EOL changes (if changes at all).

webmin/os-eol-lib.pl
iliajie: Agreed! I have factored this logic out. Nevertheless, there is still a lot to do to prevent multiple callers from having to reimplement this logic - there has to be a dedicated subroutine for this.

iliajie:
Jamie, all done this time! Please have another look!

makedist.pl
```diff
@@ -84,8 +88,17 @@
 		@mlist = @mlist_excluded;
 		}
 	}
-@dirlist = ( "vendor_perl" );
 
+# Build EOL data
```
jcameron, line 92: What's this code doing here? I don't know if it's a good idea to have makedist.pl download the EOL file..

webmin/os-eol.map
```diff
@@ -0,0 +1,1810 @@
+=pod
```
jcameron, line 1: Is this just the file that the API provides?

makedist.pl
iliajie: This code builds/updates `os_eol.json` upon Webmin build. Otherwise, we'd have to remember to do it manually.

webmin/os-eol.map
iliajie: The API doesn't provide this. I made it as a reference to remember what the data structures are like and what Webmin thinks about OSes.

makedist.pl
jcameron: But isn't this file just downloaded from the API we call?

iliajie: Yes, I gave up on the idea of updating this file on the user side. Instead, we will download it at Webmin build time and locally provide this file for Webmin users.

webmin/os-eol.map
jcameron: So is this file used by any Webmin code? I'm confused as to why we need to check it in ..

iliajie: No, this is just a .map file. It won't go to the package itself. I checked it in for the reference!

makedist.pl
jcameron: Ah that makes sense. But in that case, do we still need all the code that in os-eol-lib.pl that checks for cache times?

iliajie: Yes, we do. It's new code by the way, and it's crucial so we don't read and parse 31 kB `os_eol.json` file everytime! It makes things much more efficient.

webmin/os-eol.map
jcameron: I think we should remove it, as it's kind of confusing.

iliajie: Alright, it's done. I just wanted to make it easier for future support.

makedist.pl
jcameron: So is that gets included in the package just the JSON, or does it get parsed into some other format first before packaging?

iliajie: No, it will be just a single `os_eol.json` file that we will include into Webmin build and update it everytime new Webmin is out.

jcameron: What is the `eol_update_cache` function in this PR needed for then?

iliajie: To extract a few bytes of EOL-related data for the _current_ OS, rather than dealing with hundreds of systems (from `os_eol.json` file) all the time.

webmin/os-eol-lib.pl
```diff
@@ -0,0 +1,253 @@
...
+                $updated++;
+                }
+        else {
+                # Nothing to update, Webmin config has it all
+                return;
+                }
+        }
+&get_miniserv_config(\%miniserv);
+if (!$gconfig{'os_eol_none'} ||
+    ($miniserv{'server'} && $gconfig{'os_eol_none'} ne $miniserv{'server'})) {
```
jcameron, line 213: What is this checking the miniserv config for?

makedist.pl
jcameron: What I mean is, why is there a need to check any cache time here : 

image

webmin/os-eol-lib.pl
iliajie: It checks if `os_eol.json` was updated/changed after Webmin upgrade. It's useful in case we add or remove some EOL data for an OS.

makedist.pl
iliajie: > What I mean is, why is there a need to check any cache time here

Do you mean why not just do it once, after Webmin was upgraded?

iliajie: Actually, I [just tweaked](https://github.com/webmin/webmin/commit/0b21cbdc7c7a2f32c13552b63e8efb1f34823d7a) it a bit. Considering that the EOL database can only be changed during a Webmin update, we don't need to perform checks every 30 days!

jcameron: Thanks, that looks more reasonable!

webmin/os-eol-lib.pl
```diff
@@ -0,0 +1,230 @@
...
+}
+
+# eol_update_cache()
+# Caches the use of EOL data for next 30 days
+# or until the next Webmin update
+sub eol_update_cache
+{
+my %miniserv;
+&get_miniserv_config(\%miniserv);
+if ($gconfig{'os_eol_db'} ne $miniserv{'server'}) {
```
jcameron, line 196: What does the `miniserv{'server'}` field contain?

iliajie: Current Webmin version.

iliajie: Though it also contains `Miniserv/` string but that's not important as long as it's different.

jcameron: Oh .. better to use the get_webmin_version function then

iliajie:
Yeah, I know. But it's much more simple to compare miniserv server version, as it's always getting updated on Webmin upgrade.

webmin/os-eol-lib.pl
jcameron: That's really best-effort ... we have a function to get the version, so we should just use it

makedist.pl
```diff
@@ -84,8 +88,17 @@
...
+if (-r "$root_directory/webmin/os-eol-lib.pl") {
+	print "Building OS EOL data\n";
+	do "$root_directory/web-lib-funcs.pl";
+	do "$root_directory/webmin/os-eol-lib.pl";
+	&eol_build_all_os_data("$root_directory/os_eol.json");
```
jcameron, line 97: What does this function do other than just downloading the JSON file? Could we replace it with just a `wget` command?

iliajie: Not really. We need to parse JSON as well.

But we can remove it and run it manually or in pre-build scripts.

jcameron: Does the JSON get converted into some other format? It wasn't clear to me what that function is doing..

iliajie: > Does the JSON get converted into some other format? It wasn't clear to me what that function is doing..

It's converted to the JSON string.

jcameron: But isn't what's downloaded from the API already in JSON format?

iliajie: > But isn't what's downloaded from the API already in JSON format?

It is, but we need to combine it for multiple different OSes.

jcameron: Ah, I see ... ok that makes sense.

webmin/os-eol-lib.pl
jcameron: Ok, that makes sense

webmin/os-eol-lib.pl
```diff
@@ -0,0 +1,229 @@
...
+                                                $eol_data->{'_ext_eol_in'} :
+                                        $text{'os_eol_reaching2'};
+                        }
+                }
+        }
+return $eol_data;
+}
+
+# eol_update_cache()
+# Caches the use of EOL data for next 30 days
```
jcameron, line 190: It's not 30 days anymore is it?

iliajie: > It's not 30 days anymore is it?

It isn't, just an old comment. Fixed now.

webmin/os-eol-lib.pl
```diff
@@ -0,0 +1,228 @@
...
+# Updates given EOL hash reference with
+# human readable date and the time ago
+sub eol_populate_dates
+{
+my ($eol_data) = @_;
+if (!$eol_data->{'_eol_timestamp'}) {
+        &error_stderr("The provided data is not a valid EOL data hash reference");
+        return undef;
+        }
+my $eol_date = &make_date($eol_data->{'_eol_timestamp'}, { '_' => 1 });
```
jcameron, line 122: Just curious, what does the `_` option mean?

iliajie: This is a pseudo option to tell `make_date` return an object.

PR #2130, opened 2024-04-17
Fix to correctly disable manual upgrades

iliajie:
This is the issue described in [this thread](https://forum.virtualmin.com/t/should-this-warning-be-displayed/126011) and a couple of others.

Essentially, it disables an option in the Webmin and Usermin configuration pages to perform self-upgrades.

### DIFF
--- a/webmin/index.cgi
+++ b/webmin/index.cgi
@@ -61,6 +61,13 @@ push(@wlinks, "edit_ssl.cgi", "edit_ca.cgi");
 push(@wtitles, $text{'ssl_title'}, $text{'ca_title'});
 push(@wicons, "images/ssl.gif", "images/ca.gif");
 
+# Enfoce to disable Webmin updates
+if ($gconfig{'noselfwebminup'}) {
+	$access{'disallow'} ||= '';
+	$access{'disallow'} = join(' ',
+		split(/\s+/, $access{'disallow'}), 'upgrade');
+	}
+
 # Hide dis-allowed pages
 my %allow = map { $_, 1 } split(/\s+/, $access{'allow'});
 my %disallow = map { $_, 1 } split(/\s+/, $access{'disallow'});
--- a/webmin/index.cgi
+++ b/webmin/index.cgi
@@ -61,16 +61,16 @@ push(@wlinks, "edit_ssl.cgi", "edit_ca.cgi");
 push(@wtitles, $text{'ssl_title'}, $text{'ca_title'});
 push(@wicons, "images/ssl.gif", "images/ca.gif");
 
-# Enfoce to disable Webmin updates
-if ($gconfig{'noselfwebminup'}) {
-	$access{'disallow'} ||= '';
-	$access{'disallow'} = join(' ',
-		split(/\s+/, $access{'disallow'}), 'upgrade');
-	}
-
 # Hide dis-allowed pages
 my %allow = map { $_, 1 } split(/\s+/, $access{'allow'});
 my %disallow = map { $_, 1 } split(/\s+/, $access{'disallow'});
+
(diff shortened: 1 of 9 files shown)


jcameron:
We don't want this for regular Webmin installs, because the user could have just installed from a manually downloaded RPM and not setup a repo at all.

iliajie:
We could automatically install Webmin repo in case no Virtualmin repos present? Probably not, though.

Or we could suggest setup Webmin repo on the dashboard. I think this is the best.

But either way, we should steer away from such upgrades, and have package manager handle it, at least for `rpm` and `deb`.

Furthermore, we could test if Webmin or Virtualmin repos present and if not, show internal upgrade button.

jcameron:
The package manager does still handle it though, by directly downloading the RPM and installing it with the `rpm` command.

Best thing we could do is check if the YUM or APT repo is setup and then disable the internal upgrade button.

iliajie:
> Best thing we could do is check if the YUM or APT repo is setup and then disable the internal upgrade button.

Sure, I will take another look tomorrow.

iliajie:
Alright, this has been fixed on the latest commit.

webmin/webmin-lib.pl
```diff
@@ -108,6 +108,39 @@ =head1 webmin-lib.pl
...
+available for Webmin and Usermin updates.
+
+=cut
+sub has_repos
+{
+my ($force) = @_;
+my $has_repos = 0;
+if (-d '/etc/apt' || -d '/etc/yum.repos.d') {
+	my $out = &backquote_command(
+		"grep -irE 'webmin\.com|webmin\.dev|virtualmin\.com|virtualmin\.dev' ".
```
jcameron, line 127: Instead of shelling out to the grep command, it's more efficient to do the reading and parsing of these files in Perl

webmin/index.cgi
```diff
@@ -61,6 +61,13 @@ push(@wlinks, "edit_ssl.cgi", "edit_ca.cgi");
 push(@wtitles, $text{'ssl_title'}, $text{'ca_title'});
 push(@wicons, "images/ssl.gif", "images/ca.gif");
 
+# Enfoce to disable Webmin updates
+if ($gconfig{'noselfwebminup'}) {
```
jcameron, line 65: Instead of changing %access, better to change the %disallow hash below

webmin/webmin-lib.pl
iliajie: > Instead of shelling out to the grep command, it's more efficient to do the reading and parsing of these files in Perl

I know but sadly it's way to much code. Though here is the patch https://github.com/webmin/webmin/commit/575c1ca5a3c90049a48aacfe389994dc6305b96a.

webmin/index.cgi
iliajie: Fixed too.

webmin/webmin-lib.pl
```diff
@@ -108,6 +108,68 @@ =head1 webmin-lib.pl
...
+		}
+	elsif (-f $path) {
+		# It's a file
+		$process_file->($path);
+		}
+	}
+# Store the result in the config
+if ($force || !defined($gconfig{'noselfwebminup'}) ||
+    $gconfig{'noselfwebminup'} ne $has_repos) {
+	$gconfig{'noselfwebminup'} = $has_repos;
```
jcameron, line 165: Swap this line and the next one

iliajie: > Swap this line and the next one

Thanks! Good suggestion - done!

jcameron:
Thanks for the patch!

iliajie:
Don't mention it!

PR #2131, opened 2024-04-18
Add ED25519 & ED448 DNSSEC Key Algorithm

Romanitch:
No support of **ED25519** & **ED448** DNSSEC Key Algorithm for new bind versions (like 9.18).
ED25519 is a fairly popular algorithm that requires attention from project developers.
I successfully tested the changes on:

Debian Linux 12 (Linux 6.1.0-20-amd64 on x86_64)
BIND version 9.18

 I think integration into the code needs some work and verification by the project leaders.


jcameron:
Thanks for this patch! It looks good ...

PR #2168, opened 2024-05-24
Add support for blocking a given IP temporarily or permanently

iliajie:
G'day Jamie!

Many users have been saying that they miss a feature to manually block a random IP address in FirewallD. This PR adds that functionality.

https://forum.virtualmin.com/t/configserver-csf/126902/15

image


iliajie:
I updated the layout a bit:

image

firewalld/index.cgi
```diff
@@ -148,6 +148,15 @@ if ($ok) {
 		print &ui_form_end();
 		}
 
+	# Block given IP
+	print "<br>".&ui_form_start("block_ip.cgi");
+	print &ui_hidden("zone", $zone->{'name'});
+	print &ui_submit($text{'index_blockip_go'}, "block"),
+	      &ui_textbox("ip", undef, 21, undef, undef,
+	      	"placeholder='1.2.3.4 $text{'or'} 2001:db8::1/64'")."&nbsp;".
```
jcameron, line 156: Better to make this a string with substitutions, because we can't assume that all languages put the work "Or" in the same place

iliajie: Good suggestion, fixed on the latest commit.

jcameron:
Looks good, thanks!

Issue #2169, opened 2024-05-25
System Logs Viewer - Some improvements


shoulders:
|  SYSTEM INFORMATION||
|----------------------|---------------------------|
|  OS type and version |     Ubuntu Linux 22.04.4     |
|  Webmin version      |     2.111      |
|  Usermin version     |     2.010     |
|  Virtualmin version  |     7.10.0       |
|  Theme version       |     21.10      |
|  Package updates     |     All installed packages are up to date   |

### background

- On my dev system I got use to looking at the mail.err log file easily just to see mail related stuff and I could do this because I had `rsyslog` installed which used the webmin module `System Logs`
- Ubuntu seem to be moving to `journald` from `rsyslog` and non-longer installs by default `rsyslog` in favour of going forward with `journald`

### the issue

the `System Logs Viewer` is missing some feature to make it a very handy utility with as much functionality as `System Logs`

- Ability to add own journals - current you can only add log files.
- Add some premade/default `journald` logs for the mails systems
  - journalctl -t postfix/smtpd -t postfix/smtp
  - journalctl -t postfix/smtp 
  - journalctl -t postfix/smtpd
  - journalctl -t dovecot
- Ability to add a service filter (maybe this is optional if you can add your own `journald log commands`)
- On the current ones show the full command when you edit it. Currently they are concatenated an when you edit them you they are still concatenated 
  ![image](https://github.com/webmin/webmin/assets/319997/5937fe70-98e8-41cf-b0dd-a76891ec2561)
  ![image](https://github.com/webmin/webmin/assets/319997/5780ca3c-fd0f-46a5-b163-059b3cbc8296)

### Additional

@stigjoergensen asked for the ability to add custom logs into this webmin modue in #2074 

The `-f` command could just be removed by webmin if someone tries to use a command with it. Maybe a note somewhere that `-f` is not currently supported.

### References
- https://askubuntu.com/questions/1508910/is-it-expected-to-not-have-rsyslog-installed-after-a-fresh-install-of-ubuntu-s
- https://forum.virtualmin.com/t/system-logs-viewer-missing-logs/126913/12

iliajie:
Hello,

> * Add some premade/default `journald` logs for the mails systems

I like this idea, I will look into adding it!

iliajie:
Implemented here https://github.com/webmin/webmin/pull/2170.

Issue #2074, opened 2024-01-12
Suggestion: System log viewer


stigjoergensen:
A way to let people add new commands to interact with the journal... 

Like for the DHCP server it would be nice to be able to see the logs executed by this command:
```
journalctl -u isc-dhcp-server -f
```

and for the DNS server that command could be
```
journal -u named -f
```

perhaps i dont need a fancy webinterface, i could just be a txt file located some please that could be appended to the existing set of commands.

iliajie:
Implemented here https://github.com/webmin/webmin/pull/2170.

PR #2170, opened 2024-05-25
Add support for additional units in systemd log viewer

iliajie:
Hey Jamie!

As discussed in #2074 and #2169 this PR adds support for custom `systemd` units. It also adds standard default units for Debian and EL based systems.

To be clear, I didn't reuse `list_systemd_services` as it's much slower for the task we need. Caching is also implemented; cache is cleared upon module config save.

Also, units are sorted better now and have visual icons in description column.

The module config page looks like:

image

The main page looks like:

image
image

Unit logview page looks like this:

image


swelljoe:
I don't really like this user interface. I didn't like it before, when it was just a handful of queries, and I really don't like it when it's dozens of different queries.

What I want to be able to do is use journalctl queries myself, directly in the UI. The simplest is just being able to choose a unit to display, e.g. from a dropdown menu.

I think the ideal journal UI would be a direct view of the most recent lines from the journal (what you get when you type just `journalctl`), and a few query options for selecting a unit (`-u`) and a time range (`--since` and `--until`) and optional search terms. That's the minimum viable product for a usable GUI for journalctl, IMHO. There are other queries one can do, but that is what I do most of the time when I use journalctl. We don't currently provide the ability to do any of that, except search terms. Units and time ranges are the bread-and-butter of spelunking the journal, we should really consider that the baseline.

As an aside, I think I'd like it if we were also participating in the journal with our logs. It makes it easier to ingest into cloud-based logging infrastructure, since everything else on the system is in the journal it's probably already being ingested. You kinda get logging for free if you're running under systemd. If you output logs to STDOUT/STDERR they're automatically ingested into the journal (whether we still need text logs after that I guess depends on whether the Webmin log tools, like the Actions log, get the ability to pull from the journal instead of a text file...probably less urgent).

iliajie:
Thanks for the feedback, Joe! I like all your suggestions.

I will take a closer look and update this PR with progress.

jcameron:
Yeah I agree with Joe that having a large list of all the various servers is hard to scan though..

iliajie:
Hard to scan?

jcameron:
I mean hard to find what you're looking for, especially since a lot of those units aren't servers that produce logs that the admin would be interested in.

iliajie:
Yeah, right! Almost done ..

iliajie:
Hey [Joe](@swelljoe)! It was all fixed according to your suggestion!

To put it simple, the main page of the module will only display the following and there will no longer be  distro depended options for units in module config, e.g.:

image


As you can see there is now an option for **Messages for specific unit** option. When clicked it will display all available units for the system:

image

With ability to filter messages by time (i.e. using `--since`), e.g.:

image

Please have a look and let me know if there is anything else we could add to make it even better?

iliajie:
You may need the latest Webmin nightly build for things to be displayed nicely.

swelljoe:
Getting close to usable. If I'm reading this right, it's doing a forward search by default (since). It's good to have the ability to do since, but a lot of the time (maybe most of the time) you want an active tail (`-fu unit-name` in `journalctl`). It's be ideal if that use case was possible.

I think the UI for the front page is not discoverable. Without you telling/showing me that there's page with query options hidden under `Output from journalctl -u` I would never, literally never, think to look there. That's as hidden and undiscoverable as I can imagine a UI being.

My ideal would be to default to showing the last 20 or 40 (or some short number) lines from the journal (e.g. the output `journalctl -n 40` if scripting or just `journalctl` if interactive), and a query options interface at the top.

What I don't like about the current UI is all the "Output from..." mess. I would be OK with those maybe being in a section at the bottom of the page, like "Common queries" or something, but we need users to understand the journal can be searched. We need that to be obvious. We have users doing all sorts of crazy shit (like switching back to syslog, even without having any clue how that works either, but at least they can find logs they know how to read, and the UI we provide for actual logs if usable...journal UI is not currently really usable)  every day because they don't understand the journal or how to get info out of it, and I think we make it worse with the current UI; it just hides so much and all the usual tactics for working with the journal aren't possible in the UI.

swelljoe:
I would also get rid of "Output from" from those items. Just have a list of "Common Queries" and show the command and description of what it is.

iliajie:
Sounds good - I'd like to have an active tail too! I'll look into adding it.

Perhaps we could completely get rid of the first page and just show latest logs from the start? What do you think Joe? Jamie, what do you also think if we got rid of the index page in its current form and jumped straight to the logs?

shoulders:
Having all of the logs listed when you go to `System Logs viewer` allows the user to use right click on the view button easily opening different logs. I also like this sort of index because you can read the description for what the log actually does.

Obviously I am not an expert but if possible having the ability to add full custom `journlctrl` logs in the config could be a useful function for the more advanced,  The examples below are just some simple ones to show what it would look like.

image

iliajie:
> Having all of the logs listed when you go to `System Logs viewer` allows the user to use right click on the view button easily opening different logs.

That's a very questionable advantage, as you can achieve the same by cloning the current page using the `Ctrl+Alt+D` shortcut or opening a new tab directly from the navigation menu.

> If possible having the ability to add full custom journlctrl logs in the config could be a useful function for the more advanced. The examples below are just some simple ones to show what it would look like.

You won't need any of this once I fully finish reworking the logs viewer. You will either be able to do it using filter while selecting Postfix unit or tags for the unit directly. But, I'll see if adding native tags support would provide any extra benefit—it's definitely more efficient than `grep`  when dealt with `journalctl`, but for now, I think it wouldn't make any practical difference.

Thank you for your suggestions.

iliajie:
Joe, okay, this feature to tail logs in real time was implemented!

Let me show quick screenshot (so far only supported in Framed Theme) of how the module would look by default, after it was opened:

image

As you can see it will go just straight to logs, after clicking menu link, without further ado. Though, it would be possible to use module configuration to revert to the old behavior and display old style index page with buttons first. But I think it's pretty useless!

Next, by default **Since** will be set to **Latest available** which is _default_ `journalctl` behaviour, and also I don't want it to do **Real-time follow** by default.

**Since** select has the following options, as shown in this screenshot:

image

**Real-time follow** option will monitor select unit/type just as `journalctl -f -u unitname` would do. Very handy, e.g.:


https://github.com/webmin/webmin/assets/4426533/c1059eae-2e55-4183-bb3a-772e85ae0b79

The new real-time logs will be added to the bottom or the top depending on module configuration.

The JavaScript code is compatible with all themes and will be checking for a callback `logviewer_progress_update` to update new lines in any non-standard way.

I also fixed numerous of bugs that made a use of `journalctl` ineffective, like using `tac` and `grep` with it was clearly an oversight!

Jamie, have a look, I have finished the work on it. It can be reviewed and merged.

logviewer/config
```diff
@@ -1,3 +1,5 @@
+skip_index=1
 lines=100
-others=1
+others=0
```
jcameron, line 3: Why change this?

logviewer/logviewer-lib.pl
```diff
@@ -27,33 +27,135 @@ sub can_edit_log
 return 0;
 }
 
-# 
-# Returns standard 
+# get_journal_since
+# Returns a list of journalctl commands to get logs since various times,
+# which should correspond with language strings journal_since0,
+# journal_since1, journal_since2, etc.
+sub get_journal_since
```
jcameron, line 34: Why not just pass in the time in seconds or something, and have the caller determine the time range?

logviewer/config
iliajie: I'd like to have only `systemd`-related units displayed, and in the next (other) PR change Webmin to logging completely to journal.

logviewer/logviewer-lib.pl
```diff
@@ -27,33 +27,135 @@ sub can_edit_log
...
+	  'desc' => $text{'journal_journalctl_debug_info'},
+	  'id' => "journal-6", },
+	{ 'cmd' => "journalctl -n $lines -k ",
+	  'desc' => $text{'journal_journalctl_dmesg'},
 	  'id' => "journal-7", } );
+
+# Add more units from config if exists on the system
+my (%ucache, %uread);
+my $units_cache = "$module_config_directory/units.cache";
+&read_file($units_cache, \%ucache);
```
jcameron, line 79: How long is this cache kept for?

logviewer/logviewer-lib.pl
iliajie: Because this is also used as a template to validate user input.

logviewer/logviewer-lib.pl
iliajie: Until module config is resaved.

logviewer/view_log.cgi
```diff
@@ -94,99 +112,169 @@ else {
...
-	else {
-		$got = undef;
-		}
+	print "<i>$text{'view_empty'}</i>\n"
+		if (!$got || $safe_proc_out =~ /-- No entries --/m);
+	print "</pre>\n";
+	}
+# Progressive output
+else {
+	print "<pre id='logdata' data-reversed='$reverse'>";
```
jcameron, line 239: Why do we need this instead of regular Webmin progressive output?

logviewer/logviewer-lib.pl
iliajie: It will also clear this cache, if refresh button in the dashboard clicked. But I'm thinking of add clear button the the UI on the next PR, along side with controlling "hovering".

logviewer/view_log_progress.cgi
```diff
@@ -0,0 +1,47 @@
...
+# view_log_progress.cgi
+# Returns progressive output for some system log
+
+require './logviewer-lib.pl';
+&ReadParse();
+&foreign_require("proc", "proc-lib.pl");
+
+# System log to follow
+my @systemctl_cmds = &get_systemctl_cmds(1);
+my ($log) = grep { $_->{'id'} eq $in{'idx'} } @systemctl_cmds;
```
jcameron, line 11: What if there is no matching log?

logviewer/view_log.cgi
iliajie: Oh, because we need to update specific element, it works much faster and better with SPA themes like Authentic. It's much more simple and straight forward.

logviewer/view_log_progress.cgi
iliajie: It will fail on lines 19-23.

syslog/lang/en
```diff
@@ -1,4 +1,4 @@
-index_title=System Logs
+index_title=System Logs RS
```
jcameron, line 1: What is "RS" ?

iliajie: RS represents Rsyslog in more specific way, e.g.:

image

While more general and widely available `systemd` will use title **System Logs**.

jcameron: Not all syslog versions are rsyslog though

iliajie: Yeah, the other will display as **System Logs NG**.

logviewer/view_log_progress.cgi
jcameron: You should add a check immediately after line 11 for $log being set

logviewer/logviewer-lib.pl
```diff
@@ -138,5 +240,12 @@ sub extra_log_files
 return @rv;
 }
 
+# config_post_save
+# Called after the module's configuration has been saved
+sub config_post_save
```
jcameron, line 245: What has saving the config got to do with the cache?

logviewer/logviewer-lib.pl
jcameron: Why not just keep the cache for only 1 day?

logviewer/logviewer-lib.pl
jcameron: I don't understand?

logviewer/config
jcameron: Webmin isn't going to be able to log to a journal on the many many systems without systemd though. We should be very careful not to tie ourselves to systemd, or assume it is available.

logviewer/view_log_progress.cgi
iliajie: Right, fixed here https://github.com/webmin/webmin/commit/dccc3fb10e3f9dd85f0f14be1e2666bf183d655f.

logviewer/logviewer-lib.pl
iliajie: You're right now it doesn't have much to do with module config, but originally it did, as the extra units were defined there before I deleted it.

I think saving config and updating cache isn't horrible idea. It makes sense. But I will implement separate buttons for this in the new PR later tomorrow.

logviewer/logviewer-lib.pl
iliajie: Because in production system units are rarely changed. We could clear cache after some package was installed using Webmin.

Also, [Joe](@swelljoe) mentioned once that running `systemctl list-units --all` is slow but the fact is that running `list_systemd_services` is slow. The difference in my tests was just 100ms between reading cached file and calling command, where reading cache file was faster.

logviewer/config
iliajie: > Webmin isn't going to be able to log to a journal on the many many systems without systemd though. We should be very careful not to tie ourselves to systemd, or assume it is available.

That's right! Though those other systems can always use System Logs NG or RS .. But if you want, we could make this option system dependent..

logviewer/logviewer-lib.pl
iliajie: Look, `get_journal_since` is the source of truth for **Since** select options, and what it shows in the UI. It is also used to validate user input (i.e. passed with `$in{'since'}`) to make sure nothing arbitrary gets to the command execution..

iliajie:
Okay, I have finished the progressive logs integration with Authentic. I have also fixed some crucial bugs to abort existing connections, ensuring background calls won't hang forever.

I have added nice connection status indicator for progressive logs in the header. The search highlight will also be supported with progressive logs.

The logs will be opened directly after clicking menu button, without doing any extra work:

https://github.com/webmin/webmin/assets/4426533/6a3ffe27-576d-47b9-86c1-d013cbbbef27

logviewer/config
jcameron: Sure ... but coming back to the original question, is this change related to the rest of the PR?

logviewer/logviewer-lib.pl
jcameron: Oh I see now ... this is just returning the options.  But are we displaying these exact strings to the user? What about localization?

logviewer/logviewer-lib.pl
jcameron: Why not instead just have a timeout on the cache lifetime? This is what's done elsewhere in Webmin for slow operations.

logviewer/logviewer-lib.pl
jcameron: A button to clear the cache? That feels like unnecessarily exposing internal details to the user..

logviewer/view_log.cgi
jcameron: Does this mean that the JS will be called every time a log is viewed, or only in progressive mode?

logviewer/config
iliajie: That's right! It should be part of a future change when we first enable logging to the journal. I reverted it.

logviewer/logviewer-lib.pl
iliajie: Ah, I see. I explained it in the comment above:

```perl
# get_journal_since
# Returns a list of journalctl commands to get logs since various times,
# which should correspond with language strings journal_since0,
# journal_since1, journal_since2, etc.
```

logviewer/view_log.cgi
iliajie: No, this is only for progressive mode. The rest of the modes will work as before.

logviewer/logviewer-lib.pl
iliajie: But there is a better way of doing it https://github.com/webmin/webmin/commit/a420c7142faf66313dcf4e2686cf3969a3bdb820.

logviewer/logviewer-lib.pl
iliajie: I don't find it reasonable to do it in this case.

What if we instead cleared this kind of cache on Webmin restart?

logviewer/logviewer-lib.pl
iliajie: Agreed.

syslog/lang/en
jcameron: Let's just leave the `syslog` module as **System Logs**

logviewer/view_log.cgi
jcameron: Ok great!

logviewer/logviewer-lib.pl
jcameron: That might not happen very often though. What's wrong with clearing based on a timeout?

jcameron: Nice, I like that more!

logviewer/logviewer-lib.pl
jcameron: Nice, I like that more

syslog/lang/en
iliajie: Please, let's not. The `syslog` module is being replaced with `journal` and is no longer installed and recommended by many distros we support, in particular with Virtualmin!

**RSyslog** is the only well-supported syslog version out there nowadays.

I think it would be better to classify names of log modules this way. Old users can always change the name to whatever they wish in the Webmin Configuration page, whereas Virtualmin users won't be confused with the  module name that they will actually use.

logviewer/logviewer-lib.pl
iliajie: > That might not happen very often though. What's wrong with clearing based on a timeout?

I don't think it's reasonable in case of _systemd_ units cache.

> Nice, I like that more!

Do you mean to clear it upon Webmin restart? If so, I will add it later to a separate PR as soon as we deal with this one.

logviewer/logviewer-lib.pl
iliajie: > Nice, I like that more

Yeah, it's better. A little bit more complex though.

syslog/lang/en
jcameron: Maybe just call them RSyslog and Syslog NG then?

logviewer/logviewer-lib.pl
```diff
@@ -27,33 +27,139 @@ sub can_edit_log
...
 }
 
-# 
-# Returns standard 
+# get_journal_since
+# Returns a list of journalctl since commands
+sub get_journal_since
+{
+return [
+        { "" => $text{'journal_since0'} },
```
jcameron, line 35: Why not return this as a 2-element array rather than a hash?

iliajie: Already did, and you liked it.

syslog/lang/en
iliajie: I really think it's easier to keep it simple.

jcameron: I don't think anyone is going to know what RS stands for though!

logviewer/logviewer-lib.pl
jcameron: What I mean, return it as : 

```
[ "-f", $text{'journal_since1'} ],
[ "-b", $text{'journal_since2'} ],
["-S '7 days ago'", $text{'journal_since3'} ],
```

logviewer/index.cgi
```diff
@@ -114,29 +112,51 @@ foreach $e (&extra_log_files()) {
...
+if (!@acols) {
+	$print_header->();
+	&ui_print_endpage($text{'index_elogs'});
+	}
+
+# If we jump directly to logs just redirect
+if ($config{'skip_index'} == 1) {
+	my $link;
+	$link = $acols[0]->[2];
+	$link =~ s/.*?href\s*=\s*(["']?)(?<link>[^"'\s>]+).*\1?/$+{link}/g;
```
jcameron, line 140: Extracting the URL from the HTML seems like a bad idea! It would completely fail if the format of the link ever changes..

syslog/lang/en
iliajie: It's somewhat new but expected, as RSyslog uses it as part of the name and in the logo too. It also goes along well with Syslog-ng.

logviewer/logviewer-lib.pl
iliajie: This is what I was trying to say, I already did!

https://github.com/webmin/webmin/blob/ca3362ee849031a8f72b4bf404c68d8517c15f0e/logviewer/logviewer-lib.pl#L30-L47

logviewer/index.cgi
iliajie: Format of the link will never change and it will either be `<a href=link>link</a>` or `<a href='link'>link</a>` or `<a href="link">link</a>` ..

jcameron: Regardless, it seems like bad practice to convert something into HTML and then convert it back again. Instead why not just store the URL in a separate array?

iliajie: I can do that ..

iliajie: > Regardless, it seems like bad practice to convert something into HTML and then convert it back again. Instead why not just store the URL in a separate array?

Fixed!

shoulders:
image

did custom journals ever get implemented I wanted to look at email and as an example run the following command:

NB: I know I can run this from the command line and also the switches below might not be correct for use in your module as I am not a linux guy and I just wanted to use your module to view email logs 😄 

> journalctl -t postfix/smtpd -t postfix/smtp -f
>
> or
>
> journalctl -u postfix -f

image

| SYSTEM INFORMATION ||
|----------|----------|
|  OS type and version |  Ubuntu Linux 22.04.5              |
|  Usermin version     |  2.102                             |
|  Virtualmin version  |  7.20.2 Pro                        |
|  Theme version       |  21.20.7                           |
|  Apache version      |  2.4.52                            |
|  Package updates     |  46 package updates are available  |

iliajie:
> did custom journals ever get implemented I wanted to look at email and as an example run the following command:

Those are for the files.

Can’t you just choose the Postfix service from the dropdown?

image

shoulders:
I do have those. but I was looking for `postfix/smtpd` and `postfix/smtp` as this is what every tutorial that I have come across says to use. The first one is the daemon and the other is the client, easy to understand.

I think you use **-t** service name?  and what I was looking for is **-u** unit name

What is the difference between `postfix.service` and `postfix@-.service` ?

thanks

iliajie:
> I do have those. but I was looking for `postfix/smtpd` and `postfix/smtp` as this is what every tutorial that

Sure, fine. Just use “Filter lines with text” and enter postfix/smtp there…

> What is the difference between `postfix.service` and `postfix@-.service` ?

One is a standard unit file, and the other is a template unit.

PR #2180, opened 2024-05-31
Dev/better-xterm-logging

iliajie:
Hey Jamie!

This PR implements a small improvements to how logging is done in Terminal module.

### DIFF
--- a/xterm/index.cgi
+++ b/xterm/index.cgi
@@ -194,11 +194,10 @@ my $shellserver_cmd = "$module_config_directory/shellserver.pl";
 if (!-r $shellserver_cmd) {
 	&create_wrapper($shellserver_cmd, $module_name, "shellserver.pl");
 	}
-my $tmpdir = &tempname_dir();
 $ENV{'SESSION_ID'} = $main::session_id;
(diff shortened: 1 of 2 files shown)


xterm/index.cgi
```diff
@@ -194,11 +194,10 @@ my $shellserver_cmd = "$module_config_directory/shellserver.pl";
 if (!-r $shellserver_cmd) {
 	&create_wrapper($shellserver_cmd, $module_name, "shellserver.pl");
 	}
-my $tmpdir = &tempname_dir();
 $ENV{'SESSION_ID'} = $main::session_id;
 &system_logged($shellserver_cmd." ".quotemeta($port)." ".quotemeta($user).
 	       ($dir ? " ".quotemeta($dir) : "").
-	       " >$tmpdir/ws-$port.out 2>&1 </dev/null");
+	       " >$module_var_directory/websocket-connection-$port.out 2>&1 </dev/null");
```
jcameron, line 200: Is this change related to logging?

iliajie: Yes, it will write logs to the right place!

jcameron: Ok makes sense!

Issue #2045, opened 2023-11-13
ProFTPd - Add GUI option to enforce TLS/FTPS easily


shoulders:
|  SYSTEM INFORMATION||
|----------------------|---------------------------|
|  OS type and version |     Ubuntu Linux 22.04.3     |
|  Webmin version      |     2.101      |
|  Usermin version     |     2.001     |
|  Virtualmin version  |     7.8.2       |
|  Theme version       |     21.04      |
|  Package updates     |      3  package updates are available   |

I would like the ability to force TLS for FTP from the GUI for ProFTPd. This is useful because as a lot of people use FTP and do not want to use SFTP + chroot for their clients but yet want more security than just using plain text. The TLS system for ProFTPd is enables, so why not give a quick option to enforce it.

### Solution

in **/etc/proftpd/conf.d/virtualmin.com** you have the following section that enables TLS. The hard work is done. :smile:

```html
# Enable TLS
<IfModule !mod_tls.c>
  LoadModule mod_tls.c
</IfModule>

TLSEngine                     on
TLSRequired                   off
TLSRSACertificateFile         /etc/ssl/certs/proftpd.crt
TLSRSACertificateKeyFile      /etc/ssl/private/proftpd.key
TLSOptions                    NoSessionReuseRequired
TLSVerifyClient               off
TLSLog                        /var/log/proftpd/tls.log
<IfModule mod_tls_shmcache.c>
  TLSSessionCache             shm:/file=/var/run/proftpd/sesscache
</IfModule>
```

This would require the following
- In `Webmin -->Servers --> ProFTPD --> Network Options` add a html option for
  - `TLSRequired` on/off radio field
  - `TLSProtocol`  input fiield,  with a suitable value `TLSv1.2 TLSv1.3` that is pulled from the conf file. If needed this field could have an off value where this values is not written to the conf.

The on/off is obvious, making this easy for the server admin (and newbies) just to turn FTPS on quickly. The `TLSProtocol` option is for going forward as some sysadmins might want to disable TLSv1.2 or even just see what protocols are currently enabled.

**For reference:** 
http://www.proftpd.org/docs/howto/TLS.html

jcameron:
Good idea, I'll look into this ...

jcameron:
Ok this has been implemented for inclusion in the next Webmin release..

shoulders:
[...]

PR #2181, opened 2024-05-31
Fix how modules are loaded in ProFTPd

iliajie:
Hello Jamie!

This PR fixes a bug in this related thread https://github.com/webmin/webmin/issues/2045.


jcameron:
Thanks for this patch!

iliajie:
You're welcome!

PR #2182, opened 2024-05-31
Dev/take-out-ws-lib-and-make-it-work-with-themes

iliajie:
Hey Jamie!

I'm currently working on making the theme stats collection use websockets, which is why I made this change.


xterm/xterm-lib.pl
```diff
@@ -4,7 +4,7 @@
 use WebminCore;
 &init_config();
 our %access = &get_module_acl();
-do "$module_root_directory/websockets-lib-funcs.pl";
+do "$root_directory/websockets-lib-funcs.pl";
```
jcameron, line 7: Is websockets-libs-funcs going to be moved up to the Webmin root dir?

iliajie: Yes.

iliajie:
Thanks!

PR #2191, opened 2024-06-04
Improve WebSockets API

iliajie:
Howdy Jamie!

Since we initially created WebSockets solely for the Terminal module, there were quite a few dependencies involved.

Today, I've been working on addressing all these issues.

Additionally, I've added a new function to get the socket URL based on the configuration. The configuration for the WebSocket base port and host has been moved to the Webmin configuration, where it belongs.

Streamlining the cleanup of no longer used WebSockets is also part of this update.

Please take a look.


web-lib-funcs.pl
```diff
@@ -13522,6 +13522,139 @@ sub get_lock_links_dir
...
+    }
+&unlock_file(&get_miniserv_config_file());
+}
+
+# get_miniserv_websockets_modules()
+# Returns a list of modules and themes that use websockets
+sub get_miniserv_websockets_modules
+{
+my @rv;
+foreach (&get_all_module_infos(), &list_themes()) {
```
jcameron, line 13652: Use a proper loop variable instead of $_

cron/cron-lib.pl
```diff
@@ -1605,9 +1605,8 @@ sub cleanup_temp_files
 closedir(DIR);
 
 # Cleanup old websockets
-if (&foreign_check("xterm")) {
-	&foreign_require("xterm");
-	&xterm::cleanup_miniserv_websockets();
+foreach (&get_miniserv_websockets_modules) {
```
jcameron, line 1608: Shouldn't this be `&get_miniserv_websockets_modules()`

web-lib-funcs.pl
iliajie: I had trouble naming this variable nicely, so I decided not to use it at all. Is it really necessary? It's a very simple call.

cron/cron-lib.pl
iliajie: Oh, you're right! Fixed now! Thanks!

web-lib-funcs.pl
iliajie: Okay. Fixed now too.

jcameron: Thanks! I just dislike the magical $_ perl variable

iliajie: Sure! No problem

jcameron: Wait, did you actually fix this? I don't see it in the diff..

iliajie: > Wait, did you actually fix this? I don't see it in the diff..

https://github.com/webmin/webmin/pull/2191/files#diff-bbebbbbee3314a0e1686dfb6d99107e442573a1d68b87102c51fe94f3cc18b8dR13649-R13656

jcameron: Thanks! Sometimes I think github's "Files changed" tab is laggy..

iliajie: No problem!

PR #2192, opened 2024-06-06
Set websocket URL depending on the proxy unless defined in config

iliajie:
Hi Jamie!

This PR will automatically set the WebSocket URL based on headers provided by the proxy.

We generally cannot trust these kinds of headers, but in this case, it's just an attempt to make it work automatically. In fact, it should work in most cases. If a WebSocket host is defined in the Webmin configuration page, it will override automatic configuration.


jcameron:
Thanks for the patch! In what specific situation is this needed though?

iliajie:
> In what specific situation is this needed though?

When Webmin is placed behind a proxy like Apache or Nginx, users previously had to configure it manually for the Terminal to work. Now, it will just work in most cases.

> Thanks for the patch!

Welcome!

Issue #2160, opened 2024-05-17
getting MS TEAMS meeting invites that do not show time and date of meeting


gstlouisgit:
I have clients that use webmin, beautiful UI and great webmail.  I came across a problem, which I believe is known on some webmail apps.  [https://answers.microsoft.com/en-us/msteams/forum/all/microsoft-teams-meeting-invites-not-showing-date/bbf9eadc-1f44-4876-8678-5f985c94950b](url)

If you ensure you are seeing the email in HTML in webmail of the Microsoft Teams Meeting Invite, you can see the links etc, but there is no date and time of the meeting invite.  If I forward this same email in thunderbird let's say, you can see the invite details.  I'm assuming this is because thunderbird picks up the attach ICS file for its details?

Anyway, in webmin I looked into an option to try to get this working and I think there just isn't.  Here is how to email looks like.  I don't know if there is a way in webmin to be able to see this?

![Screen Shot 2024-05-17 at 8 38 40 AM](https://github.com/webmin/webmin/assets/20375753/ee4dfe21-e7bc-496e-bab7-c86bac3bcf0c)

iliajie:
Hello,

> If you ensure you are seeing the email in HTML in webmail of the Microsoft Teams Meeting Invite, you can see the links etc, but there is no date and time of the meeting invite.

This is a bit hard to follow. Could you share an email (original source file) that isn't being displayed as expected and a screenshot of how it should look? You can send it to my email address (which you can deduct by visiting my GitHub user profile), and I'll take a closer look.

gstlouisgit:
Hello @iliajie
I cannot seem to find your email in your profile.  Can you write to my email so I have it?
Have a great weekend!

iliajie:
> I cannot seem to find your email in your profile.

You need to guess my email based on the already provided domain name. It should be very obvious.

chris001:
@gstlouisgit I'd try firstname at webmin dot com

iliajie:
> @gstlouisgit I'd try firstname at webmin dot com

This is not what I suggested. --) However, webmin dot dev should work.

chris001:
Thanks.  firstname at ross dot gdn might also work!

iliajie:
Thanks @gstlouis for contacting me! I have taken a closer look, and one improvement we could make in Webmin and Usermin is to support displaying calendar events directly in the message, if an email has an `.ics` file attached.

iliajie:
[...]

PR #2193, opened 2024-06-08
Add support for embedding calendar events inline in email messages

iliaross:
G'day Jamie!

With this PR I have finished implementation of embedding calendar events into an email message. This is how it looks like, e.g.:

image

Further discussion can be found here https://github.com/webmin/webmin/issues/2160#issuecomment-2156186118.


mailboxes/mailboxes-lib.pl
```diff
@@ -1306,59 +1324,238 @@ sub parse_calendar_file
...
+			    $event{'_obj_dtend_local_time'} =
+			        make_date($event{'dtend_local_timestamp'},
+					{ _ });
+			# Build when local, e.g.:
+			# Tue Jun 04, 2024 04:30 PM – 05:15
+			# PM (Asia/Nicosia +0300)
+			# or
+			# Tue Jun 04, 2024 04:30 PM – Wed Jun 05, 2024 01:15
+			# AM (Asia/Nicosia +0300)
+			$event{'dtwhen_local'} =
```
jcameron, line 1400: There's a lot of indentation here! Can you split this out into a separate function?

mailboxes/mailboxes-lib.pl
```diff
@@ -1306,59 +1324,238 @@ sub parse_calendar_file
...
+			my $dtstart_obj =
+				$event{'_obj_dtstart_time'} =
+					make_date($event{'dtstart_timestamp'},
+						{ tz => $event{'tzid'} });
+			my $dtend_obj =
+				$event{'_obj_dtend_time'} =
+					make_date($event{'dtend_timestamp'},
+					{ tz => $event{'tzid'} });
+			# Build original when
+			if (!$event{'tzid_missing'}) {
```
jcameron, line 1460: Same here, separate function would be much nicer to read

mailboxes/mailboxes-lib.pl
```diff
@@ -1286,14 +1286,32 @@ sub user_read_dbm_file
...
 # Parses an iCalendar file and returns a list of events
 sub parse_calendar_file
 {
 my ($calendar_file) = @_;
-my (@events, %event);
+my (@events, %event, $line);
 eval "use DateTime; use DateTime::TimeZone;";
 return \@events if ($@);
+# Timezone map
+my %timezone_map = (
```
jcameron, line 1298: Do we really need this? Seems like there a lot of timezones that aren't covered

iliajie: Yes, we do need this. This is what ICS is using. Which one aren't covered? Do you have the list?

mailboxes/mailboxes-lib.pl
iliajie: > There's a lot of indentation here! Can you split this out into a separate function?

It would rase complexity, and I'd prefer not to do it, as it fits 80 char width.

iliajie: Okay, actually there was a way to reduce indentation. Fixed now.

mailboxes/mailboxes-lib.pl
jcameron: I mean, surely there are more timezones in the world than this

iliajie: Yes, for sure! Though those are just to cover zones used by ICS.

jcameron: So ICS only uses those small number of timezones? Is this documented somewhere?

iliajie: No, it isn't. Fix on the latest commit.

iliajie: To be clear, zones like `Asia/Nicosia` or `America/Los_Angeles` don't need to be listed here, as they work perfectly fine with the `DateTime::TimeZone` module.

iliajie:
We might want to integrate this into `make_date` later to support human-readable names like "Eastern European Time." Since the module uses IANA time zone database names, passing a human-readable name currently causes a fatal error.

mailboxes/mailboxes-lib.pl
jcameron: Are there two lists of timezones in this PR now? I also see a list in parse_calendar_file

iliajie: > Are there two lists of timezones in this PR now? I also see a list in parse_calendar_file

You're right! Fixed now!

PR #2195, opened 2024-06-12
Fix to allow disabling the enforcement of sudo-capable logins

iliajie:
Hello Jamie!

This PR addresses a problem where sudo-capable users might not always want to use sudo or su to gain root access in Terminal.

### DIFF
--- a/xterm/acl_security.pl
+++ b/xterm/acl_security.pl
@@ -10,6 +10,14 @@ my ($o) = @_;
 print &ui_table_row($text{'acl_user'},
 	&ui_opt_textbox("user", $o->{'user'} eq '*' ? undef : $o->{'user'},
 			20, $text{'acl_sameuser'}));
+
+if ($o->{'user'} eq "root" && $remote_user ne $o->{'user'}) {
+	print &ui_table_row($text{'acl_sudoenforce'},
+		&ui_yesno_radio("sudoenforce", $o->{'sudoenforce'} == 1 ? 1 : 0));
+	}
+else {
+	print &ui_hidden("sudoenforce", $o->{'sudoenforce'});
+	}
 }
 
 sub acl_security_save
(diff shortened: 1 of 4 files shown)


xterm/acl_security.pl
```diff
@@ -10,11 +10,20 @@ sub acl_security_form
 print &ui_table_row($text{'acl_user'},
 	&ui_opt_textbox("user", $o->{'user'} eq '*' ? undef : $o->{'user'},
 			20, $text{'acl_sameuser'}));
+
+if ($o->{'user'} eq "root" && $remote_user ne $o->{'user'}) {
```
jcameron, line 14: We should just always show all fields in the ACL, not make them conditional on other fields.

iliajie: Well, this option has nonsense for a regular user, right?

jcameron: Right, but the `$o->{'user'}` field that controls if this field is visible is also editable on the ACL form. So it the admin changes the `user` field from something else to `root` he won't be able to adjust this option without saving and then re-opening the page.

iliajie: Oh, that's right. What about this fix then https://github.com/webmin/webmin/pull/2195/commits/a12f385a5bc9ba1c2c8dc81fe42126546fdaa18c.

jcameron: Just keep it simple and always show that field ... no need for any if statement

iliajie: Well .. okay!

jcameron:
So just to clarify, by default any user who has logged into Webmin as a sudo-capable root user and opens a terminal will get a shell running as that user rather than root. Unless this option is changed?

iliajie:
> So just to clarify, by default any user who has logged into Webmin as a sudo-capable root user and opens a terminal will get a shell running as that user rather than root. Unless this option is changed?

Yes, exactly like that!

PR #2207, opened 2024-06-27
Add apply patch CLI API

iliaross:
G'day Jamie!

Time after time, users want to apply the fix immediately, so I added this API to simplify the process for the users.

```perl
=pod

=head1 NAME

patch

=head1 DESCRIPTION

Apply a patch to Webmin core or its modules from GitHub or a local file.

=head1 SYNOPSIS

patch module-name patch-url/file

=head1 OPTIONS

=over

=item --help, -h

Give this help list.

Examples of usage:

  Apply a patch to the Package Updates module from GitHub using the commit ID.
		
    - webmin patch package-updates https://github.com/webmin/webmin/commit/fbee8
	
  Apply a patch to Webmin root files (e.g., `web-lib-funcs.pl`) from 
  GitHub using the commit ID.
  
    - webmin patch . https://github.com/webmin/webmin/commit/e6a2bb15b0.patch

  Apply a patch to Virtualmin GPL module from GitHub using commit ID.
  
    - webmin patch virtualmin-gpl \
        https://github.com/virtualmin/virtualmin-gpl/commit/f4433153dbbb017932b9
  
  Apply a patch to Virtualmin Pro module from local file.
  
    - webmin patch virtualmin-pro /root/virtualmin-pro/patches/patch-1.patch

=back

=head1 LICENSE AND COPYRIGHT

 Copyright 2024 Ilia Ross
```

### DIFF
--- a/bin/patch
+++ b/bin/patch
@@ -8,7 +8,6 @@ use 5.010;
 
 use Getopt::Long qw(:config permute pass_through);
 use Pod::Usage;
-use Term::ANSIColor qw(:constants);
 use File::Basename;
 use Cwd qw(cwd);
 
@@ -38,13 +37,13 @@ init_config();
 
 # Check if curl is installed
 if (!has_command('curl')) {
-	print RED, "curl is not installed\n", RESET;
+	print "curl is not installed\n";
 	exit 1;
 	}
 
 # Check if git is installed
 if (!has_command('git')) {
-	print RED, "git is not installed\n", RESET;
+	print "git is not installed\n";
 	exit 1;
 	}
 
@@ -60,7 +59,7 @@ if (!$patch) {
 # Patch check
 if ($patch !~ /^https?:\/\//) {
 	if (!-r $patch) {
-		print RED, "Patch file $patch doesn't exist\n", RESET;
+		print "Patch file $patch doesn't exist\n";
 		exit 1;
 		}
 	}
@@ -83,7 +82,7 @@ if ($patch =~ m{https://(github|gitlab)\.com/[^/]+/([^/]+)/commit/[^/]+}) {
 
 # Check if module exists
 if (!-d "$path/$module") {
-    print RED, "Module $module doesn't exist\n", RESET;
+    print "Module $module doesn't exist\n";
     exit 1;
 }
 
@@ -100,7 +99,7 @@ else {
 # Apply patch using Git
 my $output = `$cmd 2>&1 | git apply --reject --verbose --whitespace=fix 2>&1`;
 if ($output !~ /applied patch.*?cleanly/i) {
-	print YELLOW, "Patch failed: $output\n", RESET;
+	print "Patch failed: $output\n";
 	exit 1;
 	}
 print "Patch applied successfully to:\n";
--- a/bin/patch
+++ b/bin/patch
@@ -106,7 +106,7 @@ Apply a patch to Webmin core or its modules from GitHub or a local file.
 
 =head1 SYNOPSIS
 
-patch module-name patch-url/file
+webmin patch module-name patch-url/file
 
 =head1 OPTIONS
 
--- a/bin/patch
+++ b/bin/patch
@@ -26,27 +26,24 @@ if (!-r "$path/$lib") {
 		}
 }
 
-# Get module and patch
-my $module = $ARGV[2];
-my $patch = $ARGV[3];
+# Check if curl is installed
+if (!`which curl`) {
+	print RED, "curl is not installed\n", RESET;
+	exit 1;
+	}
+
+# Check if git is installed
+if (!`which git`) {
+	print RED, "git is not installed\n", RESET;
+	exit 1;
+	}
+
+# Get patch URL or file
+my $patch = $ARGV[2];
 
 # Params check
-if (!$module || !$patch) {
+if (!$patch) {
     pod2usage(0);
-    exit 1;
-}
-
-# Special modules handling
-if ($module =~ /^virtualmin-(gpl|pro)$/) {
-	if ($module =~ /^virtualmin-pro/) {
-		$module = 'virtual-server/pro';
-		}
-	else {
-		$module = 'virtual-server';
-		}
-	}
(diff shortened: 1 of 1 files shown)


iliajie:
Though I'm thinking now that for the URLs, we can automatically detect the module to apply the patch to. It would be cleaner and easier to use, like:

```
webmin patch https://
```

jcameron:
That's cool, but please name the command something like "webmin-patch" so it doesn't collide with the standard "patch" command.

iliajie:
> Though I'm thinking now that for the URLs, we can automatically detect the module to apply the patch to. It would be cleaner and easier to use, like:
> 
> ```
> webmin patch https://
> ```

That was just implemented. It's now possible to apply the patch using:

```
webmin patch https://github.com/webmin/webmin/commit/e6a2bb15b0.patch
```

or 

```
webmin patch https://github.com/virtualmin/virtualmin-gpl/commit/f4433153d
```

Though, when applying from a file it will require to `cd` to a directory first, e.g.:

```
cd /usr/libexec/webmin/virtual-server/pro && webmin patch /root/virtualmin-pro/patches/patch-1.patch
```

But the most patches will be applied from the URL.

> That's cool, but please name the command something like "webmin-patch" so it doesn't collide with the standard "patch" command.

No, there is no need for this, as `webmin/bin` is not on the path and won't interfere in any way with native `patch` command!

jcameron:
Ok sounds good!

bin/patch
```diff
@@ -0,0 +1,139 @@
...
+my $lib = "web-lib-funcs.pl";
+if (!-r "$path/$lib") {
+	$path = dirname(dirname($0));
+	if (!-r "$path/$lib") {
+		$path = $path = Cwd::realpath('..');
+		}
+}
+
+# Check if curl is installed
+if (!`which curl`) {
```
jcameron, line 30: Can't we use has_command here?

bin/patch
```diff
@@ -0,0 +1,139 @@
...
+	print RED, "curl is not installed\n", RESET;
+	exit 1;
+	}
+
+# Check if git is installed
+if (!`which git`) {
```
jcameron, line 36: Also here

bin/patch
iliajie: No, not really, as the env ins't set.

bin/patch
iliajie: Same here.

bin/patch
jcameron: What env? You mean $PATH? Because `which` also depends on that.

bin/patch
iliajie: I mean that I tried `has_command` in the first place and it didn't work. I assumed this is because of env.

jcameron: You should add some debugging code to check if $PATH is set

iliajie: Well, we could use `has_command` like this https://github.com/webmin/webmin/pull/2207/commits/d89f6411b62ae55e10e7bb48c87be64b1241aa78. But I think it's overhead and doesn't have any benefits?

jcameron: I think it's always better to use our standard APIs wherever possible. For example, that `which` command isn't always available on all operating systems and shells.

iliajie: Okay, is there anything else on this PR?

bin/patch
```diff
@@ -0,0 +1,153 @@
...
+# patch - Apply a patch to Webmin core or its modules from GitHub or a local file
+
+use strict;
+use warnings;
+
+use 5.010;
+
+use Getopt::Long qw(:config permute pass_through);
+use Pod::Usage;
+use Term::ANSIColor qw(:constants);
```
jcameron, line 11: Is this dependency needed?

bin/patch
```diff
@@ -0,0 +1,153 @@
...
+# Check if module exists
+if (!-d "$path/$module") {
+    print RED, "Module $module doesn't exist\n", RESET;
+    exit 1;
+}
+
+# Download command or cat patch file
+my $cmd;
+if ($patch =~ /^https?:\/\//) {
+	$cmd = "curl -s $patch";
```
jcameron, line 92: Use quotemeta on $patch

bin/patch
```diff
@@ -0,0 +1,153 @@
...
+	chdir "$path/$module";
+	}
+else {
+	$cmd = "cat $patch";
```
jcameron, line 96: Also quotemeta here

bin/patch
iliajie: Yeah, we use it everywhere for colors ..

bin/patch
```diff
@@ -0,0 +1,153 @@
...
+
+# Apply patch using Git
+my $output = `$cmd 2>&1 | git apply --reject --verbose --whitespace=fix 2>&1`;
+if ($output !~ /applied patch.*?cleanly/i) {
+	print YELLOW, "Patch failed: $output\n", RESET;
+	exit 1;
+	}
+print "Patch applied successfully to:\n";
+print "  $1\n" while $output =~ /^Applied patch\s+(\S+)/mg;
+system('/etc/webmin/restart');
```
jcameron, line 107: What if the config directory isn't /etc/webmin ?

bin/patch
jcameron: But I see no reference to `constants` in this PR? Also, I strongly prefer that we do not add coloring to the output of commands, as it messes up when they are fed into grep or copy/pasted in some terminals.

bin/patch
iliajie: Fixed.

bin/patch
iliajie: Fixed too.

bin/patch
iliajie: Fixed as well.

iliajie:
Anything else here?

jcameron:
See my comment about `constants`

bin/patch
iliajie: Thanks!

I missed this comment. Alright, I removed `Term::ANSIColor` dependency.

PR #2211, opened 2024-07-02
Add ability to preserve original file ACLs

iliaross:
Hey Jamie!

This PR fixes this bug described here:

https://github.com/webmin/authentic-theme/discussions/1511#discussioncomment-9913902

### DIFF
--- a/web-lib-funcs.pl
+++ b/web-lib-funcs.pl
@@ -10681,9 +10681,8 @@ elsif (defined($main::open_tempfiles{$_[0]})) {
 	my $file_acls;
 	if ($getfacl && $setfacl) {
 		# Set original ACLs
-		$file_acls = &transname();
-		system("$getfacl --absolute-names ".
-			quotemeta($_[0]).">$file_acls");
+		my $qaclfile = quotemeta($_[0]);
+		$file_acls = `$getfacl --absolute-names $qaclfile`;
 		}
 	# Get status info for a file
 	my @st = stat($_[0]);
@@ -10709,7 +10708,9 @@ elsif (defined($main::open_tempfiles{$_[0]})) {
 		}
 	if ($file_acls) {
 		# Set original ACLs
-		system("$setfacl --restore=$file_acls");
+		open(my $pipe, '|-', "$setfacl --restore=-");
+		print($pipe $file_acls);
+		close($pipe);
 		}
 	&reset_file_attributes($_[0], \@old_attributes);
 	delete($main::open_tempfiles{$_[0]});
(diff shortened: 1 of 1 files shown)


web-lib-funcs.pl
```diff
@@ -10676,6 +10676,16 @@ sub close_tempfile
...
 	my $noerror = $main::open_tempfiles_noerror{$_[0]};
 	&webmin_debug_log("CLOSE", $_[0]) if ($gconfig{'debug_what_write'});
+	my $getfacl = &has_command("getfacl");
+	my $setfacl = &has_command("setfacl");
+	my $file_acls;
+	if ($getfacl && $setfacl) {
+		# Set original ACLs
+		$file_acls = &transname();
+		system("$getfacl --absolute-names ".
+			quotemeta($_[0]).">$file_acls");
```
jcameron, line 10686: Can we do this by reading the current ACLs into a variable rather than a temp file?

iliajie: > Can we do this by reading the current ACLs into a variable rather than a temp file?

Yes, why not. Check this https://github.com/webmin/webmin/pull/2211/commits/82f5284ffca200e341111db21c552fd3480de60b patch out.

web-lib-funcs.pl
```diff
@@ -10676,6 +10676,15 @@ sub close_tempfile
 	# Closing a file
 	my $noerror = $main::open_tempfiles_noerror{$_[0]};
 	&webmin_debug_log("CLOSE", $_[0]) if ($gconfig{'debug_what_write'});
+	my $getfacl = &has_command("getfacl");
+	my $setfacl = &has_command("setfacl");
+	my $file_acls;
+	if ($getfacl && $setfacl) {
+		# Set original ACLs
+		my $qaclfile = quotemeta($_[0]);
+		$file_acls = `$getfacl --absolute-names $qaclfile`;
```
jcameron, line 10685: Use backquote_command instead of `

web-lib-funcs.pl
```diff
@@ -10697,6 +10706,12 @@ sub close_tempfile
 		chmod($st[2], $_[0]);
 		chown($st[4], $st[5], $_[0]);
 		}
+	if ($file_acls) {
+		# Set original ACLs
+		open(my $pipe, '|-', "$setfacl --restore=-");
```
jcameron, line 10711: There's an execute_command Webmin function that can be used to feed input from a string to a command

web-lib-funcs.pl
iliajie: Fixed.

web-lib-funcs.pl
iliajie: Yeah, but I won't use it from inside `close_tempfile`, which can be called hundreds or thousands of times during a call. Using a direct pipe avoids the overhead of forking and will be more performant for this specific case, in my opinion.

Issue #2216, opened 2024-07-16
Software Package Updates - Filter by security updates


shoulders:
|  SYSTEM INFORMATION||
|----------------------|---------------------------|
|  OS type and version |     Ubuntu Linux 22.04.4     |
|  Webmin version      |     2.111      |
|  Usermin version     |     2.010     |
|  Virtualmin version  |     7.10.0       |
|  Theme version       |     21.10      |
|  Package updates     |      107  package updates are available   |

I don't know if this is possible 😄 

### The background

on the page `Webmin --> Software Package Updates` you can filter by (installed |Only Updates | Only new)

![image](https://github.com/user-attachments/assets/81117ef2-59f8-4d5d-bd15-b71b942dc710)

### The request

Is it possible to add a further filter option, `Only Security Updates`

This would be useful for people like me who come along when there are hundreds of updates on a live server pending but only a few of them are security updates which `must` be installed. This new option would allow me to filter down to the security updates and concentrate on them rather than a blanket install everything at once.

You could perhaps add another column with the title `Is Security Update` with the options (Yes | No)

### Additional request (Virtualmin)

I can add a separate issue to Virtulmin for this if required, but here are some ideas to improve the virtualmin widget

- Add the ability to show only security updates and a summary of the other updates
- Add a button to update security updates only
    ![image](https://github.com/user-attachments/assets/378e68fc-4889-4ffb-b506-25ec01ffc631)
- Maybe add a column for each update showing weather it is a security update
- Is the install all updates button a safe thing to have here?

### Additional information

https://github.com/webmin/webmin/issues/2075

iliajie:
Hello,

Thanks for the heads up!

> Is it possible to add a further filter option, `Only Security Updates`

We already had this button. However, it was displayed conditionally. I think showing it always is better. Here is the patch https://github.com/webmin/webmin/pull/2217.

> ### Additional request (Virtualmin)
> I can add a separate issue to Virtulmin for this if required, but here are some ideas to improve the virtualmin widget
> 
> * Add the ability to show only security updates and a summary of the other updates
> * Add a button to update security updates only

No, we should keep it simple!

PR #2217, opened 2024-07-16
Fix to always show security updates button

iliajie:
This PR will always show the button to display security package updates only.

Ref.: https://github.com/webmin/webmin/issues/2216


jcameron:
Won't this look odd if there are no security updates?

iliajie:
it won't look any odder then it is now for "Only updates" not being available:

image

PR #2234, opened 2024-07-26
Fix to optimize getting table index structure for large databases

iliajie:
Hey Jamie!

This PR will optimize getting table index structure on the huge databases.

It was tested with TikiWiki database which over 400 tables and indexes.

The performace boost is around 500%.

### DIFF
--- a/mysql/mysql-lib.pl
+++ b/mysql/mysql-lib.pl
@@ -1304,6 +1304,37 @@ return &ui_textbox($name, $value, 8)."\n".
 		    [ "M", "MB" ], [ "G", "GB" ] ]);
 }
 
+# get_table_index_stats(db)
+# Retrieves index stats for all tables in the given database
+sub get_table_index_stats
+{
+my ($db) = @_;
+my $table_list = join(", ", map { "'$_'" } &list_tables($db));
+my $sql_query = "
+    SELECT 
+        TABLE_SCHEMA,
+        TABLE_NAME,
+        INDEX_NAME,
+        NON_UNIQUE,
+        SEQ_IN_INDEX,
+        COLUMN_NAME,
+        COLLATION,
+        CARDINALITY,
+        SUB_PART,
+        PACKED,
+        NULLABLE,
+        INDEX_TYPE,
+        COMMENT,
+        INDEX_COMMENT
+    FROM 
+        INFORMATION_SCHEMA.STATISTICS
+    WHERE 
+        TABLE_SCHEMA = '$db' AND TABLE_NAME IN ($table_list);
+";
+my $rs = &execute_sql_safe($db, $sql_query);
+return $rs;
+}
+
 # list_indexes(db)
 # Returns the names of all indexes in some database
 sub list_indexes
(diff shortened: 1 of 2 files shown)


iliajie:
@marclaporte Thanks for reporting this issue in the first place.

If you have Webmin 2.200+ installed, you can apply this patch to run tests as simply as:

```
webmin patch https://github.com/webmin/webmin/commit/b03f1
```

To be clear, for me, the load time dropped from 9.5 seconds to 1.5 seconds.

mysql/mysql-lib.pl
```diff
@@ -1304,6 +1304,37 @@ sub mysql_size_input
 		    [ "M", "MB" ], [ "G", "GB" ] ]);
 }
 
+# get_table_index_stats(db)
+# Retrieves index stats for all tables in the given database
+sub get_table_index_stats
+{
+my ($db) = @_;
+my $table_list = join(", ", map { "'$_'" } &list_tables($db));
```
jcameron, line 1312: There's a quoting function in mysql-lib.pl you can use instead of doing it yourself

mysql/mysql-lib.pl
```diff
@@ -1304,6 +1304,37 @@ sub mysql_size_input
...
+        SUB_PART,
+        PACKED,
+        NULLABLE,
+        INDEX_TYPE,
+        COMMENT,
+        INDEX_COMMENT
+    FROM 
+        INFORMATION_SCHEMA.STATISTICS
+    WHERE 
+        TABLE_SCHEMA = '$db' AND TABLE_NAME IN ($table_list);
```
jcameron, line 1332: Use a proper SQL parameter instead of substituting $db into the string

iliajie: Yes, sure! Check this patch https://github.com/webmin/webmin/pull/2234/commits/5210f623f3f1a0b8540fa6ac0d2c94689a4f1f8f.

mysql/mysql-lib.pl
iliajie: I don't think we even need this with formal params ..

marclaporte:
> huge databases.
> 
> It was tested with TikiWiki database which over 400 tables and indexes.

Yes, Tiki is good to push limits :-)

All Tiki's features are built-in, so it's big on a fresh install (mostly empty tables) but the number of tables doesn't grow after. It's quite the opposite model to other plugin/module centric systems like WordPress/Drupal/Joomla!/...
- https://tiki.org/FLOSS-Web-Application-with-the-most-built-in-features
- https://pluginproblems.com/

iliajie:
Marc, did you manage to apply the patch and see how much faster it gets for you?

marclaporte:
@iliajie I didn't try yet, but I am sure it's going to be great now!

mysql/mysql-lib.pl
jcameron: Can a table name contain ' ?

jcameron: Oh never mind I see you fixed this

mysql/mysql-lib.pl
jcameron: Excellent thanks!

PR #2241, opened 2024-07-29
[FIX] Add missing `title` parameter to `ui_print_header` method on the documentation

bim-g:
While reading the `ui_print_header` method documentation, I realized that the `title` parameter was missing from the method definition while on parameter list description it was defined, which could lead to confusion and misuse of the method.
image


iliajie:
Thanks for the patch!

PR #2273, opened 2024-09-06
Updated Chinese translations

leangjia:
Updated Chinese translations


iliajie:
Thanks!

iliajie:
I'd like to point out, when making translations, it would be reasonable to skip translating all obsolete modules. The current modules distributed with Webmin are listed here:

https://github.com/webmin/webmin/blob/master/mod_def_list.txt

leangjia:
Thanks for your reminder!

PR #2284, opened 2024-09-16
Add support for listing `deb822-style` repos

iliajie:
Hello Jamie!

As we [discussed earlier](https://forum.virtualmin.com/t/package-repositories-empty-ubuntu-24-04/129152/11?u=ilia), here’s the PR to add support for the `.sources` repos format.

Unfortunately, `deb822-style` isn’t very compatible with how we currently delete, enable, and disable repos, as everything revolves around a single line, rather than a block of connected lines. Unless we completely change the logic in `software::list_package_repos`, I've kept it simple for now and just added support for repos listing.

Please have a look.

### DIFF
--- a/software/apt-lib.pl
+++ b/software/apt-lib.pl
@@ -457,7 +457,34 @@ my @rv;
 foreach my $f ($sources_list_file, glob("$sources_list_dir/*")) {
 	my $lref = &read_file_lines($f, 1);
 	my $lnum = 0;
+	my (%repo, @types);
+	my $repo = sub {
+		foreach my $type (@types) {
+			my @suites = @{$repo{'suites'}};
+			foreach my $suite (@suites) {
+				my @comps = @{$repo{'comps'}};
+				foreach my $comp (@comps) {
+					my $repo =
+					  {
+					    'cannot' => 1,
+					    'file' => $f,
+					    # 'line' => $lnum,
+					    'url' => $repo{'url'},
+					    'enabled' => !$repo{'disabled'},
+					    'words' => [$comp, $suite],
+					    'name' => join("/", $comp, $suite).
+					      ($type =~ /src/ ? " ($type)" : ""),
+					    'signed_by' => $repo{'signed_by'},
+					  };
+					$repo->{'id'} =
+						$repo->{'url'}.$repo->{'name'};
+					push(@rv, $repo);
+					}
+				}
+			}
+		};
 	foreach my $l (@$lref) {
+		# Debian classic format (using .list files)
 		if ($l =~ /^(#*)\s*deb.*?((http|https)\S+)\s+(\S.*)/) {
 			my $repo = { 'file' => $f,
 				     'line' => $lnum,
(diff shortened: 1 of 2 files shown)


software/apt-lib.pl
```diff
@@ -457,7 +457,34 @@ sub list_package_repos
 foreach my $f ($sources_list_file, glob("$sources_list_dir/*")) {
 	my $lref = &read_file_lines($f, 1);
 	my $lnum = 0;
+	my (%repo, @types);
+	my $repo = sub {
```
jcameron, line 461: Why use an anonymous sub here? It seems like this is only called once

software/apt-lib.pl
```diff
@@ -457,7 +457,34 @@ sub list_package_repos
...
+		foreach my $type (@types) {
+			my @suites = @{$repo{'suites'}};
+			foreach my $suite (@suites) {
+				my @comps = @{$repo{'comps'}};
+				foreach my $comp (@comps) {
+					my $repo =
+					  {
+					    'cannot' => 1,
+					    'file' => $f,
+					    # 'line' => $lnum,
```
jcameron, line 471: Why comment out this line?

software/apt-lib.pl
```diff
@@ -471,6 +498,40 @@ sub list_package_repos
...
+					}
+				elsif ($key eq 'Signed-By') {
+					$repo{'signed_by'} = $value;
+					}
+				elsif ($key eq 'Enabled') {
+					$repo{'disabled'} =
+						(lc($value) eq 'no') ? 1 : 0;
+					}
+				}
+			elsif ($l =~ /^\s*$/ || $lnum == $#{$lref}) {
```
jcameron, line 525: I think it's better to add the repo to the list when the first line is seen, not when it's the last one or the end of the file.

software/apt-lib.pl
iliajie: It won't fit within the 80-character-per-line limit because there are too many tabs and code.

software/apt-lib.pl
iliajie: I wanted to point out that line numbers won't function as usual here because we are dealing with a block of interconnected values.

software/apt-lib.pl
iliajie: We consider it a separator because the repository consists of a block (stanzas) of interconnected values, each divided by a new line.

https://manpages.debian.org/bookworm/dpkg-dev/deb822.5.en.html

That's why I add the repository only after the entire stanza has been read.

software/apt-lib.pl
jcameron: Ok but then why is it defined inside the loop?

software/apt-lib.pl
jcameron: I don't understand ... is $lnum not valid? Leaving this commented out is very confusing, either set the `line` field properly or not at all if we don't need it.

software/apt-lib.pl
jcameron: Is this still going to work if there's no blank line at the end of the file? Seems like this branch won't be taken if the statement `if ($l =~ /^([\w\-]+):\s*(.+)$/) {` above matches.

software/apt-lib.pl
iliajie: > but then why is it defined inside the loop?

This is because it uses locally scoped variables that are only accessible within the current context of the specific repo file.

software/apt-lib.pl
iliajie: > I don't understand ... is $lnum not valid? Leaving this commented out is very confusing, either set the `line` field properly or not at all if we don't need it.

Yes, I have removed it! I just wanted to point this out!

software/apt-lib.pl
iliajie: Oh, excellent catch! Thanks, I’ve just fixed that in this patch https://github.com/webmin/webmin/pull/2284/commits/c199cffb3d763a6f041437f54b0fd8b45fb8d062.

iliajie:
This is all good and needs to be merged. I nearly forgot about it!

software/apt-lib.pl
jcameron: Currently inside the loop, $repo is used for both this sub and some other variables, which is quite confusing.

iliajie: You're right! Fixed now.

software/apt-lib.pl
```diff
@@ -471,6 +497,40 @@ sub list_package_repos
...
+					}
+				elsif ($key eq 'Enabled') {
+					$repo{'disabled'} =
+						(lc($value) eq 'no') ? 1 : 0;
+					}
+				}
+			if ($l =~ /^\s*$/ || $lnum == $#{$lref}) {
+				# Process and push the current repo if we
+				# got an empty line or it's the last line
+				if (%repo) {
```
jcameron, line 527: Combine this with line 524 above

software/apt-lib.pl
```diff
@@ -457,7 +457,33 @@ sub list_package_repos
 foreach my $f ($sources_list_file, glob("$sources_list_dir/*")) {
 	my $lref = &read_file_lines($f, 1);
 	my $lnum = 0;
+	my (%repo, @types);
+	my $repo_proc = sub {
```
jcameron, line 461: I don't understand why this code needs to go into a sub when it's only called once?

iliajie: Fox readability, as it won't fit otherwise into 80 characters in width.

jcameron: AH ok .. I suppose I would have preferred to split this out into a true separate sub, but I suppose it's OK

PR #2316, opened 2024-11-19
Fix to respect nolog directive

iliaross:
Hey Jamie!

This issue aims to address what we discussed over email.

Essentially, it’s about ensuring the `nolog` directive is _always_ respected. Ref. https://github.com/webmin/authentic-theme/commit/85f1b12f11341577daf862ef9b245b9f55845cc2


jcameron:
So this would block logging of even error messages from no-log URLs? I worry a bit that this would hide attacks from the logs..

iliajie:
> So this would block logging of even error messages from no-log URLs? I worry a bit that this would hide attacks from the logs..

Well, yes, but we set this up intentionally. Plus, those files aren’t accessible to unauthenticated users.

Maybe we could use a different key — `nolog_err` might be a better option, and also keep it separate from `nolog`?

iliajie:
After your comment, I looked into it further, and it seems that _all_ 404 errors are no longer being logged. Now I’m unsure if this is the best approach.

What would you suggest, considering the original report?

jcameron:
Yeah since we could have many users on the system with different themes, we need to handle the case where 404.cgi doesn't exist. Simplest way would be to just revert to the default error handling in miniserv.pl if it's not found.

jcameron:
I'll take a look into it

iliajie:
> I'll take a look into it

Great! Thanks! Please let me know when you have the patch ready...

jcameron:
Ok check out this patch : https://github.com/webmin/webmin/commit/80e99b7b259afa8a91b0635adcf57744f6d06d72

iliajie:
> Ok check out this patch : [80e99b7](https://github.com/webmin/webmin/commit/80e99b7b259afa8a91b0635adcf57744f6d06d72)

Look great!

PR #2318, opened 2024-11-25
Update de

MacSteini:


iliajie:
Hallo,

> @MacSteini
[Update de](https://github.com/webmin/webmin/pull/2318/commits/37d512097704f511220edc6d22088d85ae82af4f)

Thanks for this one!

> @MacSteini
[Files updated by Language Manager](https://github.com/webmin/webmin/pull/2318/commits/9d25a07c123573eb6d9581c5006948d8477625cd)

This one needs to be unrolled; it’s not correct.

iliajie:
> What happened? I followed suit with what you once said in the forum. Aren't we supposed to use the language manager at all?

Of course we should! 🙂 But we also need to pay attention that the actual translation has happened...

iliajie:
> I see, is there any rule of thumb when one can trigger the language manager?

I don’t understand your question, sorry!

What I meant is that the files you submitted weren’t really translated—the English strings were simply copied into the other language files...

iliajie:
The language manager handles string translations. In the future, I plan to add AI support for translations. For now, you’ll need to use the Google API to perform the actual translations.

I suggest rolling back the last commit in this PR so I can merge it...

iliajie:
> So, there's no reliable way to know when EN files got updated?

Every time a new key is added, it will be automatically translated when the script runs. If an existing key in the English file is edited, it’s Jamie’s or my responsibility to handle it by either force-translating the modified key or changing the key name, for example, from `keyname` to `keyname1`.

> Could you please point me to a manual how to use the Google API for this use-case?

I think, it's[ here](https://cloud.google.com/translate/docs/reference/rest/).

> Will revert the PR asap.
>> Rolled back.

Thanks!

PR #2320, opened 2024-11-25
Add check if we have to use a new MariaDB commands

iliaross:
Hey Jamie!

As discussed here:

https://github.com/virtualmin/virtualmin-gpl/issues/968#issuecomment-2496441791

### DIFF
--- a/mysql/mysql-lib.pl
+++ b/mysql/mysql-lib.pl
@@ -76,6 +76,51 @@ use DBI;
 EOF
 }
 
+# Check if we have to use a new MariaDB commands
+if (!$config{'mysql_mariadb_updated'}) {
+	my ($mversion, $mvariant) = &get_remote_mysql_variant();
+	if ($mvariant =~ /mariadb/i) {
+		if (&compare_version_numbers($mversion, "10.5") > 0) {
+			my $config_updated;
+			foreach my $key (grep { /^mysql(?!_)/ } keys %config) {
+				my $cmd = $config{$key};
+				next if ($cmd !~ /^\//);
+				# Check if symlink
+				if (-l $cmd) {
+					my $target = readlink($cmd);
+					my $dir = $cmd =~ s|/[^/]+$||r;
+					if ($target =~ /^mariadb/i) {
+						# Update config if symlinked
+						# to mariadb
+						$config{$key} = "$dir/$target";
+						$config_updated++;
+						}
+					}
+				# If command doesn't exist, try mariadb version
+				else {
+					my $mariadb_cmd = $cmd;
+					$mariadb_cmd =~ s/mysql(\w+)/mariadb-$1/;
+					if (&has_command($mariadb_cmd)) {
+						$config{$key} = $mariadb_cmd;
+						$config_updated++;
+						}
+					}
+				}
+			if ($config_updated) {
+				# Update start stop commands
+				foreach my $key ('start_cmd', 'stop_cmd') {
+					next if ($config{$key} !~ /systemctl/);
+					$config{$key} =~
+						s/mysql(?:\.service)?$/mariadb/;
+					}
+				# Update config file
+				$config{'mysql_mariadb_updated'} = $mversion;
+				&save_module_config();
+				}
+			}
+		}
+	}
+
 # Fix text if we're running MariaDB
 sub fix_mysql_text
 {
--- a/mysql/mysql-lib.pl
+++ b/mysql/mysql-lib.pl
@@ -76,51 +76,6 @@ use DBI;
 EOF
 }
 
-# Check if we have to use a new MariaDB commands
-if (!$config{'mysql_mariadb_updated'}) {
-	my ($mversion, $mvariant) = &get_remote_mysql_variant();
-	if ($mvariant =~ /mariadb/i) {
-		if (&compare_version_numbers($mversion, "10.5") > 0) {
-			my $config_updated;
-			foreach my $key (grep { /^mysql(?!_)/ } keys %config) {
-				my $cmd = $config{$key};
-				next if ($cmd !~ /^\//);
-				# Check if symlink
-				if (-l $cmd) {
-					my $target = readlink($cmd);
-					my $dir = $cmd =~ s|/[^/]+$||r;
-					if ($target =~ /^mariadb/i) {
-						# Update config if symlinked
-						# to mariadb
-						$config{$key} = "$dir/$target";
-						$config_updated++;
-						}
-					}
-				# If command doesn't exist, try mariadb version
-				else {
-					my $mariadb_cmd = $cmd;
-					$mariadb_cmd =~ s/mysql(\w+)/mariadb-$1/;
-					if (&has_command($mariadb_cmd)) {
-						$config{$key} = $mariadb_cmd;
-						$config_updated++;
-						}
-					}
-				}
-			if ($config_updated) {
-				# Update start stop commands
-				foreach my $key ('start_cmd', 'stop_cmd') {
-					next if ($config{$key} !~ /systemctl/);
-					$config{$key} =~
-						s/mysql(?:\.service)?$/mariadb/;
-					}
-				# Update config file
-				$config{'mysql_mariadb_updated'} = $mversion;
-				&save_module_config();
-				}
-			}
-		}
-	}
-
 # Fix text if we're running MariaDB
 sub fix_mysql_text
 {
--- a/mysql/postinstall.pl
+++ b/mysql/postinstall.pl
@@ -8,4 +8,49 @@ if ($mysql_version && $mysql_version >= 0) {
 	&save_mysql_version($mysql_version);
 	&create_module_info_overrides();
 	}
+
+# Check if we have to use a new MariaDB commands
+if (!$config{'mysql_mariadb_updated'}) {
+	my ($mversion, $mvariant) = &get_remote_mysql_variant();
+	if ($mvariant =~ /mariadb/i) {
+		if (&compare_version_numbers($mversion, "10.5") > 0) {
+			my $config_updated;
+			foreach my $key (grep { /^mysql(?!_)/ } keys %config) {
+				my $cmd = $config{$key};
+				next if ($cmd !~ /^\//);
+				# Check if symlink
+				if (-l $cmd) {
+					my $target = readlink($cmd);
+					my $dir = $cmd =~ s|/[^/]+$||r;
+					if ($target =~ /^mariadb/i) {
+						# Update config if symlinked
+						# to mariadb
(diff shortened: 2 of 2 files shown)


jcameron:
Better to do this in `postinstall.pl`, which is guaranteed to be run at install or upgrade time.

But even better, can't we identify the distros where there is no `mysql` command and create appropriate `config-*` files for them? That's the Webmin standard way of using the correct paths by default.

iliajie:
> Better to do this in `postinstall.pl`, which is guaranteed to be run at install or upgrade time.

Great suggestion! Fixed now!

> But even better, can't we identify the distros where there is no mysql command and create appropriate config-* files for them?

No, it doesn’t fit this scenario because the command might still exist, but a specific version will consistently throw warnings. We need to address this before the `mysql` command potentially stops existing.

> That's the Webmin standard way of using the correct paths by default.

Technically, the existing code in this PR already does the same thing—it adjusts the config file to be MariaDB-compatible.

jcameron:
So do you mean that some distributions might have either MySQL or MariaDB installed for the same distro version? Or is the issue that the `mysql` and `mariadb` commands exist, but `mysql` doesn't work?

iliajie:
> So do you mean that some distributions might have either MySQL or MariaDB installed for the same distro version?

Yes, this is too! For sure!

> Or is the issue that the mysql and mariadb commands exist, but mysql doesn't work?

The problem is that currently for systems with MariaDB, both commands exist but at some point it won't. Also, some newer versions of MariaDB started to complain about using `mysql*` programs which are symlinks. So to address that, we will test which database is used, and fix the Webmin MySQL config in advance...

jcameron:
So it looks like in the original bug https://github.com/virtualmin/virtualmin-gpl/issues/968 the user had installed MariaDB from a custom repository, and isn't using the OS packages at all. Honestly I think that if a user does this, it's on them to adjust the paths in the module config, and we shouldn't be try this kind of automatic detection.

The exception is if the distro itself offers two different packages and we need a better way to support both.

iliajie:
> So it looks like in the original bug [virtualmin/virtualmin-gpl#968](https://github.com/virtualmin/virtualmin-gpl/issues/968) the user had installed MariaDB from a custom repository, and isn't using the OS packages at all. Honestly I think that if a user does this, it's on them to adjust the paths in the module config, and we shouldn't be try this kind of automatic detection.
> 
> The exception is if the distro itself offers two different packages and we need a better way to support both.

Not really. There are nearly all contemporary distros that we support have MariaDB 10.5+ which has all those commands available and all old commands that Webmin references in the config are symlinks to a new `mariadb_` command equipments...

iliajie:
⬆️

jcameron:
So the old `mysql` commands still work then? If so, I really don't think we need to change anything.

For example, if we were to re-write them in the config to the `mariadb` commands, this could cause breakage if the user later switches from MariaDB to MySQL. For better or worse, the `mysql` command is the API to all mysql-compatible databases now.

iliajie:
They work for now, but most of our Virtualmin users are on MariaDB, and some later versions have started showing annoying warnings (related to commands deprecation).

Besides, I just don’t think relying on users to manually switch to MySQL is a statistically sound approach.

jcameron:
Oh which distros and versions are showing warnings? Is this showing up in the Webmin UI?

I think the right thing is for distros that have switched to MariaDB by default, to create a new config-XYZ file that uses the correct paths.

iliajie:
> I think the right thing is for distros that have switched to MariaDB by default, to create a new config-XYZ file that uses the correct paths.

There are distort that provide both MySQL and MariaDB. We cannot know what user prefers.

> Oh which distros and versions are showing warnings?

It could be any distro; users can manually upgrade to MariaDB 11.

> Is this showing up in the Webmin UI?

The warning is simply printed to stdout. Whether it appears in the UI depends on how the command is executed.

iliajie:
... this is why I think, it's better to fix it inlace, as we don't have, and shouldn't have a separate MariaDB module.

jcameron:
I'm going to have to spin up a VM with a distribution that's showing this issue to decide what the best solution is ...

jcameron:
Actually, is there a distro on which the current behavior causes a problem out of the box?

iliajie:
> Actually, is there a distro on which the current behavior causes a problem out of the box?

As far as I know, you can currently reproduce this issue only by manually installing MariaDB 11 from their repos. However, it doesn’t make much difference since users can install either MySQL or MariaDB on the same distro, and we’re handling both within a single module...

iliajie:
What is the status of this one?

jcameron:
I don't think we should make this change - automatically updating the paths to `mysql` is just too "magical" and goes against the way Webmin sets default paths. Instead we can just support the versions of mysql or mariaDB that come standard with distros..

iliajie:
> Instead we can just support the versions of mysql or mariaDB that come standard with distros..

But, distros have support for both MariaDB and MySQL...

jcameron:
But do any of those distros not have a command called `mysql` when MariaDB is installed?

iliajie:
Yes, currently they have but it will change in the near future.

jcameron:
Well, let's worry about that if and when it happens then

iliajie:
> Well, let's worry about that if and when it happens then

I believe it would be better to address it sooner since it is already possible to install it manually, right?

If you’re not happy with this particular implementation, what would be a more effective way to approach it? Or perhaps we should consider reverting the commands back from MariaDB to MySQL as well (not just one way, as it is now in this PR)?

iliajie:
I'm sorry, but I need to bump it up again. What are we going to do to resolve it, considering everything we discussed earlier?

jcameron:
I still don't think this is a good idea, or necessary. Once we see a distribution that replaces the `mysql` command with `mariadb` by default in the standard packages, we'll address it in the Webmin-standard way by creating a new config-* file.

iliajie:
> I still don't think this is a good idea, or necessary. Once we see a distribution that replaces the `mysql` command with `mariadb` by default in the standard packages, we'll address it in the Webmin-standard way by creating a new config-* file.

Jamie, you’re assuming that distros with the new MariaDB version—where `mysql*` commands are replaced with `mariadb_*`—will still have MySQL packages in their repos, right? Can you explain how Webmin will handle this with the current `config-` design? Because a user could have either installed!

jcameron:
> So, how can it be handled?

Let's say Debian 19 comes with a default MariaDB package that uses the command `mariadb` instead of `mysql`. If so, we will create `mysql/config-Debian-Linux-19-*` that contains `mysql=/usr/bin/mariadb` so that the correct command is detected.

That's all we need to do!

iliajie:
> That's all we need to do!

Is that so!?

> Let's say Debian 19 comes with a default MariaDB package that uses the command mariadb instead of mysql.

Sure, let's assume that, but also remember that Debian 13 will come with MySQL packages **as well**! So what happens when a user installs MySQL instead of MariaDB? All those `mariadb` commands we include in `mysql/config-Debian-Linux-13-*` won't work!

Therefore, as I already mentioned, the solution is either similar to this PR or more complex, requiring the creation of logic for Webmin to handle alternative config filenames such as `mysql/config-Debian-Linux-13-mysql-8*` and `mysql/config-Debian-Linux-13-mariadb-11*`.

The Trixie release is just around the corner, which is why I'd like to have it working rather than not in the upcoming Webmin release.

jcameron:
Sure one of MySQL or MariaDB will be the default though? We can just set the default configs to match that.

iliajie:
> Sure one of MySQL or MariaDB will be the default though? We can just set the default configs to match that.

What do you mean? Both are available, right?

Example:

```
root@ubuntu24-gpl:~# apt-get install mysql-server
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
The following packages were automatically installed and are no longer required:
  galera-4 libmariadb3 libsnappy1v5 liburing2 pv socat
Use 'apt autoremove' to remove them.
The following additional packages will be installed:
  libevent-pthreads-2.1-7t64 libmecab2 libmysqlclient21 libprotobuf-lite32t64 mecab-ipadic mecab-ipadic-utf8 mecab-utils
  mysql-client-8.0 mysql-client-core-8.0 mysql-server-8.0 mysql-server-core-8.0
Suggested packages:
  mailx tinyca
The following packages will be REMOVED:
  mariadb-client mariadb-client-core mariadb-plugin-provider-bzip2 mariadb-plugin-provider-lz4 mariadb-plugin-provider-lzma
  mariadb-plugin-provider-lzo mariadb-plugin-provider-snappy mariadb-server mariadb-server-core
The following NEW packages will be installed:
  libevent-pthreads-2.1-7t64 libmecab2 libprotobuf-lite32t64 mecab-ipadic mecab-ipadic-utf8 mecab-utils mysql-client-8.0
  mysql-client-core-8.0 mysql-server mysql-server-8.0 mysql-server-core-8.0
The following packages will be upgraded:
  libmysqlclient21
1 upgraded, 11 newly installed, 9 to remove and 282 not upgraded.
Need to get 29.8 MB of archives.
After this operation, 56.3 MB of additional disk space will be used.
Do you want to continue? [Y/n] n
Abort.

```

iliajie:
> What do you mean? Both are available, right?

Jamie, I'd like to ask you to comment on my [previous comment](https://github.com/webmin/webmin/pull/2320#issuecomment-2614095834).

jcameron:
Just to clarify, on Ubuntu 24 if the user installs the `mariadb` packages does the command `mysql` actually no longer exist?

iliajie:
> Just to clarify, on Ubuntu 24 if the user installs the `mariadb` packages does the command `mysql` actually no longer exist?

It exists only for _now_, but it _won't_ be in the _near_ future in newer releases, just as it won't be in Debian 13 or RHEL 10. Some of these releases are coming this summer...

jcameron:
In that case, let's wait until this actually happens before worrying about it. Unless there is some statement by the Debian or RHEL developers canonically saying that the `mysql` command is going away.

My opinion is that this will never happen, just like how the `sendmail` command still exists on systems using Postfix.

iliajie:
> Unless there is some statement by the Debian or RHEL developers canonically saying that the `mysql` command is going away.

I highly doubt it’s controlled by either Debian or RHEL. This change is coming from the developers of MariaDB, for versions 11 and up.

jcameron:
Well, the distro packagers are going to be the ones controlling what binaries and symlinks to binaries are installed.

iliajie:
The symlinks are going to be removed, so we have to use `mariadb_` commands directly—that’s the reason for the hassle.

jcameron:
Like I said before, let's wait till we see this in a MariaDB package supplied by a distro vendor before we make any changes.

PR #2326, opened 2024-11-28
Remove one mention of inetd

swelljoe:
inetd is pretty much deprecated on Linux systems with the move to systemd (socket-activation takes the place of it and inetd or xinetd will be missing on any modern Linux system).

I haven't tried to remove inetd mentions from everywhere, as inetd is still maybe in common use in the BSD world (though Mac OS has `launchd` instead). But, it probably shouldn't be prominently recommended anywhere, since most Webmin users are on Linux and leading them down ancient cobblestone roads isn't a great user experience.


proftpd/lang/ru
```diff
@@ -6,8 +6,8 @@ index_econf=Файл настроек ProFTPD $1 не существует ил
 index_apply=Применить изменения
 index_applymsg=Нажмите на эту кнопку для применения текущих настроек отправкой сигнала SIGHUP запущенным процессам ProFTPD.
 index_start=Запуск сервера
-index_startmsg=Нажмите на эту кнопку для запуска сервера FTP с текущими настройками. Вы также можете настроить автоматический запуск сервера по необходимости, используя модуль <a href='$1'>Службы интернет</a>.
-index_startmsg2=Нажмите на эту кнопку для запуска сервера FTP с текущими настройками. Вы также можете настроить автоматический запуск сервера по необходимости, используя inetd ил другие эквивалентные программы.
+index_startmsg=Нажмите на эту кнопку для запуска сервера FTP с текущими настройками.
+index_startmsg2=Нажмите на эту кнопку для запуска сервера FTP с текущими настройками.
```
iliajie, line 10: I think we could just use the language manager to update it for all other languages—it’s too much to change manually. I’ll take care of it—no worries.

PR #2335, opened 2024-12-16
Blanks Removal

MacSteini:
Removed blanks before or after the equal sign


MacSteini:
Script's here in case you want to use it:
https://gist.github.com/MacSteini/bda6ce5bc947c3c0a26d5247d3ae9d00

PR #2345, opened 2024-12-28
A user was confused by 'The Webmin server'

swelljoe:
https://forum.virtualmin.com/t/redirecting-traffic-per-country-ips/131286/6?u=joe


iliajie:
Thanks for the patch, Joe!

PR #2346, opened 2024-12-30
Updated de & fr

MacSteini:
Updated de

Made a fundamental mistake when I ran an automated script and messed up quota/lang/fr


file/lang/de
```diff
@@ -1,6 +1,6 @@
 index_title=Java-Dateimanager
 index_nojava=Dieses Modul benötigt Java, um zu funktionieren, aber Ihr Browser unterstützt Java nicht.
-index_eremote=Es gibt keine:n Unix-Benutzer:in, der dem Webmin-Login $1 entspricht.
+index_eremote=Es gibt keine:n Unix-Benutzer:in, der/die dem Webmin-Login $1 entspricht.
```
iliajie, line 3: Thanks! Though, isn’t there a more inclusive way to phrase this? What about using `xier` instead of the cumbersome `der/die`?

MacSteini: This is the common and established way it is being used in German language.

iliajie: Alright!

iliajie:
Thanks for the patches!

PR #2349, opened 2025-01-02
Generation of repo files for (open)suse distributions added

rhabacker:
The current script `webmin-setup-repo.sh` does not work without manual adjustments for openSUSE distributions.

With the changes from this pull request, adding the webmin repo will work for the mentioned distributions without further customization.

### DIFF
--- a/webmin-setup-repo.sh
+++ b/webmin-setup-repo.sh
@@ -14,6 +14,8 @@ debian_repo_file="/etc/apt/sources.list.d/webmin.list"
 debian_repo_file_testing="/etc/apt/sources.list.d/webmin-testing.list"
 rhel_repo_file="/etc/yum.repos.d/webmin.repo"
 rhel_repo_file_testing="/etc/yum.repos.d/webmin-testing.repo"
+suse_repo_file="/etc/zypp/repos.d/webmin.repo"
+suse_repo_file_testing="/etc/zypp/repos.d/webmin-testing.repo"
 download_curl="/usr/bin/curl"
 download="$download_curl -f -s -L -O"
 testing_mode=0
@@ -84,6 +86,7 @@ detect_os() {
 
   osid_debian_like=$(echo "$osid" | grep "debian\|ubuntu")
   osid_rhel_like=$(echo "$osid" | grep "rhel\|fedora\|centos\|openEuler")
+  osid_suse_like=$(echo "$osid" | grep "suse")
   repoid_debian_like=debian
   if [ -n "${ID}" ]; then
     repoid_debian_like="${ID}"
@@ -106,6 +109,11 @@ detect_os() {
       install="$install_cmd -y"
       clean="yum clean all"
     fi
+  elif [ -n "$osid_suse_like" ]; then
+    package_type=rpm-md
+    install_cmd="zypper install"
+    install="$install_cmd -y"
+    clean="zypper clean --all"
   else
     echo "${RED}Error:${NORMAL} Unknown OS : $osid"
     exit 1
@@ -172,6 +180,34 @@ download_key() {
 
 setup_repos() {
   case "$package_type" in
+    rpm-md)
+      if [ "$testing_mode" = "1" ]; then
+        echo "  Setting up Webmin testing repository .."
+        cat << EOF > "$suse_repo_file_testing"
+[webmin-testing]
+name=Webmin Testing
+baseurl=$webmin_download_testing
+enabled=1
+autorefresh=1
+type=rpm-md
+gpgkey=$webmin_key_download
+gpgcheck=1
+EOF
+      else
+        echo "  Setting up Webmin repository .."
+        cat << EOF > "$suse_repo_file"
+[webmin]
+name=Webmin Stable
+baseurl=$webmin_download/download/newkey/yum
+enabled=1
+autorefresh=1
+type=rpm-md
+gpgkey=$webmin_key_download
+gpgcheck=1
+EOF
+      fi
+      echo "  .. done"
+      ;;
     rpm)
       echo "  Installing Webmin key .."
       rpm --import "$webmin_key"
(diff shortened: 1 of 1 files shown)


webmin-setup-repo.sh
```diff
@@ -106,6 +109,11 @@ detect_os() {
       install="$install_cmd -y"
       clean="yum clean all"
     fi
+  elif [ -n "$osid_suse_like" ]; then
+    package_type=rpm-md
+    install_cmd="zypper install"
+    install="$install_cmd -y"
+    clean="zypper clean."
```
iliajie, line 116: Why is there a period at the end of the command?

rhabacker: Thanks for the pointer. I fixed that.

webmin-setup-repo.sh
```diff
@@ -172,6 +180,34 @@ download_key() {
...
+EOF
+      else
+        echo "  Setting up Webmin repository .."
+        cat << EOF > "$suse_repo_file"
+[webmin]
+name=Webmin Stable
+baseurl=$webmin_download/download/newkey/yum
+enabled=1
+autorefresh=1
+type=rpm-md
```
iliajie, line 204: I think it would be cleaner to abstract the logic here using a few variables specifically needed for Open(SUSE), rather than re-creating the entire repository, since most of the fields are identical.

Also, I believe you should keep `-noarch` in the repository name, as Open(SUSE) supports the concept of noarch packages...

webmin-setup-repo.sh
```diff
@@ -256,6 +292,8 @@ detect_os
 ask_confirmation
 check_downloader
 check_gpg
-download_key
+if [ ! $package_type == "rpm-md" ]; then
```
iliajie, line 295: We want this script to be POSIX-compatible; use the following check instead:

```
if [ "$package_type" != "rpm-md" ]; then
```

webmin-setup-repo.sh
```diff
@@ -256,6 +292,8 @@ detect_os
...
+    download_key
```
iliajie, line 296: Why not download the key in SUSE? I don’t think it’s right.

I’d suggest making fewer changes and adding only the SUSE-specific lines using a _variable_ to add to existing repos that will remain empty for EL systems...

iliajie:
Thanks!

Check my comments on the commits... By the way, I can take care of adding SUSE support, if you don't mind me stomping on your PR; otherwise, take your time. No rush.

webmin-setup-repo.sh
rhabacker: See https://download.opensuse.org/repositories/devel:/tools:/building/15.6/ as an example for a openSUSE repo:

The associated [repo file](https://download.opensuse.org/repositories/server:/php:/extensions/15.6/server:php:extensions.repo) looks as: 

[server_php_extensions]
name=PHP extensions (15.6)
type=rpm-md
baseurl=https://download.opensuse.org/repositories/server:/php:/extensions/15.6/
gpgcheck=1
gpgkey=https://download.opensuse.org/repositories/server:/php:/extensions/15.6/repodata/repomd.xml.key
enabled=1

It is normal behavior to use a remote url for the gpg key.

iliajie: Thanks!

> type=rpm-md

There is no need for this, as it's the default.

> It is normal behavior to use a remote url for the gpg key.

Yes, I'm aware, but why not install it as is, considering the existing code already does it?

I'm looking into this now...

rhabacker: According to my current knowledge, the public gpg key file repodata/repomd.xml.key (and maybe repodata/repomd.xml.asc) is necessary for openSUSE in order to avoid the error message listed in https://github.com/webmin/webmin/issues/2350. It is therefore superfluous to download the gpg public key separately.

iliajie:
Danke! I have decided to merge your PR. However, I had to revise the support for SUSE distros. You can view the full changes here https://github.com/webmin/webmin/commit/bdf6e37de880fc9e478c173842815a7ed6d45042.

webmin-setup-repo.sh
rhabacker: I just checked the repo: the .asc file is already there, but repodata/repomd.xml.key is still missing.

iliajie: >the public gpg key file repodata/repomd.xml.key and repodata/repomd.xml.asc 

Not "and", I think "or". 

> I just checked the repo: the .asc file is already there, but repodata/repomd.xml.key is still missing.

Is something not actually working!?

rhabacker: Newly adding the webmin repo with the script mentioned at https://webmin.com/download/ works as expected, 
but updating a repo setup as 
```
cat /etc/zypp/repos.d/webmin.repo 
[webmin]
name=webmin
enabled=1
autorefresh=1
baseurl=https://download.webmin.com/download/yum/
gpgcheck=1
gpgkey=https://download.webmin.com/download/yum/repodata/repomd.xml.asc
```
with 
```
sudo zypper ref webmin
```
fails:
```
Suche nach der GPG-Key-ID 11F63C51 in cache /var/cache/zypp/pubkeys.
Suche nach der GPG-Key-ID 11F63C51 in Repository webmin.
  gpgkey=https://download.webmin.com/download/yum/repodata/repomd.xml.asc
Warning: The file “repomd.xml” from the repository “webmin” is signed with the unknown key “D97A3AE911F63C51”. 
```

Changing 
```
gpgkey=https://download.webmin.com/download/yum/repodata/repomd.xml.asc
```
to 
```
gpgkey=https://download.webmin.com/developers-key.asc
```
and rerunning
```
sudo zypper ref webmin
```
does not help.

iliajie: > Newly adding the webmin repo with the script mentioned at https://webmin.com/download/ works as expected,

I'm sorry, I'm a bit confused. Do the latest changes in the Webmin repo setup script make things work correctly on SUSE? Is that right?

If so, what are the other mentioned steps for? Could you just delete `/etc/zypp/repos.d/webmin.repo` and set it up again with the new working Webmin repo setup script?

Issue #2353, opened 2025-01-07
webmin and Alpine linux: sucess but now add a minimal compatibility


mckaygerhard:
I was able to get the minimal webmin working at 99% on alpine linux. With some tweaks, but for now, until we have Alpine on the list, we will assume Gennto as the distro (I assumed that because of the openrc) that is the most close choose due the `webmin/webmin-geento-init` script type of openrc

## Feature REQUEST

This feature request is **not for official support but at least "minimal support" or "forward compatibility" for Alpine Linux**. 

**MEMORANDUM OF UNDERSTANDING**: Because we must understand that adding another layer of support is further complicating the already long list of supported distros and combinations, **that was i put as "minimal support" or "forward compatibility" for**!

### i can collaborate, so what i need?

1. **files to touch**: actually, although i must do this by myselft, it is not a bad idea to be told which files to touch at least so that Alpine Linux can be managed. I listed the files in the next section below!
2. **package manager** we already know that package manager of alpine linux is unique so we need to define this.. the most close to is apt command, apk package manager is far simple from any other, most like minimalist distros, so any tip what i should take care for?
3. **pam module integration** relies on Authen::PAM module for perl that is not yet as package on alpine so manual compilation its need, by the moment can manage by special instructions at my guide https://t.me/alpine_linux_english/81731 so, unfortunately _usermin_, and therefore _virtualmin_ does not work yet out of the box, but of course can be addressed

so **then i can send a pull request** for!

### Alpine file list TODO

By the moment i listed those files and the minimal case use for:

- [ ] `webmin/webmin-geento-init` due the openinitrc but alpine also provides s6 and supervisor features
- [ ] `webmin/os_list.txt` to list supported versions of alpine linux number releases, from 3.0 its well supproted as i tested
- [ ]  `webmin/dhcpd/config.info[.*]` for DHCPd based but as we know ISC DHCP will be superset by KEA https://www.isc.org/dhcp_migration/
- [ ] `webmin/dhcpd/edit_iface.cgi` for the same as previous pointed
- [ ] `webmin/upgrade.cgi` to manage package install on upgrades into `mode` variable
- [ ] `webmin/lang/*` for the moment this is not so importat for minimal funcionality! we can leave this for later
- [ ] `webmin/firewall/lang/*` for the moment this is not so importat for minimal funcionality! we can leave this for later
- [ ] `webmin/cluster-webmin/lang/*` for the moment this is not so importat for minimal funcionality! we can leave this for later
[...]

Issue #835, opened 2018-02-27
init: systemd auto-detection doesn't work on Gentoo


vampik:
Webmin 1.870. Gentoo x86_64 with systemd.

Webmin does not detect SystemD if "Boot system" is set to "Detect automatically"

This happens because the detection code checks for existence of `$config{'init_base'}` for some reason. 

`init_base` doesn't exist in Gentoo config because Webmin doesn't support OpenRC.

I've just removed this check to fix the issue. I cannot understand why is it needed? This is a config option for old init-based systems.

```
--- /usr/libexec/webmin/init/init-lib.pl.bak    2017-12-06 21:16:48.000000000 +0300
+++ /usr/libexec/webmin/init/init-lib.pl        2018-02-27 18:06:57.716000000 +0300
@@ -62,8 +62,7 @@
        &execute_command("/sbin/init --version") == 0) {
        $init_mode = "upstart";
        }
-elsif ($config{'init_base'} && -d "/etc/systemd" &&
-       &has_command("systemctl") &&
+elsif (-d "/etc/systemd" && &has_command("systemctl") &&
        &execute_command("systemctl list-units") == 0) {
        $init_mode = "systemd";
        }

```

jcameron:
You're right - I'll fix this in the 1.890 Webmin release.

PR #2354, opened 2025-01-08
update openrc based outdated script in sync with current gentoo and alpine

mckaygerhard:
* This script its from ancient webmin era.. we need up to date to be able to use for alpine and modern gentoo linux distributions, gentoo still uses Openrc for many things no matter if you choose other init systems, also as or issue #835 we can noted there are much more options to setup on geento that seems are not well supported, webmin 
seems do not use the scripts from the OS init system, it seems if you use the web UI it will use directly the scripts from the config `WEBMIN_CONFIG` directory
* Fix status command not available to, by the detection of the pid, this is using the config pid variable from webmin config miniserv is need for future support on issue #2353 and related to #835 that just assume init sysv systems or the other s*** only. Still is pending future support for openrc init system but with this path next changes are more easy to do
* The pull request is open to maintainers editing, because we can just merge or improve the  `updateboot.pl` 
script with a new procedure method defined for `openrc` and not just assumed `sysvinit` as the only other present
and also `atboot.pl` with new procedure for openrc the alpine linux init system at the moment! This cos we need to setup and use `WEBMIN_CONFIG` variable to setup the init system in future improvements

### DIFF
--- a/webmin-gentoo-init
+++ b/webmin-gentoo-init
@@ -1,19 +1,52 @@
-#!/sbin/runscript
-# Copyright 1999-2002 Gentoo Technologies, Inc.
+#!/sbin/openrc-run
+
+# Copyright 2024 webmin & mckaygerhard
 # Distributed under the terms of the GNU General Public License, v2 or later
 
+
+name="webmin"
+# config file must be in sync with setup script, so i will not touch cos process will be the same as webmin-systemd
+# like calling updateboot.pl with a new procedure method defined as openrc and not just assumed sysvinit if not systemd is present
+# and also atboot.pl with new procedure for openrc the alpine linux init system at the moment
+#conf_file="WEBMIN_CONFIG/miniserv.conf"
+conf_file="/etc/webmin/miniserv.conf"
+
+
 depend() { 
-	need net logger 
+    need logger localmount
+    use net
+    after bootmisc
 }
 
 start() {
-	ebegin "Starting webmin"
-	/etc/webmin/start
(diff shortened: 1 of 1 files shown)


jcameron:
Thanks for the patch! Does gentoo ever use systemd though?

mckaygerhard:
yes @jcameron   it allows to choose five different init systems.. so as i can see over the source code seems  webmin determine and set an init way based on the distro choose only.. apart of the check of #835 ..

i tested webmin init changes.. and webmin just try to detect and just falbacks to sysvinit bash based init script and seems to work under openrc (tested over Alpine) 

so seems well i dont know i must rely on your tips and directions, i conclude that we can rely on the fiable sysvinit  `webmin-init` template for Alpine and Geento and Devuan no matter if the OS uses openrc or sysvinit system...

mckaygerhard:
apart of the previous comment i have another question..  apart of the usage of the sysvinit (`webmin/webmin-init`)vs openrc (`webmin/webmin-gentoo-init`) scripts .. the setup script uses env variables to setup the other init system.. but seems the openrc scritp is not used since long time ago right?  (`webmin/webmin-gentoo-init`, please rea the previous comment in this pull Request, its closed but is to not over flood the other issues about init scripts)   @jcameron ? @iliajie ?

mckaygerhard:
hi @jcameron  @iliajie  i kown you are busy guys but i need answered thos two questions (i made here in #2354 to not made flood over the issue #2353

PR #2369, opened 2025-01-20
Delete record extra log

cstrotm:
Add log entry to Webmin logs for every DNS record deleted

### DIFF
--- a/bind8/delete_recs.cgi
+++ b/bind8/delete_recs.cgi
@@ -76,7 +76,7 @@ else {
 		# Delete the actual record
 		&lock_file(&make_chroot($r->{'file'}));
 		&delete_record($r->{'file'}, $r);
-    &webmin_log("delete", "record", $dom, $r);
+		&webmin_log("delete", "record", $dom, $r);
 		splice(@recs, $d, 1);
 		}
 	&bump_soa_record($zone->{'file'}, \@recs);
--- a/bind8/delete_recs.cgi
+++ b/bind8/delete_recs.cgi
@@ -76,6 +76,7 @@ else {
 		# Delete the actual record
 		&lock_file(&make_chroot($r->{'file'}));
 		&delete_record($r->{'file'}, $r);
+    &webmin_log("delete", "record", $dom, $r);
 		splice(@recs, $d, 1);
 		}
 	&bump_soa_record($zone->{'file'}, \@recs);
@@ -86,5 +87,3 @@ else {
 	&webmin_log("delete", "recs", scalar(@d));
 	&redirect("edit_recs.cgi?zone=$in{'zone'}&view=$in{'view'}&type=$in{'type'}&sort=$in{'sort'}");
 	}
-
-


jcameron:
Thanks for the fix!

iliajie:
This doesn’t seem to be the correct change, as we already had logging for deleted records, as shown in the screenshot below:

image

However, this change introduces an additional log entry (above "Deleted 2 records"), which is quite confusing.

Furthermore, the new additional log entry has no actual diff, making it seem like it adds no real value.

I’ll revert the merge and leave it open for further discussion.

cstrotm:
The existing log entry just gives the sum of the deleted records, but no information which records got deleted. 

I would argue there is value in having one log entry for each record deleted AND the total sum. The total sum entry gives the diff, it would be redundant to have a diff on each record delete. But the diff is not easy searchable. With the new change, it is possible to search for the record name. In some environments it is a common task to search for deleted records in the log by zone name or by owner-name (domain name).

iliajie:
> With the new change, it is possible to search for the record name. In some environments it is a common task to search for deleted records in the log by zone name or by owner-name (domain name).

I like the reasoning and completely agree with making it more searchable, but I’m not in favor of having two separate records for one action. So, let’s modify the current code to just include “in zone name” so it looks like:

```
Deleted 2 records in zone ubuntu24-gpl.virtualmin.dev
```

iliajie:
Or, alternatively, we could log each delete action as you suggested earlier, but ensure that each entry includes a meaningful diff and avoid printing a redundant summary log. Another option is to print a single log message but include details like the record type (e.g., Address), the specific record (e.g., www or ns), and the zone name...

iliajie:
> If given the choice, I would prefer a line for each deleted record over a summary log entry.

Alright, I agree—it sounds reasonable. Let’s go ahead with it, but _make sure_ the diff is properly saved as well.

Will you take a look at it, or would you prefer us to handle it?

cstrotm:
> Alright, I agree—it sounds reasonable. Let’s go ahead with it, but _make sure_ the diff is properly saved as well.

Thank you.

> 
> Will you take a look at it, or would you prefer us to handle it?

I'm pretty new to the Webmin code base, I need to learn my ways around. Currently I'm busy with other projects, but I can look into implementing this in February. 

But if you or someone else finds time before I work on this, please do.

jcameron:
Now I think about this some more, the log message probably shouldn't include all the deleted records as there could be a large number. But the diff should certainly be logged.

Perhaps as a compromise we could change what's logged if just one record is deleted?

iliajie:
> Now I think about this some more, the log message probably shouldn't include all the deleted records as there could be a large number. But the diff should certainly be logged.
> 
> Perhaps as a compromise we could change what's logged if just one record is deleted?

This change https://github.com/webmin/webmin/commit/1ad15e0bfb6e05650c1a30f250b4f4792570d387 logs every deleted record on that page, in a similar way adding records work. And, it does log the change for each correctly.

jcameron:
I don't think that's the right approach though, as in general we only perform one logging call at the end of a CGI script.

iliajie:
I can easily revert it, but isn’t this the same change was introduced in this PR, just with a non-functional diff?

jcameron:
Yes, upon further consideration the original PR wasn't a good approach

cstrotm:
The use case of such logging is for the user of Webmin to quickly find the answer to the question: ***Who*** has deleted the record ***foo.bar.baz***  ***when*** (time, user, record content).

What other ways would be available to find this information?

jcameron:
Maybe what we really need is a way to search logs by the contents of the diffs?

iliajie:
> Maybe what we really need is a way to search logs by the contents of the diffs?

That would definitely help, though using a simpler search, like `Ctrl+F` in the browser, wouldn’t hurt either.

> Yes, upon further consideration the original PR wasn't a good approach

What’s the issue with having it displayed multiple times in the log, really? Or is it not about the display at all?

Either way, for now, I’ve reverted my patch here https://github.com/webmin/webmin/commit/ae5c7a9db956ed6dca5d6390c0c4104c75af4c09.

jcameron:
> Maybe what we really need is a way to search logs by the contents of the diffs?

I'll add this to my TODO list.

> What’s the issue with having it displayed multiple times in the log, really? Or is it not about the display at all?

Because it goes against the current standard of one user action generating one log entry.

jcameron:
> I'll add this to my TODO list.

Actually this feature already exists - you can search by diff contents in the Webmin Actions Log module already using the **
File change or command run** box.

cstrotm:
(Thinking loud): maybe a kind of "hierarchical" logging is the solution: a single log-line with the summary information in one line (42 records deleted in zone "bar.baz"), and then a UI interface that expands this log entry and shows all the detail information (information about each record deleted). 

The logfile (on disk) could contain both the summary and the detail information, and each log-entry would have a kind of unique ID (transaction ID) that makes it possible to connect them. 

The WebUI would hide all the details unless the user will select/opens the summary entry.

The search function would have an option to search in summary (default) or detail entries.

(I'm aware that this might be a major rework of the logging function in Webmin)

cstrotm:
> > I'll add this to my TODO list.
> 
> Actually this feature already exists - you can search by diff contents in the Webmin Actions Log module already using the **File change or command run** box.

I did not know that, I will try and report if this is the solution.

iliajie:
> Because it goes against the current standard of one user action generating one log entry.

Sure! Instead of just saying "Deleted 2 records," we could improve the main log message by saying something like "Deleted 2 records of type Address, Mail Server in zone domain-name.tld." This would be much clearer, regardless of whether it's searchable or not.

jcameron:
Oh yeah we should definitely include the zone name in the log message! I will fix that ...

Issue #2307, opened 2024-11-07
Bandwidth Monitoring not working on Debian 12?


Sebastian1989101:
I have a fresh installed Debian 12 vServer and just added Webmin. One reason I added Webmin was for the Bandwidth Monitoring as this Server works as a VPN. However, the module just reports: 

`Neither of the System Logs modules are installed on this system and supported by your OS. The Bandwidth Monitoring module cannot operate without one of them.`

The command `journalctl -k --grep "BANDWIDTH_IN|BANDWIDTH_OUT"` also just yields `-- No entries --`

iliajie:
Hey, Sebastian!

Thanks for the heads up! I like your suggestion! You’re absolutely right—we need to fix this! The problem is that most modern systems no longer use `rsyslog` and rely on `firewalld` to interact with `iptables`.

I dug in a bit and found that it might be simpler than I initially thought.

Try running this bash script, replacing the interface name (and possibly zone name) with the one you’re using:

```bash
#!/usr/bin/env bash

# Incoming traffic logging without rate limit (limit could be imposed using `limit value="1/m"`)
firewall-cmd --permanent --zone=public --add-rich-rule='rule family="ipv4" source address="0.0.0.0/0" log prefix="BANDWIDTH_IN: " level="info" accept'

# Outgoing traffic logging without rate limit
firewall-cmd --permanent --zone=public --add-rich-rule='rule family="ipv4" destination address="0.0.0.0/0" log prefix="BANDWIDTH_OUT: " level="info" accept'

# Reload FirewallD to apply changes
firewall-cmd --reload

# Verify the rules are in place
firewall-cmd --zone=public --list-rich-rules
```

Then you'd be able to get bandwidth logs using:

```bash
journalctl -f -t kernel | grep "BANDWIDTH_"
```

iliajie:
@jcameron Jamie, what are your thoughts on this? Should we fix bandwidth monitoring for systems using `firewalld` and `journalctl`?

jcameron:
@iliajie yes that sounds like a good idea, although potentially complex to support firewalld since we'd first need to add support for rich rules in Webmin.

iliajie:
I don’t think we need to care about rich rules support at all, as it’s handled by FirewallD, right?

iliajie:
Can you elaborate on this? This is handled by `firewall-cmd` command? What do you mean by saying “we do need to support that”?

iliajie:
[...]

PR #2371, opened 2025-01-20
Add support for bandwidth monitoring with Firewalld and Journald

iliajie:
Hey Jamie!

As discussed here (https://github.com/webmin/webmin/issues/2307), this PR adds support for bandwidth monitoring with Firewalld and Journald. It’s also compatible with a combination of Firewalld and Rsyslog.

The implementation wasn’t simple, but it turned out to be much cleaner than I initially expected.

It uses a new direct rules API for Firewalld, which I added while working on this PR. Note that the Firewalld patches aren’t included in this PR because I committed them directly.

Take a look, please!

### DIFF
--- a/bandwidth/bandwidth-lib.pl
+++ b/bandwidth/bandwidth-lib.pl
@@ -8,17 +8,17 @@
 BEGIN { push(@INC, ".."); };
 use WebminCore;
 &init_config();
-if (&foreign_installed("syslog-ng", 1) == 2) {
+if (&has_command('journalctl')) {
+	$syslog_module = undef;
+	$syslog_journald = "journald" ;
+	}
+elsif (&foreign_installed("syslog-ng", 1) == 2) {
 	&foreign_require("syslog-ng");
 	$syslog_module = "syslog-ng";
 	}
 elsif (&foreign_installed("syslog")) {
 	&foreign_require("syslog");
 	$syslog_module = "syslog";
-	}
-else {
-	$syslog_module = undef;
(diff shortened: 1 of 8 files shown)


bandwidth/bandwidth-lib.pl
```diff
@@ -8,26 +8,30 @@
...
 
 $bandwidth_log = $config{'bandwidth_log'} || "/var/log/bandwidth";
-$hours_dir = $config{'bandwidth_dir'} || "$module_config_directory/hours";
+$hours_dir = $config{'bandwidth_dir'} ||
+	     (-e "$module_config_directory/hours" ?
+	     		"$module_config_directory/hours" :
+			"$module_var_directory/hours");
 $cron_cmd = "$module_config_directory/rotate.pl";
-$pid_file = "$module_config_directory/rotate.pid";
+$pid_file = "$module_var_directory/rotate.pid";
```
jcameron, line 34: isnt this going to break existing systems where rotate job is already running?

iliajie: Alright, thanks! Good catch, fixed now!

jult:
Nice. Does this module have a specific info page somewhere? Looks like the one under bandwidth monitoring is not there..

iliajie:
> Nice. Does this module have a specific info page somewhere? Looks like the one under bandwidth monitoring is not there..

What about Webmin documentation:

https://webmin.com/docs/modules/bandwidth-monitoring/

PR #2373, opened 2025-01-23
Fix to replace `Authen::OATH` validation with simpler `Digest::HMAC_SHA1`

iliajie:
Hello Jamie,

As discussed in the chat, this PR removes thousands of lines of code and eliminates the dependency on `Authen::OATH`, replacing it with a much simpler implementation for validating TOTP.

Please take a look!


jcameron:
Thanks, this is excellent!

iliajie:
> Thanks, this is excellent!

You're welcome!

PR #2388, opened 2025-02-04
Added updated modules

abcbarryn:
Added updated Custom Link module and a java script based VNC client module. Both use an iframe to embed content in the Webmin page. This should work with most if not all themes, old and new.


abcbarryn:
The original modules were legacy. VNC is almost a complete rewrite, so no longer legacy. The updated Custom Link module is now compatible with modern themes as well. I will possibly rework some of the other modules currently flagged as legacy to bring them up to date. Is there a branch for current/new modules other than master?

iliajie:
I can see you have added PRs to [modules/legacy](https://github.com/webmin/webmin/tree/modules/legacy) branch—I will close this one.

PR #2392, opened 2025-02-05
Add ability to show reverse dependencies

iliajie:
Hey Jamie!

As discussed in the chat, this PR adds the ability to return a reverse dependency. For example, if `php-calendar` is given as the package name, the returned package will be `php-common`.


jcameron:
Excellent, thanks! This is perfect ...

iliajie:
> Excellent, thanks!

You're welcome!

PR #2400, opened 2025-02-09
Update de

MacSteini:


firewall/lang/de.auto
```diff
@@ -0,0 +1 @@
+index_no_comment=
```
iliajie, line 1: Why add `index_no_comment=` to a separate `de.auto` file?

MacSteini: I usually discard the change since it keeps getting pulled back in, but this time I forgot. It might be worth looking into why this keeps happening - whenever I exclude de.auto, it comes back in with the next push cycle.

iliajie: Ah, that's right—this is a stray key, and the language manager expects a value.

PR #2402, opened 2025-02-10
Update de and Question

MacSteini:
The original en file includes:
```
+value_dmarcri=Reporting interval
+value_dmarcrf=Report format
+value_dmarcafrf=Authentication Failure Reporting Format
```

Are the `+` signs intentionally prefixed? If so, ignore my submission, will then redo.


iliajie:
Thank you for pointing this out—it's definitely a bug! It's fixed now!

PR #2406, opened 2025-02-13
Add better way to handle trust level for proxy headers option

iliaross:
G'day Jamie!

As discussed in the chat, this PR improves and simplifies the UI options to determine whether Webmin should trust headers from a proxy.


webmin/postinstall.pl
```diff
@@ -46,8 +46,8 @@ sub module_install
 	}
 
 # Disable trusting SSL certs unless already enabled
-if (!$miniserv{'trust_real_ip'} && !defined($miniserv{'no_trust_ssl'})) {
-	$miniserv{'no_trust_ssl'} = 1;
+if ($miniserv{'trust_real_ip'} == 1 && &get_webmin_version() <= 2.300) {
```
jcameron, line 49: There's no need for a version check because we know what version this is, right??

miniserv.pl
```diff
@@ -1704,7 +1704,7 @@ sub handle_request
 	}
 
 # Check for SSL authentication
-my $trust_ssl = $config{'trust_real_ip'} && !$config{'no_trust_ssl'};
+my $trust_ssl = $config{'trust_real_ip'} == 2;
```
jcameron, line 1707: Why not just keep these two config variables internally, but show only one selector in the UI?

webmin/postinstall.pl
iliajie: If a user installs version 2.300 and then changes this option back to `1`, we don't want it to be reverted upon the release of Webmin 2.301 and etc...

miniserv.pl
iliajie: There is no need for an additional variable because we now combine it under one radio element. Adding it would represent unnecessary complexity.

webmin/postinstall.pl
jcameron: Yes but what if the user skips version 2.300 and installs 2.301 ? We cannot rely on checking the version in the post-install script.

miniserv.pl
jcameron: Yes, but you have moved the complexity elsewhere!

PR #2414, opened 2025-02-17
Update de

MacSteini:


iliajie:
Thanks for submitting this change!

However, is it possible to fix UI choices, such as those mentioned in radio buttons, for example **● Yes ○ No** to put in tooltips (those edited HTML files in the PR) not as `<tt>Yes</tt>` but rather as `"Yes"`? Example https://github.com/webmin/webmin/commit/42b823e3e9b32b1a4d0148ddbd4952892332bdf7.

Also, ideally, it would be nice to wrap the text with an 80-character width and add two spaces after `</header>` and before `<hr>`, if it exists.

MacSteini:
The majority of the original German translation consisted of one-liners, so I stuck with that format, given that all the files are in HTML and I assumed CSS would take care of the formatting. I’m really sorry, but I won’t be trawling through all these files again anytime soon just to tweak them based on your very timely requirements.

Honestly, it feels a bit unfair to ask for these changes now, especially after you had all the information from my updates.

Also, there’s a lovely abundance of <tt> tags in the original English files – why don’t you have a go at updating those first? Once you’ve sifted through the hundreds of files, give me a shout, and I’ll be thrilled to follow suit.

iliajie:
> Honestly, it feels a bit unfair to ask for these changes now, especially after you had all the information from my updates.

Oh, I'm sorry, I didn't expect you to manually do that; I thought it was automated.

> Also, there’s a lovely abundance of <tt> tags in the original English files – why don’t you have a go at updating those first? Once you’ve sifted through the hundreds of files, give me a shout, and I’ll be thrilled to follow suit.

Certainly, but it has to be automated; nothing should be done manually. It would be a nightmare otherwise.

Thank you very much for your contribution! Merging now...

PR #2428, opened 2025-03-02
Fix to flush and sync the temp and new files

iliajie:
Hey Jamie!

This PR should fix the issue with occasional config file corruption.

### DIFF
--- a/web-lib-funcs.pl
+++ b/web-lib-funcs.pl
@@ -10758,6 +10758,11 @@ my $notempty = $_[1];
 if (defined($file = $main::open_temphandles{$fh})) {
 	# Closing a handle
 	my $noerror = $main::open_tempfiles_noerror{$file};
+	# Flush and sync before closing the handle
+	if (defined($fh)) {
+		$fh->flush();
+		$fh->sync();
+		}
 	if (!close($fh)) {
 		if ($noerror) { return 0; }
 		else { &error(&text("efileclose", $file, $!)); }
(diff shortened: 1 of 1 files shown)


web-lib-funcs.pl
```diff
@@ -10790,10 +10791,22 @@ sub close_tempfile
 		if ($noerror) { return 0; }
 		else { &error("Temporary file @{[html_escape($main::open_tempfiles{$_[0]})]} is empty!"); }
 		}
+	# Flush and sync the temp file before renaming
```
jcameron, line 10794: Why not do the flush just before the close() call on line 10760 ?

web-lib-funcs.pl
```diff
@@ -15,6 +15,7 @@ =head1 web-lib-funcs.pl
 ##use warnings;
 use Socket;
 use POSIX;
+use IO::Handle;
```
jcameron, line 18: Is this module always available?

iliajie: Yes.

jcameron: Ok and just including this allows us to call sync() on any file handle?

web-lib-funcs.pl
iliajie: > Why not do the flush just before the close() call on line 10760 ?

If you only flush at the initial file close and before creating the temporary file, the “vulnerability” would still exist during the `rename` operation.

So, we make sure that the temporary file’s content is fully written to disk, then perform the file replacement, and finally flush and sync the new file...

web-lib-funcs.pl
iliajie: > Ok and just including this allows us to call sync() on any file handle?

Yes, in modern Perl, **lexical filehandles** created with `open` are _automatically_ blessed into a class that inherits from `IO::Handle`...

jcameron: I see!

web-lib-funcs.pl
jcameron: A rename call doesn't re-write the while though, it just updates the file pointer. So there's no way a rename could truncate a file.

Please try moving this up to just before the close.

jcameron: The filesystem blocks used for file contents is different form the one uses for directory entries, so it's possible that we could use the blocks containing data even if the rename happens later.

Anyway, since you have a test environment that can repro this, please try it my way and let me know how it goes.

iliajie: > Anyway, since you have a test environment that can repro this, please try it my way and let me know how it goes.

I've already tried it your way, and it seems to work just fine! However, just because something works doesn't mean it's right—moving the flush and sync up to just before close for a file handle **wouldn’t address** the buffering problem because it doesn’t guarantee that the buffered data is physically written to disk before the rename occurs, right!?!?

iliajie: > Anyway, since you have a test environment that can repro this, please try it my way and let me know how it goes.

We can try this https://github.com/webmin/webmin/pull/2428/commits/0943289abe4ede83c4489962aa3c75f6af7286ab, as you suggested—however, I'm not sure if it's a better approach!?

jcameron: Well sure it does, because the temp file is written to, flushed and closed before the rename happens. The rename isn't done before the close, it is done after.

iliajie: > Well sure it does, because the temp file is written to, flushed and closed before the rename happens.

Then we're good. Thanks!

PR #2435, opened 2025-03-16
We don't have to build debs on Debian

swelljoe:
A deb package can be built anywhere with the right tools available, it doesn't need to be on Debian.


iliajie:
That’s right! Although I’ve done the exact opposite—I’ve always built on Debian for both.

Issue #2436, opened 2025-03-17
Dovecot setting obsolete


Deepcuts:
Webmin 2.303 on Debian 12

After setting "Accept SSL Connections" to Yes and restarting the service I see this warning with systemctl status dovecot:

`[Obsolete setting](config: Warning: Obsolete setting in /etc/dovecot/dovecot.conf:277: protocol { ssl_listen } has been replaced by service { inet_listener { address } })`

![Image](https://github.com/user-attachments/assets/26e99c41-167e-4121-9061-8df41f99998c)

jcameron:
Thanks for letting us know about this! What Dovecot version are you running there?

Deepcuts:
Dovecot version 2.3.19.1 (9b53102964)

iliajie:
Jamie, in Dovecot version 2.0 and up, `inet_listener` should be used instead of `ssl_listen`.

iliajie:
...but it's not a drop-in replacement—considering the complexity, we should just drop the "Interfaces for SSL connections" option entirely, and use only the "Listen on" option.

The appropriate fix in this PR https://github.com/webmin/webmin/pull/2437.

PR #2437, opened 2025-03-18
Add support to configure listen for any type of address

iliajie:
Hello Jamie!

This PR addresses the issue described in #2436.

The goal is to avoid using either the deprecated `ssl_listen` or the new `inet_listener` separately because of the higher complexity. Instead, we will just use `listen` directive, allowing it to handle both IPv4 and IPv6 or bind to specific IPs if needed. The ranges `127.0.0.1, ::` and `fe80::21c:42ff:fea2:f273, *` are also now supported.

image

### DIFF
--- a/dovecot/edit_net.cgi
+++ b/dovecot/edit_net.cgi
@@ -34,19 +34,24 @@ print &ui_table_row($text{'net_ssl_disable'},
 
 @listens = &find("imap_listen", $conf, 2) ?
 		("imap_listen", "pop3_listen", "imaps_listen", "pop3s_listen") :
-		("listen", "ssl_listen");
+		("listen");
 foreach $l (@listens) {
 	$listen = &find_value($l, $conf);
 	$mode = !$listen ? 0 :
-	        $listen eq "[::]" ? 1 :
+		# All interfaces, put in any order, e.g. "[::], *" or "*, ::"
+	        $listen =~ /^(\*|::|\[::\]),\s*(\*|::|\[::\])$/ ? 1 :
+		# IPv6 only, e.g. "[::]" or "::"
+	        $listen =~ /^(?:\:\:|\[\:\:\])$/ ? 4 :
+		# IPv4 only, e.g. "*"
 	        $listen eq "*" ? 2 : 3,
 	print &ui_table_row($text{'net_'.$l},
 		&ui_radio($l."_mode", $mode,
 			  [ [ 0, $text{'net_listen0'} ],
 			    [ 1, $text{'net_listen1'} ],
 			    [ 2, $text{'net_listen2'} ],
+			    [ 4, $text{'net_listen4'} ],
 			    [ 3, $text{'net_listen3'} ] ])."\n".
-		&ui_textbox($l, $mode == 3 ? $listen : "", 20), 3);
+		&ui_textbox($l, $mode == 3 ? $listen : "", 40), 3);
 	}
 
 print &ui_table_end();
--- a/dovecot/edit_net.cgi
+++ b/dovecot/edit_net.cgi
@@ -32,24 +32,19 @@ print &ui_table_row($text{'net_ssl_disable'},
 		      [ @opts,
 			[ "", &getdef($sslopt, \@opts) ] ]));
 
-@listens = &find("imap_listen", $conf, 2) ?
+my @listens = &find("imap_listen", $conf, 2) ?
 		("imap_listen", "pop3_listen", "imaps_listen", "pop3s_listen") :
 		("listen");
+my $mode;
 foreach $l (@listens) {
-	$listen = &find_value($l, $conf);
-	$mode = !$listen ? 0 :
+	my $v = &find_value($l, $conf);
+	$mode = !$v ? 0 :
 		# All interfaces, put in any order, e.g. "[::], *" or "*, ::"
-	        $listen =~ /^
-		    (?!(\*,\s*\*)        # Disallow *, *
-		    | (\[::\],\s*::)     # Disallow [::], ::
-		    | (::,\s*\[::\]))    # Disallow ::, [::]                    
-		    (\*|::|\[::\]),\s*   # Match *, ::, or [::]
-		    (\*|::|\[::\])$      # Followed by another *, ::, or [::]
-		/x ? 1 :
+		$v =~ /^\*,\s*(::|\[::\])$/ || $v =~ /^(::|\[::\]),\s*\*$/ ? 1 :
 		# IPv6 only, e.g. "[::]" or "::"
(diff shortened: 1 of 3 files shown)


jcameron:
So if we don't set the `ssl_listen` directive anymore, does this mean that Dovecot can no longer be configured to accept SSL connections?

Also, what is all the extra code in save_net.cgi doing for more 3? Do we allow multiple addresses to be entered now?

iliajie:
> So if we don't set the `ssl_listen` directive anymore, does this mean that Dovecot can no longer be configured to accept SSL connections?

No, not setting a separate `ssl_listen` directive doesn’t mean SSL connections are disabled. By default, if we don’t set a separate listener for SSL, Dovecot will use the addresses specified in the `listen` directive ("Listen on" option in UI)  for both plain and SSL connections.

> Also, what is all the extra code in save_net.cgi doing for more 3? Do we allow multiple addresses to be entered now?

There is currently a bug. The `::` isn't accurately representing both IPv4 and IPv6, as it was previously set. The `listen` directive can be configured to:

```
listen = *                        # IPv4 only
listen = ::                       # IPv6 only
listen = *, ::                    # IPv4 and IPv6
listen = 1.2.3.4, 2.3.4.5, ::     # Various address
```

If no `listen` is set, i.e., the "Default" option in the UI, it will default to `*, ::`, which means it will listen on all IPv4 and IPv6 interfaces.

dovecot/edit_net.cgi
```diff
@@ -34,19 +34,24 @@ print &ui_table_row($text{'net_ssl_disable'},
...
-		("listen", "ssl_listen");
+		("listen");
 foreach $l (@listens) {
 	$listen = &find_value($l, $conf);
 	$mode = !$listen ? 0 :
-	        $listen eq "[::]" ? 1 :
+		# All interfaces, put in any order, e.g. "[::], *" or "*, ::"
+	        $listen =~ /^(\*|::|\[::\]),\s*(\*|::|\[::\])$/ ? 1 :
+		# IPv6 only, e.g. "[::]" or "::"
+	        $listen =~ /^(?:\:\:|\[\:\:\])$/ ? 4 :
```
jcameron, line 44: Just use `eq` here instead of a regexp

dovecot/edit_net.cgi
```diff
@@ -34,19 +34,24 @@ print &ui_table_row($text{'net_ssl_disable'},
...
 @listens = &find("imap_listen", $conf, 2) ?
 		("imap_listen", "pop3_listen", "imaps_listen", "pop3s_listen") :
-		("listen", "ssl_listen");
+		("listen");
 foreach $l (@listens) {
 	$listen = &find_value($l, $conf);
 	$mode = !$listen ? 0 :
-	        $listen eq "[::]" ? 1 :
+		# All interfaces, put in any order, e.g. "[::], *" or "*, ::"
+	        $listen =~ /^(\*|::|\[::\]),\s*(\*|::|\[::\])$/ ? 1 :
```
jcameron, line 42: Split this into two separate regexps for readability. Current code would match `*, *`

dovecot/save_net.cgi
```diff
@@ -12,20 +12,50 @@ $sslopt = &find("ssl_disable", $conf, 2) ? "ssl_disable" : "ssl";
...
+				}
+
+			# Add IP address to list
+			push(@ips_valid, $ip);
+
+			# Validate against wildcards
+			&error(&text("net_ealisten_invalid_mix", $ip, "*"))
+				if ($has_ip4_wildcard && &check_ipaddress($ip));
+			&error(&text("net_ealisten_invalid_mix", $ip, "::"))
+				if ($has_ip6_wildcard && &check_ip6address($ip));
```
jcameron, line 56: What case are these wildcard checks looking for?

dovecot/edit_net.cgi
iliajie: Good suggestion, thanks! Fixed in the latest commit!

dovecot/edit_net.cgi
iliajie: There is no need to split into separate regexps; there is a better way. Fixed it now as well.

jcameron: Please no, don't create really complicated regexps. Remember, readability is way way more important that conciseness! Better to expand out the options like : 

```
$listen =~ /^\*,\s*(::|\[::\])$/ ? 1 : 
$listen =~ /^(::|\[::\]),\s*\*$/ ? 1 : 
```

dovecot/save_net.cgi
iliajie: It's intended to prevent saving IPv4 or IPv6 addresses with same wildcard type, such as * and ::.

To be clear, it's not allowed to have `1.2.3.4, *` or `aaaa::bbbb::11, ::`, while it's okay to have `1.2.3.4, ::` or `aaaa::bbbb::11, *`.

dovecot/edit_net.cgi
iliajie: You probably missed my previous update—it is as **clear and simple** as it gets already:

image

And, the `/x` extended regex mode has been supported in Perl since Perl 5.005, which was released in 1998.

I believe it’s much better to use multiple regex patterns on separate lines **with comments**—it makes everything clearer and more readable.

jcameron: Sorry, but this is not at all readable! You're using a negative regexp as a kind of poor-man's if statement to disalllow a few possible matches. Someone looking at this code is going to be super confused as to what's going on, compare to a few `if` or `? :` statements. 

Regular expressions are cool, but we shouldn't use them for complex conditional matches because they are hard to read. Also, the solution I suggested is just two lines compared to like 6 lines here!

iliajie: > Sorry, but this is not at all readable! You're using a negative regexp as a kind of poor-man's if statement to disalllow a few possible matches.

😄 Alright! Now, to be honest, I completely agree! Your solution is much cleaner, and in fact it can even be one line—fixed in the latest commit.

However, I’d still like to emphasize that we shouldn’t avoid using **`/x`** combined with **named capture groups** for complex regexps, for the same reason: readability.

jcameron: Well yes /x is nice, but my general rule is that if I find myself using a complex regexp, I need to re-think my code!

dovecot/save_net.cgi
jcameron: Ah ok makes sense

dovecot/edit_net.cgi
iliajie: > Well yes /x is nice, but my general rule is that if I find myself using a complex regexp, I need to re-think my code!

Certainly, but there are cases where complexity is unavoidable.

PR #2440, opened 2025-03-20
Add support for replacements in help link via URL

iliajie:
Hello, Jamie!

This PR adds support for replacements in help link via URL.

This allows dynamic replacements in help body, such as WordPress to ClassicPress or MySQL to MariaDB.

Usage:

The 5th parameter must be an array reference of hash references, e.g.:

```
[ { 'WordPress' => 'ClassicPress' } ]
```

Why?

This is useful when `help_pre_load` cannot be used during a normal call, such as in WP Workbench, where help links need to be generated via an underlying API (e.g., Virtualmin script’s API) rather than during the module call itself.

### DIFF
--- a/help.cgi
+++ b/help.cgi
@@ -11,6 +11,12 @@ $ENV{'PATH_INFO'} !~ /[\\\&\;\`\'\"\|\*\?\~\<\>\^\(\)\[\]\{\}\$\n\r]/ ||
 	&error($text{'help_epath'});
 $ENV{'PATH_INFO'} =~ /^\/(\S+)\/(\S+)$/ || &error($text{'help_epath'});
 $module = $1; $file = $2;
+for (split(/&/, $ENV{'QUERY_STRING'})) {
+	my ($k, $v) = map { &html_strip(&un_urlize($_)) } split(/=/, $_, 2);
+	if ($k eq "replace_what") { $what = $v }
+	elsif ($k eq "replace_with" && defined $what) {
+		push(@help_replacements, { $what => $v }); undef($what); }
+	}
 
 # if it ends with .gif assume it is a direct URL
 if ($file =~ /\.(gif|jpg|jpeg|png)$/i) {
@@ -18,7 +24,7 @@ if ($file =~ /\.(gif|jpg|jpeg|png)$/i) {
 	exit;
 }
 
-# read the help file
+# Read the help file
 $help = &read_help_file($module, $file);
 $help || &helperror(&text('help_efile3',
 		&html_escape($file), &html_escape($module)));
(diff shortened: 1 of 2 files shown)


jcameron:
Can you link to where in the code a help link is generated that could make use of this?

It seems like a rather complex solution versus just having two separate help pages.

iliajie:
> Can you link to where in the code a help link is generated that could make use of this?

Yes, here https://github.com/virtualmin/virtualmin-wp-workbench/blob/main/wordpress-extension.pl#L11-L15.

jcameron:
Why not just link to different help pages depending on whether Wordpress or Classicpress is being used, assuming that you want to change the name of the product?

Alternately, we could support a proper kind of templating in help pages, sort of like we do with `$DOM` etc in Virtualmin? And then drive that via parameters ... this would be a lot cleaner than this search-and-replace approach.

iliajie:
> Why not just link to different help pages depending on whether Wordpress or Classicpress is being used, assuming that you want to change the name of the product?

Since ClassicPress is a clone, having separate help pages that are nearly identical just under a different name feels like must more work than needed.

> Alternately, we could support a proper kind of templating in help pages, sort of like we do with `$DOM` etc in Virtualmin? And then drive that via parameters ... this would be a lot cleaner than this search-and-replace approach.

This could be feasible, but I’m not convinced it’s necessary. We already manage similar replacements for MySQL/MariaDB with language files, help files, and in `help_pre_load`. We’ve never needed templating for help files before, and introducing it now seems illogical.

iliajie:
I have looked further into it, and it won't be possible to use templating either, as it's called from a different process. The only solution is the one in this PR, or to use a temporary file to read replacements from, which likely isn't a better solution. Therefore, I'd prefer to keep this implementation, as it's clear and there are simply no other better options.

If you have any suggestions, please let me know.

jcameron:
Let me see if I can implement an example so you can see what I mean. But certainly this approach is not acceptable - if we want to allow substitutions into help pages, it needs to be done properly and not via search and replace.

iliajie:
Thanks, great! I’d be interested to see a better implementation considering the problem we’re having. Looking forward to.

jcameron:
OK check out this change : https://github.com/webmin/webmin/commit/e60be7133ba39acaef17ae4c6a02fa58d3c1d1e9

iliajie:
Oh, you meant it’s still okay to use URL params—ah, then it’ll work just perfectly fine! When you mentioned templates earlier, I thought you meant server-side templating, like we’ve always done—not using URL params.

Alright, this actually fits our case even better, and as a bonus, `help.cgi` will also support server-side replacements via the `help_template` function.

Thanks—and I totally agree now, this is a _better_ implementation!

jcameron:
Excellent thanks! I may use this in other modules like MySQL, to save use from the current hack of doing a string substitution from "MySQL" to "MariaDB"

iliajie:
Yes, that would be much cleaner! Thank you, Jamie!

PR #2447, opened 2025-04-03
Fix to respect option to copy new key and certificate to Webmin

iliajie:
G'day Jamie!

There was a [bug report](https://forum.virtualmin.com/t/letsencrypt-automatic-certificate-advice/132891) on the Virtualmin Forums—I checked it out and confirmed the issue.

This fix addresses the problem where the "Copy new key and certificate to Webmin" option isn’t being respected.

### DIFF
--- a/webmin/edit_ssl.cgi
+++ b/webmin/edit_ssl.cgi
@@ -326,7 +326,7 @@ else {
 
 	# Install in Webmin now?
 	print &ui_table_row($text{'ssl_usewebmin'},
-		&ui_yesno_radio("use", !!($config{'letsencrypt_use'} // 1)));
+		&ui_yesno_radio("use", !$config{'letsencrypt_nouse'}));
 
 	# SSL key size
 	print &ui_table_row($text{'ssl_size'},
--- a/webmin/edit_ssl.cgi
+++ b/webmin/edit_ssl.cgi
@@ -326,7 +326,7 @@ else {
 
 	# Install in Webmin now?
 	print &ui_table_row($text{'ssl_usewebmin'},
-		&ui_yesno_radio("use", 1));
+		&ui_yesno_radio("use", !!($config{'letsencrypt_use'} // 1)));
 
 	# SSL key size
 	print &ui_table_row($text{'ssl_size'},
--- a/webmin/letsencrypt.cgi
+++ b/webmin/letsencrypt.cgi
@@ -88,7 +88,8 @@ else {
 
 if ($in{'save'}) {
 	# Just update renewal
-	&save_renewal_only(\@doms, $webroot, $mode, $size, $in{'subset'});
+	&save_renewal_only(\@doms, $webroot, $mode, $size,
+			   $in{'subset'}, $in{'use'});
 	&redirect("edit_ssl.cgi");
 	}
 else {
@@ -110,8 +111,12 @@ else {
 		# Worked, now copy to Webmin
 		print $text{'letsencrypt_done'},"<p>\n";
 
+		# Save the renewal schedule
+		&save_renewal_only(\@doms, $webroot, $mode,
+				   $size, $in{'subset'}, $in{'use'});
+
+		# Copy cert, key and chain to Webmin
 		if ($in{'use'}) {
-			# Copy cert, key and chain to Webmin
 			print $text{'letsencrypt_webmin'},"<br>\n";
 			&lock_file($ENV{'MINISERV_CONFIG'});
 			&get_miniserv_config(\%miniserv);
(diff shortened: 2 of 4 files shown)


webmin/edit_ssl.cgi
```diff
@@ -326,7 +326,7 @@ else {
 
 	# Install in Webmin now?
 	print &ui_table_row($text{'ssl_usewebmin'},
-		&ui_yesno_radio("use", 1));
+		&ui_yesno_radio("use", !!($config{'letsencrypt_use'} // 1)));
```
jcameron, line 329: What does `!!($config{'letsencrypt_use'} // 1)` do exactly?  I would suggest replacing this with a more readable expression.

iliajie: I’d rather keep it! It’s clean, short, and simple—and it always guarantees a return value of either `1` or `0`.

jcameron: Sorry but I find it unreadable. Why not just something like `$config{'letsencrypt_use'} ? 1 : 0`

jcameron: Also, there is no config option called `letsencrypt_use`  in `webmin/config.info`

iliajie: > Sorry but I find it unreadable. Why not just something like `$config{'letsencrypt_use'} ? 1 : 0`

We **cannot** change this because we want to **preserve** the original behavior! The test specifically checks whether `$config{'letsencrypt_use'}` is not defined, which is why it uses `$config{'letsencrypt_use'} // 1`. The `!!` will always turn any value into strict `1` or `0`.

If written differently, it would look like this:

```
defined($config{'letsencrypt_use'}) ? $config{'letsencrypt_use'} ? 1 : 0 : 1
```

jcameron: Oh so this is a new internal setting that we're saving? In that case, why make things complicated by having undefined mean "1" ? Why not instead just invert the logic so that undefined or "0" means "0", and "1" means "1" ? That's what Virtualmin does all all other similar cases to this..

iliajie: > Oh so this is a new internal setting that we're saving?

Yes.

> Why make things complicated by having undefined mean "1" ? Why not instead just invert the logic so that undefined or "0" means "0", and "1" means "1" ?

We cannot do this because the old default was `1`, which essentially means `undefined`. If we make this change, it will break existing systems. I have thought it all through, so no worries. 🙂

jcameron: What I mean is, just called the config option something like `letsencrypt_nouse` and then the code can be as simple as `&ui_yesno_radio("use", !$config{'letsencrypt_nouse'})`

iliajie: Well, sure! We could do that too, if you like it more. Here is the patch https://github.com/webmin/webmin/pull/2447/commits/2760336b40a8c401893559523f96449ab156faf1.

webmin/letsencrypt.cgi
```diff
@@ -141,38 +146,33 @@ else {
...
 sub save_renewal_only
 {
-my ($doms, $webroot, $mode, $size, $subset) = @_;
+my ($doms, $webroot, $mode, $size, $subset, $usewebmin) = @_;
 $config{'letsencrypt_doms'} = join(" ", @$doms);
 $config{'letsencrypt_webroot'} = $webroot;
 $config{'letsencrypt_mode'} = $mode;
 $config{'letsencrypt_size'} = $size;
 $config{'letsencrypt_subset'} = $subset;
+$config{'letsencrypt_nouse'} = !$usewebmin ? 1 : 0;
```
jcameron, line 175: Why not just `$usewebmin ? 0 : 1;`

webmin/edit_ssl.cgi
jcameron: Excellent, that's much cleaner!

webmin/letsencrypt.cgi
iliajie: I don't know, why not? 🙂 ... I didn't go with `$usewebmin ^ 1;` or `(!$usewebmin) + 0;` or even `1 - !!$usewebmin`... So, what difference does it really make? But sure, let’s write it the way you prefer. 

Though honestly, generally speaking, I personally really don't like double negations tied to a _variable name_. It's hard to read, especially when it's used as `$nothis = 1` and later checked as `!$nothis`...

iliajie: > Because clearly the code !$usewebmin ? 1 : 0 is more complex than $usewebmin ? 0 : 1 . I really am having trouble understanding why this isn't clear?????!!!!!! :-(

Jamie, my apologies—I didn’t mean to be annoying! And you’re right, your example is simpler! I’ll make sure to pay closer attention to keeping the Perl code as Perlish and simple as possible in the future.

> The code as written is negating the variable only to perform a boolean comparison on it!

What Perl is doing with `$usewebmin = 0` and `$usewebmin = "0"` is something other languages don’t do. In Perl, both a number `0` and a string `"0"` evaluate to false. In JavaScript or Python, the `"0"` will always be a string and evaluate to `true`. You're absolutely right that, for a Perl expert, this code may seem odd, but not everyone is a Perl expert, and using negation can actually hint that we expect a truthy value.

What I personally like is that in a programming language like Go, it would result in a compilation error, as automatic type coercion is not allowed at all but with scripting languages, the story is different, and they behave the way they do.

Nevertheless, I am more aware of your expectations and will do my best to meet them.

jcameron: Oh yeah for sure if $usewebmin was some odd string this might make sense, but we control the value so we can just use simple code.

iliajie: > Oh yeah for sure if $usewebmin was some odd string this might make sense, but we control the value so we can just use simple code.

Got it, thanks!

Issue #2459, opened 2025-04-21
Webmin action email says invalid password when logged in successfully


brianbaggs35:
Hello,

I have emails set up for webmin actions and every time I log in to webmin I get the email that says I logged in successfully, but I also get two other emails that say invalid password and invalid two-factor token, even though I logged in successfully the first time.

Thank you

jcameron:
Can you post the contents of one of these emails, or a screenshot of it?

jcameron:
Odd that it reports a two-factor login failure. Are you using two-factor for this account?

Flo-R1der:
Hi, I have the same issue for a while now on two machines running Webmin 2.303. Today I investigated a little and figured this out: 

```mermaid
flowchart TD
    A(**Sign in**
    with username/password) 
    C(**Email:**
    Invalid twofactor token)
    A --> C
    D(**Email:**
    Invalid password)
    C --> D
    B(**Enter TOTP**)
    A ===> |correct credentials:
    loading 2FA form| B
    E(**Successfull login**)
    B ===> |correct 2FA:
    loading Dashboard| E
    F(**Email:**
    Logged into Webmin)
    B --> F
```

I think the first two mails after entering the username/password should not be sent out. Maybe there is some logic like: 
- if ``login != successful`` => send mail for invalid password/2fa-token

Whereas the logic should be something like (one of the following):
- if ``login == failed`` => send mail for invalid password/2fa-token
- if ``credentials != correct`` => send mail for invalid password/2fa-token

jcameron:
Can you check if this still happens if you change the Webmin theme to the classic "Framed Theme", at Webmin -> Webmin Configuration -> Webmin Themes? I suspect it may be due to the way the newer theme checks if 2fa is needed..

iliajie:
> I suspect it may be due to the way the newer theme checks if 2fa is needed.

Yeah, this is the exact reason for this false positive message. I have been looking at `miniserv.pl`, and so far, it isn't clear exactly how to handle it properly. And, upon further review, I also don't like how `$in{'failed_twofactor_attempt'}` is used.


I think it needs to be reworked to properly support displaying the 2FA code on a separate page, even for older themes.

jcameron:
@iliajie so is the authentic theme attempting a login without 2fa, and only if that fails showing the 2fa form??

iliajie:
Well, yes! And, we discussed it many years ago—because even if 2FA is enabled on the system it doesn’t mean all users are using it.

[...]

PR #2462, opened 2025-04-27
Add ability to make authentication in two steps

iliaross:
Hey Jamie!

As we talked about in [this thread](https://github.com/webmin/webmin/issues/2459#issuecomment-2832845920), we need to show the 2FA input field only for users who have 2FA turned on.

To make it work right, we have to do it in two steps, so we don’t trigger any false positive error messages printed to logs.

This patch is needed for SPA themes to show the 2FA field only after the user enters the right login info. The theme will handle it based on the server’s response.

At first, I got it working without any changes on the Webmin side, but it meant sending the username and password in the headers. Eventually, I gave up on that idea because some proxies could log headers and eventually leak passwords—and, it's just not right.

So, I made this quite well tested change. Please take a look; I'm still here for now.

### DIFF
--- a/miniserv.pl
+++ b/miniserv.pl
@@ -4276,8 +4276,10 @@ if ($ok && (!$expired ||
 	# Log in creds were OK but two-factor auth is still pending
 	if ($twof_probe) {
 		# Two-factor auth is required
-		$validated=$already_session_id=$authuser=$baseauthuser = undef;
-		$querystring=$method=$page=$request_uri=$logged_code = undef;
+		$validated = $already_session_id = undef;
+		$authuser = $baseauthuser = undef;
+		$querystring = $method = $page = $request_uri = undef;
+		$logged_code = undef;
 		$queryargs = "";
 		# Write response
 		&http_error(401, "Two-factor authentication is required");
--- a/miniserv.pl
+++ b/miniserv.pl
@@ -2101,12 +2101,8 @@ if (!$validated) {
 			$method = "GET";
 			$querystring .= "&failed=".&urlize($failed_user)
 				if ($failed_user);
-			if ($twofactor_msg) {
-				$querystring .= "&failed_save=".&urlize($failed_save);
-				$querystring .= "&failed_pass=".&urlize($failed_pass);
-				$querystring .= "&failed_twofactor_attempt=".&urlize($failed_twofactor_attempt);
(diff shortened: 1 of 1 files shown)


miniserv.pl
```diff
@@ -4271,6 +4273,17 @@ sub handle_login
 
 if ($ok && (!$expired ||
 	    $config{'passwd_mode'} == 1)) {
+	# Log in creds were OK but two-factor auth is still pending
+	if ($twof_probe) {
+		# Two-factor auth is required
+		$validated=$already_session_id=$authuser=$baseauthuser = undef;
```
jcameron, line 4279: Split these up on to separate lines

miniserv.pl
```diff
@@ -2101,12 +2106,8 @@ sub handle_request
 			$method = "GET";
 			$querystring .= "&failed=".&urlize($failed_user)
 				if ($failed_user);
-			if ($twofactor_msg) {
```
jcameron, line 2104: Is there a reason we don't need these kinds anymore?

jcameron:
Thanks! But can you explain what's going on internally here? Does the theme just make one HTTP request to probe if the user needs two factor, and then later sends the username, password and 2fa code?

miniserv.pl
iliajie: Done!

miniserv.pl
iliajie: Yeah, we don’t need to send the password back or keep track of failed attempts anymore, since the form is now submitted over an HTTP request without reloading the page and losing the password and other data.

iliajie:
> But can you explain what's going on internally here? Does the theme just make one HTTP request to probe if the user needs two factor, and then later sends the username, password and 2fa code?

Yep, that’s exactly how it works! First, we check the username and password. If the theme doesn’t use the new `twofprobe` flag, _miniserv_ will expect the two-factor code right away, like before. If the `twofprobe` flag is set, we end the current request and send a 401 response to the theme, letting it know the username and password were good, and now it’s time to provide the TOTP code. The page then flips, the user enters the code, and if it’s right, they’re logged in—if not, they’ll see an error...

iliajie:
Okay, done with the theme too! You can see how it works here:

https://github.com/iliajie/authentic-theme-src/blob/master/unauthenticated/js/session-login.src.js#L110

jcameron:
Ok thanks for the patch!

PR #2463, opened 2025-04-28
Update setup-repos.sh

wootje:
Corrected a mistake in the URL:
NEW_SCRIPT_URL="https://raw.githubusercontent.com/webmin/webmin/master/webmin-setup-repo.sh"

### DIFF
--- a/setup-repos.sh
+++ b/setup-repos.sh
@@ -3,8 +3,7 @@
 # Sets up a production or testing repository for Webmin and Usermin packages
 # on Debian-based and RPM-based systems
 
-NEW_SCRIPT_URL="https://raw.githubusercontent.com/webmin/webmin/master/\
-webmin-setup-repo.sh"
+NEW_SCRIPT_URL="https://raw.githubusercontent.com/webmin/webmin/master/webmin-setup-repo.sh"
 
 for downloader in "curl -fsSL" "wget -qO-"; do
 	if command -v "${downloader%% *}" >/dev/null 2>&1; then


wootje:
Because of a mistake in the URL the install script didn't work. I fixed the typo.

iliajie:
Hello,

Thanks for the PR, but this isn’t the right change. We do it that way on purpose to avoid lines longer than 80 characters, and it doesn’t actually break anything—a backslash `\` at the end of a line should escape the newline, so the shell treats it like one continuous line.

What shell are you using?

iliajie:
Something like this could work too:

```shell
URL_BASE="https://raw.githubusercontent.com"
URL_PATH="/webmin/webmin/master/webmin-setup-repo.sh"

NEW_SCRIPT_URL="${URL_BASE}${URL_PATH}"
```

wootje:
> Hello,
> 
> Thanks for the PR, but this isn’t the right change. We do it that way on purpose to avoid lines longer than 80 characters, and it doesn’t actually break anything—a backslash `\` at the end of a line should escape the newline, so the shell treats it like one continuous line.
> 
> What shell are you using?
Custom Armbian op an Android TV box with Linux firmware. It is able to install Debian & Ubuntu packages.

iliajie:
Can you update your patch to use:

```
URL_BASE="https://raw.githubusercontent.com"
URL_PATH="/webmin/webmin/master/webmin-setup-repo.sh"

NEW_SCRIPT_URL="${URL_BASE}${URL_PATH}"
```

And, note that you should actually use `webmin-setup-repo.sh` directly.

iliajie:
See this patch https://github.com/webmin/webmin/commit/a7676f40049cf997770421f51a750247dd0e196d.

PR #2469, opened 2025-05-08
Add ability for modules with no ACL to edit config

iliajie:
Hello Jamie!

This PR fixes an issue where previously discussed plugins don't have their own ACL and are controlled by an actual module like Virtualmin but still should be allowed to edit its configuration.

### DIFF
--- a/config.cgi
+++ b/config.cgi
@@ -10,7 +10,8 @@ require './config-lib.pl';
 $m = $in{'module'} || $ARGV[0];
 %module_info = &get_module_info($m);
 %module_info || &error($text{'config_emodule'});
-&foreign_available($m) || &error($text{'config_eaccess'});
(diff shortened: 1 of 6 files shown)


jcameron:
Are the uconfig.cgi changes needed? Those CGIs are only used for Usermin preferences, not global module configs.

iliajie:
> Are the uconfig.cgi changes needed? Those CGIs are only used for Usermin preferences, not global module configs.

You're right! It's not needed. Fixed now!

jcameron:
Thanks! But would this allow *any* user to edit the config for a module with `noacl` set?

iliajie:
Yes, any user without limits. Yet, such modules should implement preferences used. However, it must not affect security in any way.

PR #2477, opened 2025-05-19
Fix to simplify reset-password UI for sudo-capable users

iliaross:
Hey Jamie,

This is the improvement we discussed yesterday.


acl/forgot_form.cgi
```diff
@@ -13,27 +13,25 @@ our (%in, %text);
 &can_edit_user($in{'user'}) || &error($text{'edit_euser'});
 my $u = &get_user($in{'user'});
 $u || &error($text{'edit_egone'});
-
+my $is_admin = $u->{'name'} =~ /^(?:root|admin|sysadm)$/;
```
jcameron, line 16: Only the `root` user can be sudo'd to though. Also, hard-coding usernames in general is a bad idea.

jcameron: Right, but sudo isn't applicable to any users other than `root` ?

jcameron: Yes, but a sudo-capable user won't be able to login as "admin"

iliaross: Okay, I have reverted to using the previous test, and then later, as you mentioned, we will fix it properly, because neither tests are accurate enough!

jcameron: In this case it is definitively "root", per the code at https://github.com/webmin/webmin/blob/master/miniserv.pl#L3893

acl/lang/en
```diff
@@ -519,12 +519,10 @@ twofactor_ebutton=No button clicked!
 forgot_title=Send Password Reset Link
 forgot_err=Failed to send password reset link
 forgot_header=Password reset link details
-forgot_user=Reset password for Webmin user
+forgot_user=Reset password for Webmin or sudo-capable Unix user
```
jcameron, line 522: I don't like that it mentions "sudo-capable Unix user" here, as this message is misleading for the common case where only a Webmin password is being reset. Instead just make it "Reset password for user"

acl/forgot_form.cgi
iliaross: Okay… but why is this code even relevant? We’re using a Webmin user to generate a password reset link for another user. Any Webmin non-safe user with access to ACL and Users and Groups module should be able to do that, right?

acl/lang/en
iliaross: Agreed! Fixed now!

PR #2481, opened 2025-05-21
Fix to make forgot links compatible across themes

iliaross:
Hey Jamie!

This PR fixes an issue where the “forgot password” link is generated using Authentic, but the global default theme is Framed.


jcameron:
Can you explain further why this is needed? Like what are the specific URLs being generated by the themes?

iliaross:
> Can you explain further why this is needed? Like what are the specific URLs being generated by the themes?

Yes, sure—Authentic, being a SPA theme, generates forgot password links with a `?forgot=` parameter instead of the actual `forgot.cgi` page. This PR makes those links work across themes, even if the reset link was created in one theme and the global default is another.

jcameron:
Ok makes sense!

PR #2483, opened 2025-05-25
Fix issues in new PHP package manager

iliaross:
Hey Jamie!

This PR fixes issues with installing and listing PHP packages across different systems—EL, Debian, and FreeBSD. On Linux, everything revolves around the `-common` package. Debian doesn’t have a `-runtime` package at all, so things are currently broken there. FreeBSD doesn’t have a `php-common` package and expects just `php`, while on Linux, installing `php` pulls in `mod_php`.

This PR gets everything working correctly across those systems, though the implementation could likely be improved.

### DIFF
--- a/phpini/list_pkgs.cgi
+++ b/phpini/list_pkgs.cgi
@@ -50,8 +50,7 @@ if (@pkgs) {
 else {
 	print "<b>$text{'pkgs_none'}</b> <p>\n";
 	}
-
-my @newpkgs = grep { !$got{$_->{'name'}} } &list_any_available_php_packages();
+my @newpkgs = grep { !$got{$_->{'name'}} } &list_best_available_php_packages();
 if (@newpkgs && &foreign_installed("package-updates")) {
 	# Show form to install a new version
 	print &ui_hr();
--- a/phpini/list_pkgs.cgi
+++ b/phpini/list_pkgs.cgi
@@ -49,24 +49,27 @@ if (@pkgs) {
 else {
 	print "<b>$text{'pkgs_none'}</b> <p>\n";
 	}
-my @newpkgs = grep { !$got{$_->{'phpver'}} } &list_best_available_php_packages();
-if (@newpkgs && &foreign_installed("package-updates")) {
+if (&foreign_installed("package-updates")) {
+	my @newpkgs = grep { !$got{$_->{'phpver'}} }
+		&list_best_available_php_packages();
 	# Show form to install a new version
-	print &ui_hr();
-	print &ui_form_start(
-		&get_webprefix()."/package-updates/update.cgi", "post");
-	print "$text{'pkgs_newver'}&nbsp;\n";
-	&extend_installable_php_packages(\@newpkgs);
-	# Largest version on top
-	@newpkgs = sort { $b->{'phpver'} cmp $a->{'phpver'} } @newpkgs;
-	print &ui_select("u", undef,
-		[ map { [ $_->{'name'},
-			  "PHP $_->{'shortver'}" ] } @newpkgs ]);
-	print &ui_hidden(
-		"redir", &get_webprefix()."/$module_name/list_pkgs.cgi");
-	print &ui_hidden("redirdesc", $text{'pkgs_title'});
-	print &ui_hidden("mode", "new");
-	print &ui_form_end([ [ undef, $text{'pkgs_install'} ] ]);
+	if (@newpkgs) {
+		print &ui_hr();
+		print &ui_form_start(
+			&get_webprefix()."/package-updates/update.cgi", "post");
+		print "$text{'pkgs_newver'}&nbsp;\n";
+		&extend_installable_php_packages(\@newpkgs);
+		# Largest version on top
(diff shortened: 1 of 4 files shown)


iliaross:
I’ve completely reworked my initial PR and tested it thoroughly for many hours on both Debian and EL systems. It also works on FreeBSD when the package name is just `php` and no `php-common` package is available. Dependencies are handled correctly on both install and remove, and all is displayed accurately too across all distros.

Example:

image

phpini/phpini-lib.pl
```diff
@@ -992,29 +1014,47 @@ sub list_all_php_version_packages
...
+# extend_installable_php_packages(&packages)
+# Given a list of PHP packages to install, extends them to include packages
+# that are also has to be installed, such as -cli or -fpm
+sub extend_installable_php_packages
+{
+my ($pkgs) = @_;
+my @extra = ('cli', 'fpm');
+foreach my $pkg (@$pkgs) {
+	if ($pkg->{'name'} =~ /-common$/) {
+		$pkg->{'name'} .= ' ' . join(' ',
```
jcameron, line 1037: I don't think modifying the `name` field to turn it into a list of package names is a clean approach.

phpini/phpini-lib.pl
```diff
@@ -919,14 +928,14 @@ sub list_php_base_packages
 sub list_all_php_module_packages
 {
 my ($base) = @_;
-$base =~ s/(-php|-runtime)$//;
+$base =~ s/-common$//;
```
jcameron, line 931: What if the base package name contains -php ?

phpini/phpini-lib.pl
iliaross: Well, what is the other better suggestion? We need to specifically extend it before making an install.

phpini/list_pkgs.cgi
```diff
@@ -40,24 +39,26 @@ if (@pkgs) {
...
 		}
 	print &ui_columns_end();
 	print &ui_form_end([ [ undef, $text{'pkgs_delete'} ] ]);
 	}
 else {
 	print "<b>$text{'pkgs_none'}</b> <p>\n";
 	}
-
-if (&foreign_installed("package-updates")) {
+my @newpkgs = grep { !$got{$_->{'phpver'}} } &list_best_available_php_packages();
```
jcameron, line 52: list_best_available_php_packages calls list_available_php_packages which depends on the package-updates module, so it's not safe to call this before the foreign_installed check.

phpini/phpini-lib.pl
iliaross: > What if the base package name contains -php ?

Considering we are passing a package name only with `-common` like `php8.4-common` or `php-common` or `php83-php-common` by removing the last `-common` are are getting exactly the package name we want.

phpini/phpini-lib.pl
jcameron: I'm not sure, you wrote the code! :-) 

But in this package object, `name` is a field that just contains the package name. It's not reasonable to redefine it to be something else for convenience later in the code..

phpini/list_pkgs.cgi
iliaross: Got it! Fixed now!

phpini/phpini-lib.pl
iliaross: Okay, I simplified the code to avoid using the newer `/r` flag, to have a potential support for systems with Perl 5.10...

jcameron: Instead of modifying the list passed in, this function should return a new list contaiing whatever the caller needs. That's where it would make sense to list multiple packages..

Also, resist the temption put everything on one line line this. It's a lot more readable to use temporary variables.

iliaross: > Instead of modifying the list passed in, this function should return a new list contaiing whatever the caller needs. That's where it would make sense to list multiple packages..

Agreed! Thanks for the suggestion. Fixed here https://github.com/webmin/webmin/pull/2483/commits/92dc310ce0ee39c491f7bd931d9ed68b0b567af3.



> Also, resist the temptation put everything on one line line this. It's a lot more readable to use temporary variables.

Yeah, I like one-liners.

iliaross:
Anything else left here?

phpini/phpini-lib.pl
```diff
@@ -992,29 +1022,56 @@ sub list_all_php_version_packages
...
+sub extend_installable_php_packages
+{
+my ($pkgs) = @_;
+my @pkgs;
+my @extra = ('cli', 'fpm');
+foreach my $pkg (@{$pkgs}) {
+	my $p = { 'name' => $pkg->{'name'},
+		  'ver'  => $pkg->{'shortver'} };
+	if ($p->{'name'} =~ /-common$/) {
+		$p->{'name'} .= ' ' . join(' ',
```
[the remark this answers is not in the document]
iliaross: Because it’s easier to edit and add a new value to the list rather than have it done inline. Also, it’s easier to reuse this code later for something else. 

It’s okay!

PR #2486, opened 2025-05-31
Add support to reset password for Virtualmin mail users

iliaross:
Hello Jamie!

As discussed in PM, this PR adds support for resetting passwords for Virtualmin mail users.

### DIFF
--- a/forgot.cgi
+++ b/forgot.cgi
@@ -151,7 +151,15 @@ if (defined($in{'newpass'})) {
 		&reload_miniserv();
 		print $text{'forgot_done'},"<p>\n";
 		}
-	print &text('forgot_retry', '/'),"<p>\n";
+	
+	# Print link to login for Webmin user
+	if (!$muser) {
+		print &text('forgot_retry', &get_webprefix()."/",
+			    "<tt>".$username."</tt>"),"<p>\n";
+		}
+	else {
+		print &text('forgot_retry2', "<tt>".$username."</tt>"),"<p>\n";
+		}
 
 	&webmin_log("forgot", "reset", undef,
 		    { 'user' => $username,
--- a/lang/en
+++ b/lang/en
@@ -163,7 +163,7 @@ forgot_newpass=New password for user $1
 forgot_newpass2=New password again
 forgot_passok=Change Password
 forgot_euser2=Webmin user $1 does not exist!
-forgot_euser3=Mail user $1 does not exist!
+forgot_euser3=Mailbox user $1 does not exist!
 forgot_esync=The password cannot be changed for this user
 forgot_elock=The password for this user is locked
 forgot_err=Failed to reset password
@@ -172,7 +172,8 @@ forgot_enewpass2=New passwords do not match!
 forgot_equality=Password requirements were not met : $1
 forgot_wdoing=Changing password for user $1 ..
 forgot_done=.. password change complete.
-forgot_retry=You can now <a href='$1'>login to Webmin</a> using your new password.
+forgot_retry=The password for Webmin user $2 was reset successfully. You can now <a href='$1'>log in</a> using your new password.
+forgot_retry2=The password for mailbox user $1 was reset successfully.
 forgot_edisabled=The password for a disabled Virtualmin domain cannot be reset
 forgot_vdoing=Changing password for Virtualmin domain $1 ..
 forgot_udoing=Changing password for Unix user $1 ..


forgot.cgi
```diff
@@ -26,9 +26,33 @@ time() - $link{'time'} > 60*$timeout &&
...
 &foreign_require("acl");
 my ($wuser) = grep { $_->{'name'} eq $link{'user'} } &acl::list_users();
-$wuser || &error(&text('forgot_euser2',
+
+# Get the Virtualmin mail Unix user if Webmin user is not found
+my ($muser, $muserdom);
+if (!$wuser && $link{'muser'}) {
+	# Probably Virtualmin mail user, so try to find it
+	&foreign_require("virtual-server");
+	foreach my $d (&virtual_server::list_domains()) {
```
jcameron, line 35: This function should speed up finding the domain for a random username : https://github.com/virtualmin/virtualmin-gpl/blob/master/virtual-server-lib-funcs.pl#L12516

iliaross: Thanks for the heads-up! If there’s nothing else, could you go ahead and merge it and fix it? Otherwise, I’ll have to wait until tomorrow morning to handle it.

jcameron:
Thanks, this looks good! Now the catch is how we can link to it from Usermin ..

forgot.cgi
iliaross: > This function should speed up finding the domain for a random username : https://github.com/virtualmin/virtualmin-gpl/blob/master/virtual-server-lib-funcs.pl#L12516

Done! Thanks! Check this patch out https://github.com/webmin/webmin/commit/2d5768a6123fea5b7fb0ffa84f2e9aeabe721977.

iliaross:
> Now the catch is how we can link to it from Usermin ..

I’ve taken a close look and there’s no simple solution to this unless we re-implement the part that handles password resets for Unix users. Alternatively, we could show a button on the Usermin login screen, and when it’s clicked, display an error message saying that password reset is only supported in Webmin, along with a link to the Webmin login page but only if Webmin installed and running.

And as we discussed earlier, we should always show the password reset link/button in Webmin login screen if it’s enabled. I’ll send you a PR for that as well...

iliaross:
For the fun of it, I could hack into Virtualmin from Usermin using this code:

```perl
{
    my $manage_symlinks = sub {
        my ($action) = @_;
        my @modules = qw(virtual-server postfix useradmin);
        foreach my $module (@modules) {
            my $src = "$root_directory/../webmin/$module";
            my $dst = "$root_directory/$module";

            if ($action eq 'create') {
                return 1 unless(-d $src);
                symlink($src, $dst) unless( -l $dst || -e $dst);
                }
            elsif ($action eq 'remove') {
                unlink($dst) if (-l $dst);
                }
            }
        };

    # Create temp symlinks for Webmin modules or exit the function if not available
    return if ($manage_symlinks->('create'));

    # Change DOCUMENT_ROOT to point to the Webmin directory local to this block
    local $ENV{'WEBMIN_CONFIG'} = $ENV{'WEBMIN_CONFIG'} =~ s|/usermin|/webmin|r;

    # Get user information
    my $user = 'ilia@ubuntu24-gpl.virtualmin.dev';
    my $muser;
    eval {
        local $main::error_must_die = 1;
        &foreign_require("virtual-server");
            my $d = &virtual_server::get_user_domain(lc($user));
            if ($d) {
                my @u = &virtual_server::list_domain_users($d, 0, 0, 1, 1, 0);
                ($muser) = grep { $_->{'user'} eq lc($user) } @u;
                }
        };

    return if ($@ || !$muser);
    
    # Remove temp symlinks for webmin modules
    $manage_symlinks->('remove');

    # Your user object is here
    # $muser;
    var_dump($muser);
}
```

...and, it worked! 😄

iliaross:
Still, I don’t think we need anything on the Usermin login page since password reset is already built into Usermin. I don’t really see the point in spending time to add it there. Agreed?

jcameron:
What do you mean by it being already built in?

iliaross:
Usermin already has a change password module, and the mail user still needs access to the mailbox to reset the password—whether it’s through that module or a reset option on the login page if we implemented it. The only real gain from adding it to the login page would be using a recovery email. But even that can already be done through the Webmin login page. So, do you think it’s really worth adding it to Usermin’s login page separately?

jcameron:
I was thinking that on the login page in Usermin, we could have a button that links to the forgot password page in Webmin on the same system (if installed).

iliaross:
I could add that tomorrow to the Usermin login page. Meanwhile, do you think you could add an API on Webmin side for Usermin to run such check?

jcameron:
I don't think we need an API, we could just have Usermin check the contents of `/etc/webmin/config` for `forgot_pass=1`

iliaross:
What if Webmin is installed elsewhere? Or if it isn’t running? And, we would need to construct the Webmin link properly as well… right?

jcameron:
Check out my latest commits and let me know what you think..

iliaross:
Alright, good! I've added a couple more small patches too.

However, I will need to take a closer look at how to make it work with the Authentic theme. I will send you a PR when done slightly later today.

iliaross:
By the way, I forgot to mention that the latest patch to `session_login.cgi` is missing a webprefix. Also, I suggest moving this logic into a separate function so it can be reused—it really needs to be reused.

iliaross:
> Also, I suggest moving this logic into a separate function so it can be reused—it really needs to be reused.

Probably not really needed!

> By the way, I forgot to mention that the latest patch to `session_login.cgi` is missing a webprefix.

I fixed that!

> I will send you a PR when done slightly later today.

Here, please check this PR out https://github.com/webmin/usermin/pull/127.

Issue #2504, opened 2025-06-24
miniserv http to https redirect not working on RHEL9


verne-work:
I have become very lazy and often type in http://example.com:10000 or not even use a prefix, and in the past have been very happy with miniserv either telling me to use HTTPS or even executing the redirect for me.  But on my new test RHEL9 server its not working.   Note a manual/literal HTTPS connection works fine.

I am comparing miniserv 2.202 on RHEL8 to 2.203 on RHEL9

I believe the line 

 `$more = Net::SSLeay::read($ssl_con);`

in subroutine `read_line `seems to behave differently for me, from debug print statements I added in that area.

I see my v8 box has Net::SSLeay 1.88 while my v9 box has 1.94 -- perhaps the READ function has changed?

on my older server, the read executes immediately but does not have any data (returns an empty string) and goes on to tell me what to do; on the new server it hangs and times out from the higher level code having called it with ALARM(10).

I am testing this using curl and lynx, and both work OK on my older server -- perhaps my tests are not valid for RHEL9 ?

jcameron:
Hmm, maybe `Net::SSLeay::read` did change it's behavior?

What happens in your browser when you try to access the `http://` URL? Does it just hang, or show some error message?

verne-work:
Chrome says after about 10 seconds,   

```
this page isn't working
media-wvu9.mgw.wvnet.edu didn’t send any data
ERR_EMPTY_RESPONSE
```
but it keeps trying (the little circle keeps rotating on the top left) for approximately another 10 seconds before giving up.  Once it gives up, the message above remains on the screen -- ie, there is no additional message.

with Firefox, it trys for 1:40 (100 seconds) and then displays



```
The connection was reset

The connection to the server was reset while the page was loading.

-     The site could be temporarily unavailable or too busy. Try again in a few moments.
-     If you are unable to load any pages, check your computer’s network connection.
-     If your computer or network is protected by a firewall or proxy, make sure that Firefox is permitted to access the web.

```

jcameron:
OK let me see if I can re-produce this. Have you seen it on other similar distros, like Rocky 9 ?

verne-work:
I have only tested it on RHEL9 at work.

I think I have a personal Rocky 9 and/or AlmaLinux 9 server somewhere in the cloud ... I can test it there I suppose.   Could be easier than installing one of them at work   😄

iliaross:
[...]

PR #2506, opened 2025-06-27
Fix latest SSLeay support for redirects to SSL work

iliaross:
Hey Jamie!

This PR addresses the issue with newer versions of Net::SSLeay (1.93+) where accessing in non-SSL mode causes it to hang indefinitely, as described in [this issue](https://github.com/webmin/webmin/issues/2504).

The minimal fix is to peek at the socket to get the connection type, set a flag, and avoid calling `Net::SSLeay::read($ssl_con)` when not in SSL mode at all.

I’ve tested it, and it seems to work jsut fine. A different fix is possible, but it would be much more complex.

### DIFF
--- a/miniserv.pl
+++ b/miniserv.pl
@@ -937,7 +937,7 @@ while(1) {
 						$ssl_con || exit;
 						}
 					else {
-						$plain_http = 1;
+						$use_ssl = 0;
 						}
 					}
 
@@ -3257,15 +3257,7 @@ while(($idx = index($main::read_buffer, "\n")) < 0) {
 	# need to read more..
 	&wait_for_data_error() if (!$nowait);
 	if ($use_ssl) {
-		if (!$plain_http) {
-			$more = Net::SSLeay::read($ssl_con);
-			}
-		else {
-			# Expected output from Net::SSLeay::read is empty string
-			# in non-SSL mode before version 1.93, which now just
-			# hangs indefinitely
-			$more = '';
-			}
+		$more = Net::SSLeay::read($ssl_con);
 		}
 	else {
 		my $bufsize = $config{'bufsize'} || 32768;
--- a/miniserv.pl
+++ b/miniserv.pl
@@ -918,11 +918,27 @@ while(1) {
 
 				# Initialize SSL for this connection
 				if ($use_ssl) {
-					($ssl_con, $ssl_certfile,
-					 $ssl_keyfile) = &ssl_connection_for_ip(
-							   SOCK, $ipv6fhs{$s});
-					print DEBUG "ssl_con returned $ssl_con\n";
-					$ssl_con || exit;
+					my $byte = '';
+					# Look at the first byte of the socket
+					# buffer but don't consume it
+					recv(SOCK, $byte, 1, MSG_PEEK);
+					if (length($byte) &&
+					    # Check if the first byte is a TLS
+					    (ord($byte) == 0x16 ||
+					    # Check if the first byte is SSL
+					    (ord($byte) & 0x80))) {
+						($ssl_con,
+						 $ssl_certfile,
+						 $ssl_keyfile) =
+							&ssl_connection_for_ip(
+							    SOCK, $ipv6fhs{$s});
+						print DEBUG "ssl_con returned ".
+							"$ssl_con\n";
+						$ssl_con || exit;
+						}
+					else {
+						$plain_http = 1;
+						}
 					}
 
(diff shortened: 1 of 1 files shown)


miniserv.pl
```diff
@@ -918,11 +918,27 @@ package miniserv;
...
+						 $ssl_certfile,
+						 $ssl_keyfile) =
+							&ssl_connection_for_ip(
+							    SOCK, $ipv6fhs{$s});
+						print DEBUG "ssl_con returned ".
+							"$ssl_con\n";
+						$ssl_con || exit;
+						}
+					else {
+						$plain_http = 1;
```
jcameron, line 940: Why not just set $use_ssl = 0 here, instead of creating a new variable?

jcameron: So I feel like this is adding complexity to support the old mode behavior, and I don't like the use of a global variable that is referenced later. In fact it won't necessarily even work, because miniserv supports multiple ports that might receive connections at the same time.

Can we not do this check for the wrong mode around line 1357, where it attempts to happen currently?

iliaross: And, it’s probably the best option since users will have the flexibility to connect either way. We just need to show a banner (or a special message) on the login page when someone connects using a non-HTTPS connection...

I think this is actually a bonus and also solves the original problem...

jcameron: Yesh actually, if we can reliably switch to HTTP mode for a non-SSL connection and then show a warning somewhere, that would be ideal. However, I wonder if that first-byte detection will work for all browsers and other CLI tools that might connect to Webmin?

jcameron: I think you're correct, but we do have to be careful here because if for some reason the protocol changes and TLS version 7 starts with a 0x17 byte, we would break all SSL connections!

iliaross: > if for some reason the protocol changes and TLS version 7 starts with a 0x17 byte, we would break all SSL connections!

That byte is unlikely to ever change and is the permanently assigned "ContentType" (22) value for "Handshake" records in TLS, which is controlled by [IANA](https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-5).

iliaross:
Anyway, I suggest you merge it and make any further changes if needed. Also, we probably don’t need that code block that redirects from non-SSL to SSL anymore.

I’m off for today… If you merge it, I’ll add a notice for all non-SSL connections on the login page tomorrow.

miniserv.pl
jcameron: Not if we add a comment explaining why

iliaross: I suggest we don’t take any extra precautions right now, since they’re not based on anything. We’ll have years—plenty of time to add a simple patch long before any new TLS record type goes live in production.

iliaross:
I have added a badge and an explanation about the connection not being secure:


https://github.com/user-attachments/assets/11eec61b-fa94-4d9c-b5df-ac7d6a642576

jcameron:
That's nice! Does this work in the other theme though? Also, shouldn't it have a link to the correct URL?

iliaross:
> Does this work in the other theme though?

Yes, of course! I made it work in the other theme too, e.g.:

image

> Also, shouldn't it have a link to the correct URL?

Probably not, as we clearly warn the user and explain what is going on. Moreover, a browser won't allow a user to use HTTP once HTTPS with a proper SSL certificate has been used when HSTS is enabled...

...we should actually [always have](https://github.com/webmin/webmin/commit/fb832eff8262571ed94823aa73025049ac633504) HSTS enabled by default.

jcameron:
> Probably not, as we clearly warn the user and explain what is going on. Moreover, a browser won't allow a user to use HTTP once HTTPS with a proper SSL certificate has been used when HSTS is enabled...

Ok but how does the user know what to do when they see this "not secure" warning?

iliaross:
Gosh, I don't know! Do you think we should actually add to the tooltip something like “Click this badge to switch to a secure connection” and make the badge itself clickable?

jcameron:
Can we add a link in the tooltip? That would do it ..

iliaross:
> Can we add a link in the tooltip? That would do it ..

Not without using JavaScript tooltips—native tooltips don’t support links...

jcameron:
In that case, can the badge itself be clickable?

iliaross:
Yes, sure, the badge can be clickable—it’s done now for both themes, e.g.:

image

jcameron:
Nice thanks!

PR #2510, opened 2025-07-02
Add ability to track session state using session database

iliaross:
Hey Jamie!

As we [discussed here](https://forum.virtualmin.com/t/auto-logout-after-x-not-working/133973/28?u=ilia), we need a clear and reliable way for the Miniserv to know if the client is active or not so the inactivity timer works properly.

Closing the socket isn’t the right approach, even though it’s simpler. I spent the morning reviewing Miniserv’s session code and found a good, simple solution.

At first, I thought about storing session activity in a separate file, but that turned out to be suboptimal (first commit). Instead, I implemented it properly (second commit) to store in the session database.

The theme will, based on its logic (like when the tab isn’t focused or is switched), call code after a certain delay to tell Miniserv that the user session is inactive. It will then stop updating the heartbeat for that socket.

When the user comes back, we’ll tell Miniserv again that the session is active.

The patch is simple with minimal changes. Please take a look!

### DIFF
--- a/miniserv.pl
+++ b/miniserv.pl
@@ -6,7 +6,6 @@ package miniserv;
 use Socket;
 use POSIX;
 use Time::Local;
-use Fcntl qw(:DEFAULT :flock);
 eval "use Time::HiRes;";
 
 @itoa64 = split(//, "./0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz");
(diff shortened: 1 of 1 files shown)


jcameron:
I don't think this is the right approach, and in fact we should first properly define what "idle" means. To me, the idle timer should never be reset unless the **user** is actually doing something. Just having the dashboard open, even if focused, does not indicate user activity.

What if for starters the dashboard stopped being refreshed unless it's in a visible tab?

iliaross:
> What if for starters the dashboard stopped being refreshed unless it's in a visible tab?

Well, the main issue right now isn’t with that—the dashboard is already not being refreshed when tab isn't visible. It is handled by using the `pause` param sent from the client to `stats.pl`.

The real problem is with Miniserv. It keeps the heartbeat for any opened socket connection. For example, in the Terminal module, this is probably a good thing since it would be annoying to get disconnected while idle for some time, and importantly user can control it by simply leaving the Terminal module.

But with the dashboard stats, it runs in the background all the time, and the user can’t control it. Yet, we do signal the socket, as I already mentioned, to stop sending data when browser's tab isn't focused or when user is away from the computer, but we don’t close the socket connection.

With the change in this PR, we’ll be able to tell Miniserv that the session is idle in a specific case scenario. It can be used and decided selectively by the theme.

jcameron:
So what we could do then is simply not update the last active session time here https://github.com/webmin/webmin/blob/master/miniserv.pl#L5934 if there was no activity on the socket. The connection can then be left open, and if timeouts are enabled it will get closed if the tab was inactive. 

I can create a simple change to implement that.

iliaross:
> So what we could do then is simply not update the last active session time here https://github.com/webmin/webmin/blob/master/miniserv.pl#L5934 if there was no activity on the socket. The connection can then be left open, and if timeouts are enabled it will get closed if the tab was inactive.

But this is exactly what I did in my PR? 😳

jcameron:
Sort-of - we can do this without needing to introduce any extra way for the theme to signal that the connection is idle.

iliaross:
> Sort-of - we can do this without needing to introduce any extra way for the theme to signal that the connection is idle.

Could you give me a hint on how to do it?

jcameron:
I'll send a commit later today

iliaross:
Alright, thanks! Though, will the theme need to do anything?

jcameron:
Ideally no

iliaross:
> Ideally no

It won’t work reliably. If you’re thinking of reading the previous state from the HTTP request, that’s not going to be fail-proof.

What if a user uses the terminal for 100 days straight without reloading the page? We can’t assume users won’t do that.

There’s no way to make it reliable without communicating with the client, I don't think?

iliaross:
Jamie, could you take another look at this? I still don’t think we can make it work reliably without interacting with the client browser.

jcameron:
Sure, I will take a look at this soon.

iliaross:
Thanks!

jcameron:
Ok check out the linked patch above

iliaross:
Hohoho, wow! This is nearly perfect! The only downside compared to my implementation in this PR is that it will still close the connection, for instance, when using the Terminal, if the `tail -f` command is invoked on a file that produces no output within the timeout period.

But this is an edge case, and considering it's much simpler and works with all themes automatically, we should definitely prefer your solution. I'm closing my PR, thanks!

jcameron:
Excellent, thanks!

PR #2512, opened 2025-07-05
Add improvements to systemd section of Init module

iliaross:
Hey Jamie!

As [discussed here](https://github.com/webmin/webmin/commit/72621c292910449c9eeaeec85462db08334d9a0d#commitcomment-161486940), I am adding a separate column for unit types for _systemd_ systems.

Additionally, as part of other _systemd_ improvements, I removed the "Description" column and incorporated the description into a native HTML accordion.

image

### DIFF
--- a/gray-theme/unauthenticated/gray-theme.css
+++ b/gray-theme/unauthenticated/gray-theme.css
@@ -422,7 +422,7 @@ details.ui_hidden_table_start > summary::-webkit-details-marker {
     color: #a1acc0;
 }
 
-details.on-hover:not(:hover) summary::after {
+details:not([open]).on-hover:not(:hover) summary::after {
   visibility: hidden !important;
 }
 
--- a/gray-theme/unauthenticated/gray-theme.css
+++ b/gray-theme/unauthenticated/gray-theme.css
@@ -422,6 +422,10 @@ details.ui_hidden_table_start > summary::-webkit-details-marker {
     color: #a1acc0;
 }
 
+details.on-hover:not(:hover) summary::after {
+  visibility: hidden !important;
+}
+
 details.inline {
   cursor: pointer;
   padding: 0;
--- a/init/index.cgi
+++ b/init/index.cgi
@@ -359,7 +359,7 @@ elsif ($init_mode eq "systemd" && $access{'bootup'}) {
 		my $title = ($u->{'boot'} == -1 ?
 			    &html_escape($sname) :
 			    &ui_link($l, &html_escape($sname)));
-		my $desc = &html_escape($u->{'desc'});
+		my $desc = $config{'desc'} ? &html_escape($u->{'desc'}) : '';
 		print &ui_columns_row([
 			&ui_checkbox("d", $u->{'name'}, undef),
 			!$desc ? $title : &ui_details({
--- a/init/index.cgi
+++ b/init/index.cgi
@@ -335,11 +335,13 @@ elsif ($init_mode eq "systemd" && $access{'bootup'}) {
 		   &select_invert_link("d"),
 		   &ui_link("edit_systemd.cgi?new=1", $text{'index_sadd'}) );
 	print &ui_links_row(\@links);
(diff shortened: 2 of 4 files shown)


jcameron:
Why hide the description column though? That seems like pretty useful information!

iliaross:
> Why hide the description column though? That seems like pretty useful information!

Because it’s so long, it makes the table scrollable and actually not very useful for everyday use...

iliaross:
However, the new HTML accordion makes it searchable via the normal Ctrl+F or ⌘+F, even when it is collapsed.

jcameron:
Seems like there is plenty of room for the column to me, even on a 13" macbook. And we're only adding one small extra column. ALso, having it collapsible like this isn't the way other Webmin modules work.

Perhaps instead make some of the other headers smaller, like rename "Start at boot time?" to "On boot", and "Running now" to "Running".

iliaross:
> Seems like there is plenty of room for the column to me, even on a 13" macbook.

This is how it looks on my MacBook 13' with the old layout (without new column, and it's kinda crowded):

image

> Also, having it collapsible like this isn't the way other Webmin modules work.

That is probably okay, or at least we could fix it up later, in particular for descriptions for other init systems in this module.

> Perhaps instead make some of the other headers smaller, like rename "Start at boot time?" to "On boot", and "Running now" to "Running".

We could do that, but I don't think it will help much.

I think description isn't really essential in everyday use, right? Most times a user knows the unit they are looking for to check or edit?

jcameron:
Is that a full-sized window? Here's what it looks like for me : 

image

A lot of the time the unit names are not descriptive, like for example the unit that runs Apache is just called httpd.

iliaross:
> Is that a full-sized window? Here's what it looks like for me :

Ah, yes... Even though it was the full-sized window for me, its resolution was set one step below Default in macOS Settings ⇾ Displays.

> A lot of the time the unit names are not descriptive, like for example the unit that runs Apache is just called httpd.

True, however—what's then. Let's say I want to find Apache unit, right? So what do I do? I hit ⌘+F, start typing "Apache" and hit enter, and get scrolled down to the unit itself with accordion being opened automatically. Is this bad in any way? It feels like we hit two targets with one shot. Having it not take space and being easily searchable...

iliaross:
I’d suggest giving it an actual try to see how it feels for you. I think it’s a better experience...

jcameron:
No I really don't like it, please put the column back. We shouldn't be introducing new UI styles like hidden collapsible descriptions.

iliaross:
Alright, will do...

jcameron:
Thanks! To be clear, I do like moving the type to a separate column, but I still want to keep the description column.

iliaross:
> To be clear, I do like moving the type to a separate column, but I still want to keep the description column.

Alright, I have restored the description column.

Though, maybe we should make it disabled by default in the module config for _systemd_ systems?

jcameron:
No, but we could add an option to hide it, but show it by default.

iliaross:
We already have such option. We would want to tweak it for _systemd_ though, not to show run-level…

jcameron:
We never show runlevel for systemd units though, because there is no such concept?

iliaross:
> We never show runlevel for systemd units though, because there is no such concept?

I meant on the config. Fixed here https://github.com/webmin/webmin/pull/2512/commits/76efb4ee104a176ce6b7c19d1e3206b1e2e3417d.

jcameron:
Oh I see, that makes sense, thanks!

gray-theme/unauthenticated/gray-theme.css
```diff
@@ -422,6 +422,10 @@ details.ui_hidden_table_start > summary::-webkit-details-marker {
     color: #a1acc0;
 }
 
+details:not([open]).on-hover:not(:hover) summary::after {
```
jcameron, line 425: Do we still need this if the description column is either always shown or always hidden?

PR #2514, opened 2025-07-06
Add the ability to re-enforce SSL connections and enable it by default

iliaross:
Hello Jamie,

As we [discussed earlier](https://github.com/webmin/webmin/commit/489583708e32967e7d836659395fd8f836a4e76c#commitcomment-161540167), a user suggested we change the default behavior to not redirect to SSL.

While having an option to connect to the non-SSL version of Webmin can be helpful in certain situations, I believe it’s more of an exception rather than a common use case.

This PR addresses this issue by implementing a more user-friendly redirect to HTTPS connections, and also allows this to be configurable from Webmin and Usermin configuration modules.


jcameron:
Thanks! Was there no config option to control this redirect previously though? Or did we just have it hard-coded to be always enabled?

iliaross:
> Thanks! Was there no config option to control this redirect previously though? Or did we just have it hard-coded to be always enabled?

We had a similar redirect before using JavaScript, which I didn’t really like, but back then we didn’t support both HTTP and HTTPS modes at the same time. Now we do, so I think the check should also be handled differently, as in this PR...

iliaross:
> Was there no config option to control this redirect previously though?

Ah, and about the config option—no, we never had an explicit option for this as far as I know. There’s still that odd `ssl_redirect` option in Usermin, but it doesn’t seem to do anything.

I was actually going to ask you to remove it...

jcameron:
Ok this looks good then!

iliaross:
> Ok this looks good then!

Thanks!

> I was actually going to ask you to remove it

Will you?

jcameron:
> I was actually going to ask you to remove it

Wait remove what exactly?

iliaross:
> Wait remove what exactly?

Search through `ssl_redirect` in Webmin project...

jcameron:
Oh I see it now - deleted here : https://github.com/webmin/webmin/commit/d2863b7c4c736985d138d6b8338567a250eafb9f

iliaross:
Thanks! I wonder now, what in the world is this https://github.com/webmin/webmin/commit/26138b82f5d75ddb75752eb20cc3a1fac198b171#diff-8611ded294eb739d56a6ed2f75ac74114f591707c300cd967dfac4c5a52ccf9cR82-R95. I will have to take a closer look at it.

PR #2532, opened 2025-08-20
Add support for Dovecot 2.4 in API

iliaross:
Hey Jamie,

As we discussed in private chat and to address [this issue](https://github.com/virtualmin/virtualmin-gpl/issues/1094).

I'll send you a small linked-patch for Virtualmin to work in a bit and link it here.

After you merge it, I will fix the Dovecot module UI to actually work with this new API. There's not much to do, but some fixes will be needed. I wanted to submit API changes only first.

### DIFF
--- a/dovecot/dovecot-lib.pl
+++ b/dovecot/dovecot-lib.pl
@@ -11,6 +11,11 @@ use WebminCore;
 @mail_envs = ( undef, "maildir:~/Maildir", "mbox:~/mail/:INBOX=/var/mail/%u",
 	       "maildir:~/Maildir:mbox:~/mail/" );
 
+# Dovecot version specific mapping if any
+our $base_version = &get_dovecot_based_version();
+require "$module_root_directory/dovecot$base_version-lib.pl"
+	if (-r "$module_root_directory/dovecot$base_version-lib.pl");
+
 # get_config_file()
 # Returns the full path to the first valid config file
 sub get_config_file
@@ -158,6 +163,7 @@ return @rv;
 sub find
 {
 local ($name, $conf, $mode, $sname, $svalue, $first) = @_;
+$name = &map_find($name) if (defined &map_find);
 local @rv = grep { !$_->{'section'} &&
 		   $_->{'name'} eq $name &&
 		   ($mode == 0 && $_->{'enabled'} ||
@@ -166,6 +172,12 @@ if (defined($sname)) {
 	# If a section was requested, limit to it
 	@rv = grep { $_->{'sectionname'} eq $sname &&
 		     $_->{'sectionvalue'} eq $svalue } @rv;
+	}
+if (@rv && defined &map_value) {
+	foreach my $rv (@rv) {
+		my (undef, $v) = &map_value($_[0], $rv->{'value'});
+		$rv->{'value'} = $v;
+		}
 	}
 if (wantarray) {
 	return @rv;
@@ -219,6 +231,16 @@ return wantarray ? @rv : $rv[0];
 sub save_directive
 {
 local ($conf, $name, $value, $sname, $svalue) = @_;
+if (defined &map_value) {
+	if (ref($name)) {
+		my ($nn, $vv) = &map_value($name->{'name'}, $value);
+		$name->{'name'} = $nn;
+		$value = $vv;
+		}
+	else {
+		($name, $value) = &map_value($name, $value);
+		}
+	}
 $newconf = [ grep { $_->{'file'} !~ /^\/usr\/share\/dovecot/ &&
                     $_->{'file'} !~ /^\/opt/ } @$conf ];
 if (@$newconf) {
@@ -319,6 +341,9 @@ elsif (!$dir && defined($value)) {
 sub save_section
 {
 local ($conf, $section) = @_;
+if (defined &map_members) {
+	$section->{'members'} = &map_members($section->{'members'});
+	}
 local $lref = &read_file_lines($section->{'file'});
 local $indent = "  " x $section->{'indent'};
 local @newlines;
@@ -347,6 +372,9 @@ foreach my $m (@{$section->{'members'}}) {
 sub create_section
 {
 local ($conf, $section, $parent, $before) = @_;
+if (defined &map_members) {
+	$section->{'members'} = &map_members($section->{'members'});
+	}
 local $indent = "  " x $section->{'indent'};
 local @newlines;
 push(@newlines, $indent.$section->{'name'}." ".$section->{'value'}." {");
@@ -543,6 +571,15 @@ sub get_dovecot_version
 {
 local $out = &backquote_command("$config{'dovecot'} --version 2>&1");
 return $out =~ /([0-9\.]+)/ ? $1 : undef;
+}
+
+# get_dovecot_based_version()
+# Returns the dovecot version number without the patch level and without dots
+sub get_dovecot_based_version
+{
+my $base_version = &get_dovecot_version();
+$base_version = $base_version =~ /^(\d)\.(\d)/ ? $1.$2 : '';
+return $base_version;
 }
 
 # version_atleast(ver)
--- a/dovecot/dovecot-lib.pl
+++ b/dovecot/dovecot-lib.pl
@@ -614,7 +614,8 @@ else {
 # Version aware read
 sub find_mapped
 {
-return &dovecot('find', @_) if defined &dovecot && !$dovecot{main};
+return &dovecot_shim('find', @_)
+	if defined &dovecot_shim && !$dovecot_shims{main};
 return &find(@_);
 }
 
@@ -624,8 +625,8 @@ sub find_value_mapped
 {
 my ($name, $conf, $mode, $sname, $svalue) = @_;
 my @rv;
-if (defined &dovecot && !$dovecot{main}) {
-	@rv = &dovecot('find', $name, $conf, $mode, $sname, $svalue, undef);
+if (defined &dovecot_shim && !$dovecot_shims{main}) {
+	@rv = &dovecot_shim('find', $name, $conf, $mode, $sname, $svalue, undef);
 	}
 else {
 	@rv = &find($name, $conf, $mode, $sname, $svalue, undef);
@@ -648,7 +649,8 @@ else {
 # Version aware write
 sub save_directive_mapped
 {
-return &dovecot('save_directive', @_) if defined &dovecot && !$dovecot{main};
+return &dovecot_shim('save_directive', @_)
+	if defined &dovecot_shim && !$dovecot_shims{main};
 return &save_directive(@_);
 }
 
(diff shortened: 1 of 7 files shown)


iliaross:
Virtualmin related PR is here https://github.com/virtualmin/virtualmin-gpl/pull/1097.

dovecot/dovecot-lib.pl
```diff
@@ -11,6 +11,11 @@
 @mail_envs = ( undef, "maildir:~/Maildir", "mbox:~/mail/:INBOX=/var/mail/%u",
 	       "maildir:~/Maildir:mbox:~/mail/" );
 
+# Dovecot version specific mapping if any
+our $base_version = &get_dovecot_based_version();
+require "$module_root_directory/dovecot$base_version-lib.pl"
+	if (-r "$module_root_directory/dovecot$base_version-lib.pl");
```
jcameron, line 17: Instead of creating a complex method of incuding a file based on the Dovecot version and the overhead of querying the version every time dovecot-lib.pl is loaded, why not just make the `map_find` function check the version and return a different hash?

dovecot/dovecot24-lib.pl
```diff
@@ -0,0 +1,68 @@
...
+	ssl_key                => 'ssl_server_key_file',
+	ssl_key_file           => 'ssl_server_key_file',
+	ssl_key_password       => 'ssl_server_key_password',
+	);
+return wantarray ? %map : ($map{$name} || $name);
+}
+
+# map_find(name)
+# Translate key
+sub map_find
```
jcameron, line 30: All these function names are way too short ... they should be more descriptive.

dovecot/dovecot24-lib.pl
```diff
@@ -0,0 +1,68 @@
...
+
+# Translate values
+sub map_value
+{
+my ($name, $value) = @_;
+$name = &map_find($name);
+
+if ($name =~ /^(ssl_server_cert_file|ssl_server_key_file|ssl_server_ca_file|ssl_client_ca_file)$/) {
+	# Drop obsolete "<"
+	$value =~ s/^\s*<\s*// if defined $value;
```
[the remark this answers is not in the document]
iliaross: Isn't this function for mapping value names specifically? It isn't for parameter names.

dovecot/dovecot24-lib.pl
iliaross: I just wanted the function names to be simple and correspond with the original API.

jcameron:
Actually, the more I think about ti, the more I realize that this is the wrong approach. We shouldn't be handling different Dovecot versions in the low-level API for editing the config at all. No other module works this way, and for good reason.

The only correct solution is to do the version checks at the caller level, like : https://github.com/webmin/webmin/blob/master/dovecot/edit_ssl.cgi#L23

iliaross:
Oh, yeah! I know about this approach, which is really hard to support, especially with the changes in version 2.4, and in particular in Virtualmin—it's going to make the already pretty complex callers a total nightmare.

If you're really sure you want it, go ahead, but I'd suggest taking a closer look at what it will take to fix the Virtualmin caller alone to work with the changes we have in Dovecot 2.4. Also, we will need to do something similar in the Virtualmin Install script.

In fact, there's **no good reason** for a caller to be aware about the Dovecot version and make calls based on it. The API should just work despite of software version used on the other end; otherwise, we end up duplicating a lot of code in different modules, which isn't good in my opinion...

jcameron:
I think what we could do instead is offer an API in the Dovecot module to convert directive names, and then have callers use that when calling `save_directive` . I get that this would mean that changes to Virtualmin are needed as well, but it's cleaner than changing this low-level API.

For example, imagine if the Linux `grep` command automatically translated output from the file you ran it on based on the user's language? Sure that would be "useful" in some sense, but clearly it would be an API operating at the wrong level.

iliaross:
> I think what we could do instead is offer an API in the Dovecot module to convert directive names, and then have callers use that when calling save_directive . 
> Sure that would be "useful" in some sense, but clearly it would be an API operating at the wrong level.

Yes, it definitely should stay in the Dovecot module. We can call this new API something like `find_mapped` (or  `find_m`), `create_section_mapped` and  etc.

Unlike `grep`, Dovecot devs keep making odd, breaking, and hard-to-support changes. What we should do in this particular case is create an abstraction layer for our Dovecot API. So, instead of calling `save_directive('exact-directive-name')`, we would call `save_directive_mapped('known-directive-name')`, where the caller wouldn't have to worry about changes made upstream. We would only need to change the Webmin module if and when needed.

So, are you okay with that? Keep `find` and other functions intact and create a new `*_mapped` or `*_m` variants with mapping, then call them on the Virtualmin side like that?

iliaross:
Or to keep it simpler, we can avoid changing the existing API entirely and just conditionally load `dovecot24-lib.pl` like we do now, but use function names that already exist in the main library (like `find`, `create_section` and etc). The `require` will automatically take over the current functions, overriding some of them (which we need) when `dovecot24-lib.pl` is loaded, because in Perl the last definition wins with in the same package. This solution is arguably acceptable too, however, even though it is still the simplest and might be a tiny bit cleaner, it will still require **duplicating code** in those functions with some tweaks we currently have built in in this PR.

To be clear, I'm confident, it's not really possible for the caller to **do it nicely** because it will make things more complicated and harder to support, as it sometimes requires logic before and after the current API calls. It will be a total mess. And, overall, it's just the wrong approach.

I don't know why you find the current implementation not clean enough—it doesn't duplicate code, it doesn't change the current behavior, and it just surgically does what the particular version of Dovecot needs.

Currently, the way we use the Dovecot API in Virtualmin is mostly great. We _get_ the config, _find_ the value, and _save_ it—perfect. All the background, version specific processes happening **cannot be done** on the plugin side—it's just the wrong place for it. The plugin calling an API shouldn't worry about the version or anything related to the other module.

The CGIs are the only exception in Webmin Dovecot's module, meaning user-facing UI CGIs, where we check if the given version of Dovecot supports specific options, and then use the global API with chosen, specific logic. That's fine because it's **all inside one module**. But a plugin calling the Webmin module **should never** perform any module's version-specific logic. Yes, it's easier to implement as a hot-fix, but it's not right in the long term.

The bottom line, and my suggestion is to do whatever you propose or agree on, but **without adding** anything to the currently well-implemented callers in Virtualmin. There are currently some quirky checks in Virtualmin that compare Dovecot 2 to 1, which can be safely removed because Dovecot 1 hasn't been around since CentOS 5. By doing this, we will completely eliminate incorrect version-specific checks in the caller and keep the code logic **clean and contained** within Dovecot Webmin module. The call is yours.

jcameron:
So I like the `save_directive_mapped` solution - that way it's clearer to the caller that they are getting potentially a different field. BUT this will require changes to callers, and that is OK in my opinion. Sometimes it's worth doing more work to write clean code, and in this case it is 100% no questions asked cleaner to have the Dovecot module API just behave simply and not do any "magic". 

For a counterexample, consider how PHP in the past tried to solve the problem of SQL injection at the wrong level with it's magic quotes feature. Yes this was easier for callers, but it was unquestionably wrong from a programming practice point of view.

Also, regarding Dovecot 1, we can just drop support for that now and so remove any special case checks for older directives.

iliaross:
> So I like the `save_directive_mapped` solution - that way it's clearer to the caller that they are getting potentially a different field.
> Also, regarding Dovecot 1, we can just drop support for that now and so remove any special case checks for older directives.

Alright, I'll drop that and add more patches to the existing PRs tomorrow. Thanks!

jcameron:
FYI, I dropped support for Dovecot 1 here : https://github.com/virtualmin/virtualmin-gpl/commit/1b1120cc19b529066a906c9bb7d370304849d835

iliaross:
> What we should do in this particular case is create an abstraction layer for our Dovecot API. So, instead of calling `save_directive('exact-directive-name')`, we would call `save_directive_mapped('known-directive-name')`
>> So I like the save_directive_mapped solution - that way it's clearer to the caller that they are getting potentially a different field.

As I re-read it again, it's not exactly what I meant. I didn't mean to change the caller—I was clear about how I feel about that. What I talked about earlier assumed that `save_directives` receives an abstracted value like ('ssl-cert-key-whatevert'), and the module does the mapping. All current Dovecot API keys used before version 2.4 technically become abstracted values (since we're not changing the caller's keys and create another map for earlier versions) that still work for the older versions but get transformed into something else for the new.

I get that you want to replace the caller with a new function call. The benefit of doing this and calling `save_directives_mapped` clearly shows the reader that something different is happening here and to keep it consistent with other Webmin modules' APIs. **However**, you can't deny that it has some significant trade-offs too. It's like taking one step forward and two steps back with **this specific problem** we're solving. And, it's not just about adding a new function call and changing the caller; it also involves creating a Webmin dependency check, which adds complexity and makes the code actually harder to read. And that **isn't the only** change that will be needed anyway on hundreds of callers! It also **will require** a _lot more_ changes in Dovecot CGIs too.

---

Please let me take another shot at how we could do it without changing the caller, yet **keeping original Dovecot API conventional**. This change will be with slightly more code, but make it **way cleaner and way more readable** than I originally suggested in this PR. See this final change https://github.com/webmin/webmin/pull/2532/commits/863e178c7bbdf0a14ee5e49857d8ff4846953ced — to be clear, it works with Virtualmin and saves almost all options in the Dovecot module, like the "SSL Configuration" page—yet, I haven't tested it on all Dovecot CGI pages, but that's not important for this PR—I will easily fix it later once we agree on the basic principle.

If you still find it unacceptable, despite the negative consequences I mentioned in this ticket, I will close these PRs and let you solve it. However, I want to emphasize that it's important to understand which problem we're solving here. I feel you may not have looked deeply into it, but once you do, there won't be many options left, and solving it cleanly the way you suggested from the start won't be really possible considering all the trade-offs. For the record, this PR is only +172 0 with all the final working changes. And it's only +19 -4 for Virtualmin's [PR](https://github.com/virtualmin/virtualmin-gpl/pull/1097/files).

# 😊

jcameron:
We basically cannot do this without the callers knowing about the new directives, or without changing the meaning of the API. And I really don't think changing the API to be version-aware is a good idea, because it goes against how all other such APIs in Webmin work.

A clean solution would be to add new functions in the Dovecot module to get and set all the SSL paths, which could then be aware of the version and use the correct directives and remove any special characters. Yes this would involve a Virtualmin fix to use it, but I'd much prefer cleaner code that takes longer than adding "magic" to an existing API.

iliaross:
> We basically cannot do this without the callers knowing about the new directives, or without changing the meaning of the API.

Well, alrighty then... let me try doing it the way you suggest.

iliaross:
Done! Check this patch out https://github.com/webmin/webmin/pull/2532/commits/0573074cacbef47ecd0c57ea05f3d95f6d55a49f. If that's the way we go, I will update Virtualmin PR tomorrow accordingly.

iliaross:
> I will update Virtualmin PR tomorrow accordingly.

And, here is the patch on VIrtualmin side https://github.com/virtualmin/virtualmin-gpl/pull/1097/commits/3dfefb4cd604dae4dddf82babfdf9600a0408dca.

PR #2553, opened 2025-09-23
Add final fixes to address ongoing issue with checking remote host

iliaross:
Hey Jamie,

According to what we discussed in the email earlier today.

### DIFF
--- a/web-lib-funcs.pl
+++ b/web-lib-funcs.pl
@@ -13261,10 +13261,12 @@ if (!$def && $gconfig{'webmin_email_url'}) {
 	# From a config option
 	$url = $gconfig{'webmin_email_url'};
 	}
-elsif ($ENV{'HTTP_HOST'}) {
+elsif ($ENV{'HTTP_HOST'} || $ENV{'SSL_HOST'}) {
 	# From this HTTP request
-	my $host = $ENV{'HTTP_HOST'};
 	my $port = $ENV{'SERVER_PORT'} || 80;
+	my $host = $ENV{'SSL_HOST'}
+		? "$ENV{'SSL_HOST'}:$port"
+		:  $ENV{'HTTP_HOST'};
 	if ($host =~ s/:(\d+)$//) {
 		$port = $1;        
 		}


jcameron:
So there are several issues here :

1 - SSL_HOST does not need to have a port in it, because it's the hostname used to select the SSL cert (the TLS servername field), which does not have the same meaning as HTTP_HOST.

2 - I made a mistake by mixing up the checks for SSL_HOST and HTTP_HOST, because a clever attacker could fake one but not the other. The fix for this is here : https://github.com/webmin/webmin/commit/890a4ffd3f5480865629b5469876b63507f2cb64

3 - We can allow self-signed certs, but we *must* disallow totally wildcard certs for `*` because the hostnames in the cert are what we are using to determine if the HTTP_HOST is valid.  Hence the need to check SSL_HOST_CERT cannot be removed.

4 - We cannot show a warning when not in SSL mode, we have to entirely block the request. Otherwise the attacker could just get around all these protections by degrading the connection to non-SSL mode!

You raise a good point about repeated reads of the same cert file though! I fixed that here : https://github.com/webmin/webmin/commit/d5a22a592fed2173661c389d80c943794da9f6e2

As for using pure perl to read the cert, I wish that was possible but I don't see a way. This is why everywhere else in webmin and virtualmin we shell out to openssl.

iliaross:
> As for using pure perl to read the cert, I wish that was possible but I don't see a way. This is why everywhere else in webmin and virtualmin we shell out to openssl.

Actually, I'll email you the working solution for this right now. I got it working today. It took quite a bit of time and effort—the good thing is it's literally 22-25 times faster than the original one, and it produces identical output and also works on all systems.

But unfortunately, it's really-really complicated! So complicated that it's probably unacceptable.

> You raise a good point about repeated reads of the same cert file though! 

Thanks, I'll test it out tomorrow. Unfortunately, it's too late today, and I don't have any time left.

> We cannot show a warning when not in SSL mode, we have to entirely block the request. Otherwise the attacker could just get around all these protections by degrading the connection to non-SSL mode!

Non-SSL mode is optional and turned off by default. We clearly warn people. I'm sure some intranets don't use SSL, and it might not be a problem for them—it has to work for them.

And, if using the same logic, we should remove non-SSL support in Webmin entirely, because sending a root password and/or cookie over HTTP is **much more insecure** than faking a remote host and hoping the targeted user will blindly click a link or a scanner does—risks are much higher. So, for consistency, I suggest keeping it optional.

> We can allow self-signed certs, but we must disallow totally wildcard certs for * because the hostnames in the cert are what we are using to determine if the HTTP_HOST is valid. Hence the need to check SSL_HOST_CERT cannot be removed.

I don't quite understand this. Can you unpack it more? What's the point of checking `SSL_HOST_CERT`?

And, yes—I'm all for self-signed certificates support in this case scenario because imposing unnecessary limitations is generally a bad practice.

I'll check it tomorrow, but please provide more details so I can make better patches (if still needed) to finish it up.

> I made a mistake by mixing up the checks for SSL_HOST and HTTP_HOST, because a clever attacker could fake one but not the other. The fix for this is here : https://github.com/webmin/webmin/commit/890a4ffd3f5480865629b5469876b63507f2cb64

I'll take a look, thanks!

> SSL_HOST does not need to have a port in it, because it's the hostname used to select the SSL cert (the TLS servername field), which does not have the same meaning as HTTP_HOST.

Ah, okay! This actually makes sense now because it seems that we can't control the port number, which can also be easily spoofed.

jcameron:
> Non-SSL mode is optional and turned off by default. We clearly warn people. I'm sure some intranets don't use SSL, and it might not be a problem for them—it has to work for them.

But we still need the SSL cert as a source of valid hostnames. Could we still do that even when not in SSL mode? Or maybe we can add an extra config option to allow the admin to manually enter more hostnames? 

> I don't quite understand this. Can you unpack it more? What's the point of checking SSL_HOST_CERT?
> And, yes—I'm all for self-signed certificates support in this case scenario because imposing unnecessary limitations is generally a bad practice.

For sure, nothing in the current code prevents use of *self-signed* certs as long as they have a proper hostname. 

What `SSL_HOST_CERT` indicates is that the hostname in the HTTP request matches one of the hostnames in the certs. This is how we prevent an attacker sending a request with a fake hostname like `evil.org` , which is then included in the email.

iliaross:
> But we still need the SSL cert as a source of valid hostnames. Could we still do that even when not in SSL mode? Or maybe we can add an extra config option to allow the admin to manually enter more hostnames?

Yeah, there'd no way to validate a hostname without using SSL. But no one in their right mind should use Webmin without SSL on an open network in the first place. The danger goes both ways.

And, about an extra config option—I think we already have “webmin_email_url” option just for that.

iliaross:
Alright, Jamie—please review all final changes in [files](https://github.com/webmin/webmin/pull/2553/files).

1. It allows password reset over non-SSL only if the configuration explicitly permits it, in order to comply with general non-SSL support in Webmin.
2. It prefers domain name from a certificate over HTTP_HOST in `get_webmin_email_url`

> For sure, nothing in the current code prevents use of self-signed certs as long as they have a proper hostname.

Confirmed—it works!

>> The fix for this is here : https://github.com/webmin/webmin/commit/890a4ffd3f5480865629b5469876b63507f2cb64
> What SSL_HOST_CERT indicates is that the hostname in the HTTP request matches one of the hostnames in the certs. This is how we prevent an attacker sending a request with a fake hostname like evil.org , which is then included in the email.

After your latest fixes now it makes sense! Thanks!

> SSL_HOST does not need to have a port in it, because it's the hostname used to select the SSL cert (the TLS servername field), which does not have the same meaning as HTTP_HOST.

Yeah, I reverted that change too!

Though, I'd suggest fixing the variable names so they're not confusing. Maybe `SSL_DOM` and `SSL_DOM_CERT` would work better.

> You raise a good point about repeated reads of the same cert file though! I fixed that here : https://github.com/webmin/webmin/commit/d5a22a592fed2173661c389d80c943794da9f6e2

Thanks!

> As for using pure perl to read the cert, I wish that was possible but I don't see a way. This is why everywhere else in webmin and virtualmin we shell out to openssl.

We could try using the pure Perl implementation I sent you in the email yesterday in `eval` block, and fall back to your code if mine fails. This would probably be acceptable, and we could dump an error to the `miniserv.error` file to see if users ever report a problem with it.

jcameron:
Looks good, thanks!

iliaross:
> Looks good, thanks!

Thanks for merging it!

>> Though, I'd suggest fixing the variable names so they're not confusing. Maybe `SSL_DOM` and `SSL_DOM_CERT` would work better.

And this one I address here https://github.com/webmin/webmin/commit/a027ad5dd6c4de3e35d9483fdc5e266355218fe8 — we should make it clear that SSL host doesn't contain port, and is just a name, unlike HTTP_HOST.

jcameron:
> And this one I address here https://github.com/webmin/webmin/commit/a027ad5dd6c4de3e35d9483fdc5e266355218fe8 — we should make it clear that SSL host doesn't contain port, and is just a name, unlike HTTP_HOST.

Ok seems reasonable

PR #2558, opened 2025-09-25
Fix to remove significant bottleneck of shelling out

iliaross:
Hi Jamie!

I ran more tests and found that using the `openssl` call is a huge bottleneck.

> As for using pure perl to read the cert, I wish that was possible but I don't see a way. This is why everywhere else in webmin and virtualmin we shell out to openssl.

There is a way, and it's literally 106 times (not %) faster to use `Net::SSLeay` instead for the job to get `cn` and `alt` from the certificate. If the system has thousands of domains with thousands of SSL certificates, then it will be painful to use a call to `openssl`.

This PR addresses this issue.

Test example for 1000 iterations:

```
Time 1 (shellout) : total=6.096385s, avg=6.096385ms/call
Time 2 (pure)     : total=0.057364s, avg=0.057364ms/call
Result: 2 is 106.28x faster
```

Note that I will do something similar for Virtualmin, with performance gains about 40 times faster, but it will be trickier because we need to get the modulus and exponent too. This is a simplified version of the function compared to the upcoming PR for Virtualmin.

### DIFF
--- a/miniserv.pl
+++ b/miniserv.pl
@@ -4784,7 +4784,7 @@ eval 'Net::SSLeay::CTX_set_options($ssl_ctx,
         &Net::SSLeay::OP_NO_RENEGOTIATION)';
 
 # Get the hostnames each cert is valid for
-my $info = &cert_file_info($certfile);
+my $info = &cert_names($certfile);
 my @hosts;
 push(@hosts, $info->{'cn'}) if ($info->{'cn'});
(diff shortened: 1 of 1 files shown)


iliaross:
Related https://github.com/virtualmin/virtualmin-gpl/pull/1121.

jcameron:
Nice! This looks great!

PR #2565, opened 2025-10-02
Fix to redirect to HTTPS when we get the host from the browser URL

iliaross:
Hey Jamie,

This addresses the issue we discussed here:

https://github.com/webmin/webmin/commit/73cb67f02d246b08245c45bea84ee45adee5f8f9#r166966380

### DIFF
--- a/miniserv.pl
+++ b/miniserv.pl
@@ -1381,33 +1381,8 @@ $method = $page = $request_uri = undef;
 print DEBUG "handle_request reqline=$reqline\n";
 alarm(0);
 if (!$use_ssl && $config{'ssl'} && $config{'ssl_enforce'}) {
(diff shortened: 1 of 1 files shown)


miniserv.pl
```diff
@@ -1381,33 +1381,8 @@ sub handle_request
 print DEBUG "handle_request reqline=$reqline\n";
 alarm(0);
 if (!$use_ssl && $config{'ssl'} && $config{'ssl_enforce'}) {
-	# This is an http request when https must be enforced
```
jcameron, line 1384: Is this `if` block entirely empty now?

jcameron:
Thanks for the fix!

iliaross:
> Thanks for the fix!

Great! You're welcome!

PR #2581, opened 2025-11-09
New iteration allowing BSDfdisk module to work on FreeBSD devices. Using gpart instead of fdisk when possible.

karmantyu:
New iteration allowing BSDfdisk module to work on FreeBSD devices. Using gpart instead of fdisk when possible.
The creating, removing of slices and partitions are dangerous so test the proposed changes careful, I've done basic testing only!

### DIFF
--- a/bsdfdisk/slice_form.cgi
+++ b/bsdfdisk/slice_form.cgi
@@ -20,8 +20,8 @@ $disk || &error($text{'disk_egone'});
 my $base_device = $disk->{'device'}; $base_device =~ s{^/dev/}{};
 my $disk_structure = get_disk_structure($base_device);
 my $is_gpt = (is_using_gpart() && $disk_structure && ($disk_structure->{'scheme'}||'') =~ /GPT/i);
- 
- # Check if there is any free space on the device
+
+# Check if there is any free space on the device
 my $has_free_space = 0;
 if ($disk_structure && $disk_structure->{'entries'}) {
     foreach my $entry (@{$disk_structure->{'entries'}}) {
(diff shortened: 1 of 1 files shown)


jcameron:
Thanks for the contribution! To be honest we don't test much on FreeBSD, so we'll have to take your word that this is a bug-free patch. Which FreeBSD version did you test on though?

karmantyu:
I've tested it on FreeBSD 14.3. Done some partitioning on an USB pen-drive. Seems to work OK. Not tested the fallback if no gpart on the system present. If You would have a VM with old FreeBSD that would be a good testbed.

iliaross:
I think it would be right to support both cases, old and new.

Also, do we need to depend on the `use List::Util` library?

karmantyu:
I think it would work on old systems too, but could not test it. I kept in mind that systems without gpart should have fallback to fdisk.
No, we should not use List::Util. However only create_slice.cgi and edit_part.cgi using it to find the appropriate slices, and is more efficient this way. If you wish I can remove the use of the library.

jcameron:
Yes, it would be good to remove the dependency on `List::Util` if possible.

jcameron:
@iliaross do you think this is good to merge?

karmantyu:
Yes please.

iliaross:
> @iliaross do you think this is good to merge?

I haven't checked closely yet, as I'm busy with more important tasks. It'll have to wait.

karmantyu:
A small change was made.

karmantyu (commented):
Hope it's better now.

jcameron:
@karmantyu is this ready to merge in your opinion?

karmantyu:
> @karmantyu is this ready to merge in your opinion?

I think it is ready. Also be aware that bsdfdisk is tested only on FreeBSD. If the functioning on linux is also important for this module than we should work on that.

jcameron:
FreeBSD is all that matters for this module, as it won't be visible on Linux - the `fdisk` module is for that.

PR #2589, opened 2025-11-18
Add support for using pure-Perl Linux fallback to open PTY

iliaross:
Hey Jamie,

This removes dependency on IO::Pty as discussed earlier.

### DIFF
--- a/proc/proc-lib.pl
+++ b/proc/proc-lib.pl
@@ -249,53 +249,6 @@ do {    local $oldexit = $?;
 	} while($xp > 0);
 }
 
-
-# linux_openpty()
-# Linux-only, pure-Perl openpty(3)-style helper.
-# Returns master fh, slave fh, slave path, and ioctl value on success
-sub linux_openpty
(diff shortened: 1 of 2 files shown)


proc/proc-lib.pl
```diff
@@ -249,6 +249,53 @@ sub safe_exec_reaper
 	} while($xp > 0);
 }
 
+
+# linux_openpty()
+# Linux-only, pure-Perl openpty(3)-style helper.
+# Returns master fh, slave fh, slave path, and ioctl value on success
+sub linux_openpty
```
jcameron: if this is Linux specific, could you move it to proc/linux-lib.pl ?

This also removes the need for line 258.

iliaross: Agreed! Fixed now!

jcameron:
Thanks for this fix!

iliaross:
You're most welcome, Jamie! Thanks for reviewing it!

PR #2609, opened 2025-12-26
id_ed25519.pub support for authorized_keys update

hayden-t:


jcameron:
Thanks for this patch!

hayden-t:
NP,  also maybe the dropdown in "preferred key format" in sshd config could be updated to show the new format

hayden-t:
image

jcameron:
Good idea, I'll do that

Issue #2611, opened 2025-12-27
Unexpected HSTS header, (forced redirect to ssl), possible install setting mismatch


hayden-t:
So im just setting up a new post install 'fit out' of virtualmin under debian 13. And i was trying to set up some 3rd party panels under the main server domain, and when accessing them it kept redirecting to ssl, even when i specifically type in http.

Turns out it was due to a header i didnt know about called (HSTS) Strict-Transport-Security

After digging I found it is an option you can set in webmin called "Enforce SSL", and i figured it was coming from there. 

But looking at this setting it was set to option #1"Yes" which enforces SSL but doesnt add HSTS as well (option #2)

Which according to code shouldnt send/use the header. The logic seems to be if not #2 set HSTS to 0 (disable)

But where does HSTS come from when you do choose option 2 ? 
Possibly it is persisting from install, (code below seems to suggest it may be used in install?)

So maybe the setup setting and the ui default dont match (if thats possible) , as since changing the setting for the first time, the bug has gone away and its behaving as expected.

The way to investigate it further would be a fresh install again, but im not up for that, so im just posting my findings so for for others to find.

Strict-Transport-Security max-age=31536000;

<img width="390" height="254" alt="Image" src="https://github.com/user-attachments/assets/7d926f6e-a006-4145-b24d-ad25ac1eca19" />

https://github.com/webmin/webmin/blob/179d36859a14e0b3932bb2e77238ac3683733d16/webmin/change_ssl.cgi#L78

https://github.com/webmin/webmin/blob/179d36859a14e0b3932bb2e77238ac3683733d16/setup.sh#L839

jcameron:
@iliaross I think you added this feature in https://github.com/webmin/webmin/commit/c12cf25ecdf87d379ee7fb8069159e98f6fd4183

I don't get how this is supposed to work in practice though? I see that change_ssl sets the HSTS header, but won't that only effect the current browser session? What about other Webmin users, or logins from other browsers?

iliaross:
> I don't get how this is supposed to work in practice though?

This is a security feature. Once a browser sees an HSTS header for a domain, it will keep upgrading future connections to HTTPS for that domain for as long as the stored `max-age` is still in effect. It's not a "user preference" thing, it's the browser enforcing "always use HTTPS", no matter what, primarily to block protocol-downgrade and SSL-stripping style attacks.

[...]

PR #2612, opened 2025-12-30
incorrect ssl_enforce setting for HSTS

hayden-t:
https://github.com/webmin/webmin/issues/2611


iliaross:
Hello,

Thanks for the patch! You're right, it was a bug! Merging now!

PR #2628, opened 2026-02-10
Added missing macOS versions.

pbobbenb:
Added missing macOS versions.


jcameron:
Thanks for the patch!

PR #2638, opened 2026-03-10
Add API to exclude module calls from logging

iliaross:
Hey Jamie!

This PR adds support for a module to define a list of links that should not be logged. Currently, we are logging the 2FA secret, and that’s not right.

### DIFF
--- a/miniserv.pl
+++ b/miniserv.pl
@@ -3351,6 +3351,16 @@ sub reset_byte_count { $write_data_count = 0; }
 # byte_count()
 sub byte_count { return $write_data_count; }
 
+# get_module()
+# Returns module name from SCRIPT_NAME
+sub get_module
+{
+my $sn = $ENV{'SCRIPT_NAME'} || "";
+$sn =~ s/^\Q$config{'webprefix'}\E\//\//
+	if (!$config{'webprefixnoredir'} && $config{'webprefix'});
+return $sn =~ /^\/([^\/]+)\// ? ( $1 ) : ();
+}
+
 # log_request(hostname, user, request, code, bytes)
 # Write an HTTP request to the log file
 sub log_request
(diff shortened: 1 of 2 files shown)


jcameron:
There's got to be a better way of doing this! Can we not use a POST request for 2fa CGI script? In what situation is it being logged exactly?

iliaross:
> Can we not use a POST request for 2fa CGI script?

No, not easily. It's just a link to `/webmin/qr.cgi` file.

> There's got to be a better way of doing this!

Why is this bad, though? I think it’s good because it lets each module decide which calls shouldn’t be logged, and which are not. It also has a small footprint.

jcameron:
It's bad because calling Webmin code from Miniserv breaks the separation between the webserver and the other scripts, and only works by chance because Webmin CGI scripts aren't run in a separate process most of the time.

Other than logging, passing security-critical values (like tokens and passwords) via GET request params is a bad idea for many other reasons .. like cache server logs, bookmarks, browser history, etc.

iliaross:
> Other than logging, passing security-critical values (like tokens and passwords) via GET request params is a bad idea for many other reasons .. like cache server logs, bookmarks, browser history, etc.

Agreed, though I never implemented it like that. I was just trying to address the issue that was there before.

jcameron:
> Agreed, though I never implemented it like that. I was just trying to address the issue that was there before.

What I meant is, we need a fix beyond just removing 2fa codes from logging. Is there some JS trick we can do at https://github.com/webmin/webmin/blob/master/webmin/twofactor-funcs-lib.pl#L243 to make a POST request for the image? Or maybe encrypt the `str` parameter with the current session ID so even if it's logged it is useless?

iliaross:
> Is there some JS trick we can do at https://github.com/webmin/webmin/blob/master/webmin/twofactor-funcs-lib.pl#L243 to make a POST request for the image? Or maybe encrypt the `str` parameter with the current session ID so even if it's logged it is useless?

Yes, I can definitely do something else, except encrypting. 🙂 

I will close this PR and send you the link to the patch when ready...

jcameron:
Awesome thanks!

iliaross:
Alright, Jamie! All done https://github.com/webmin/webmin/commit/ed17ade5104a349c25834ceb8a286ee92129b09e.

jcameron:
Nice thanks!

PR #2646, opened 2026-03-22
FTP backup passive login error fix on IIS FTP Server

djdeeles:
On IIS FTP server fails to login. 
This fixes the following error.

FTP port failed : PASV failed : Please login with USER and PASS.


jcameron:
Isn't this going to produce an error if the Perl module needed for SSL is missing, even if the connection doesn't use SSL?

djdeeles:
Added ssl checks. If module not installed non ssl connection works as expected.

iliaross:
Wait, you can’t change the current logic or behavior.

Try this patch https://github.com/webmin/webmin/commit/3a3b202a96c7b828554e5ecd4e654a6f062f83de and let us know if it works for you.

Issue #2659, opened 2026-04-09
Edit Cronjob Page UI issue for mobile devices.


vipulksh:
Hi,
Thank you for supporting webmin/virtualmin.

I have noticed a UI issue for mobile devices where the buttons overlap one another. Can you check this?

<img width="220" height="350" alt="Image" src="https://github.com/user-attachments/assets/78d54604-425d-4ddf-9be1-9e8e1f64c687" />

jcameron:
@iliaross could you take a look at this PR?

iliaross:
Yes, already finishing the fix...

iliaross:
@jcameron Jamie, sorry for the delay—it took longer to fix properly.  

@vipulksh The fix will be included in Webmin 2.631+. Thanks for catching that.

vipulksh:
Thank you guys

jcameron:
Thanks @iliaross for the fix!

PR #2660, opened 2026-04-09
Improve reponsive UI for edit_cron

vipulksh:
Fix right-side buttons overlap issue for smaller screens on edit_cron page. Improved responsive ui.

Fix for this github issue raised. https://github.com/webmin/webmin/issues/2659


iliaross:
Thanks, but we’ve already fixed it by implementing a proper API for these scenarios, since Cron isn’t the only model that uses it.

PR #2665, opened 2026-04-11
Dhcp config list iface listen

vsc55:
New configuration to be able to define the interfaces that appear in the list in the "Edit Network Interface" option.
You can choose between only the active interfaces (current mode), or all the interfaces that appear in the system.

### DIFF
--- a/dhcpd/config-AlmaLinux-7.0-ALL
+++ b/dhcpd/config-AlmaLinux-7.0-ALL
@@ -14,3 +14,4 @@ group_name=0
 display_max=100
 desc_name=0
 lease_vendor=0
+list_listen_interfaces=0
--- a/dhcpd/config-Amazon-Linux-2-ALL
+++ b/dhcpd/config-Amazon-Linux-2-ALL
@@ -14,3 +14,4 @@ group_name=0
 display_max=100
 desc_name=0
 lease_vendor=0
+list_listen_interfaces=0
--- a/dhcpd/config-CentOS-Linux-5
+++ b/dhcpd/config-CentOS-Linux-5
@@ -15,3 +15,4 @@ group_name=0
 display_max=100
 desc_name=0
 lease_vendor=0
+list_listen_interfaces=0
--- a/dhcpd/config-CentOS-Linux-6.0-6.9
+++ b/dhcpd/config-CentOS-Linux-6.0-6.9
@@ -15,3 +15,4 @@ group_name=0
 display_max=100
 desc_name=0
 lease_vendor=0
+list_listen_interfaces=0
--- a/dhcpd/config-CentOS-Linux-7.0-ALL
+++ b/dhcpd/config-CentOS-Linux-7.0-ALL
@@ -14,3 +14,4 @@ group_name=0
 display_max=100
 desc_name=0
 lease_vendor=0
+list_listen_interfaces=0
--- a/dhcpd/config-CentOS-Stream-Linux-8.0-ALL
+++ b/dhcpd/config-CentOS-Stream-Linux-8.0-ALL
@@ -14,3 +14,4 @@ group_name=0
 display_max=100
 desc_name=0
 lease_vendor=0
+list_listen_interfaces=0
--- a/dhcpd/config-CloudLinux-8.0-ALL
+++ b/dhcpd/config-CloudLinux-8.0-ALL
@@ -14,3 +14,4 @@ group_name=0
 display_max=100
 desc_name=0
 lease_vendor=0
+list_listen_interfaces=0
--- a/dhcpd/config-Oracle-Linux-8.0-ALL
+++ b/dhcpd/config-Oracle-Linux-8.0-ALL
@@ -14,3 +14,4 @@ group_name=0
 display_max=100
 desc_name=0
 lease_vendor=0
+list_listen_interfaces=0
--- a/dhcpd/config-Redhat-Enterprise-Linux-6.0-6.9
+++ b/dhcpd/config-Redhat-Enterprise-Linux-6.0-6.9
@@ -15,3 +15,4 @@ group_name=0
 display_max=100
 desc_name=0
 lease_vendor=0
+list_listen_interfaces=0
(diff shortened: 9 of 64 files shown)


jcameron:
In what situation would the `active_interfaces` function return a different list from `list_interfaces` though?  Both should be built from the output of the `ifconfig -a` command.

vsc55:
When list_listen_interfaces=1, we intentionally want to show all system interfaces, not only active/configured ones. active_interfaces() may omit interfaces that are down or not fully configured, while list_interfaces() includes those as well. The stored value is still the interface name, so this change affects visibility in the selector, not how the interface is identified later.

jcameron:
By inactive you mean interfaces that are in the system config files, but not actually enabled on the system? Or interfaces in some disabled state?

vsc55:
I am referring, for example, to interfaces that are configured but in a down state at that moment.

imagen

jcameron:
But the `net::active_interfaces` function should return those already? Assuming they are included in the output of the `ip addr` command.

vsc55:
Yes, active_interfaces shows the active interfaces, but not the ones that are not active.
For example, if you are preparing a dhcp server, you cannot leave it already configured so that it works as soon as the interface is up since it does not appear in the list, you would have to put the interface up and then access the configuration and configure the interface.

Hence the option to use list_interfaces to get the complete list of network interfaces.

jcameron:
I would be interested to see the output from the commands `ifconfig -a` and `ip addr` on a system with a disabled interface, if possible.

vsc55:
The system is Debian 13.

IP eth0:0 is UP:
imagen
```bash
$ ifconfig -a
eth0: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1500
        inet 192.X.X.X netmask 255.255.255.0  broadcast X.X.X.X
        inet6 .......IVP6....  prefixlen 64  scopeid 0x20<link>
        ether ....MAC........  txqueuelen 1000  (Ethernet)
        RX packets 1743787  bytes 583457732 (556.4 MiB)
        RX errors 0  dropped 55  overruns 0  frame 0
        TX packets 779033  bytes 252294515 (240.6 MiB)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0

lo: flags=73<UP,LOOPBACK,RUNNING>  mtu 65536
        inet 127.0.0.1  netmask 255.0.0.0
        inet6 ::1  prefixlen 128  scopeid 0x10<host>
        loop  txqueuelen 1000  (Local Loopback)
        RX packets 85071  bytes 8129351 (7.7 MiB)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 85071  bytes 8129351 (7.7 MiB)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0

$ ip addr
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host noprefixroute
       valid_lft forever preferred_lft forever
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    link/ether ....MAC............ brd ff:ff:ff:ff:ff:ff
    altname enp0s18
    altname ens18
    inet 192.X.X.X/24 brd X.X.X.255 scope global eth0  <<<<<< eth0
       valid_lft forever preferred_lft forever
    inet 10.X.X.X/24 scope global eth0                 <<<<<< eth0:0
       valid_lft forever preferred_lft forever
    inet6 .....IPV6............./64 scope link proto kernel_ll
       valid_lft forever preferred_lft forever
```

IP eth0:0 is DOWN:
imagen
```bash

$ ifconfig -a
eth0: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1500
        inet 192.X.X.X netmask 255.255.255.0  broadcast X.X.X.X
        inet6 .......IVP6....  prefixlen 64  scopeid 0x20<link>
        ether ....MAC........  txqueuelen 1000  (Ethernet)
        RX packets 1743787  bytes 583457732 (556.4 MiB)
        RX errors 0  dropped 55  overruns 0  frame 0
        TX packets 779033  bytes 252294515 (240.6 MiB)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0

lo: flags=73<UP,LOOPBACK,RUNNING>  mtu 65536
        inet 127.0.0.1  netmask 255.0.0.0
        inet6 ::1  prefixlen 128  scopeid 0x10<host>
        loop  txqueuelen 1000  (Local Loopback)
        RX packets 85071  bytes 8129351 (7.7 MiB)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 85071  bytes 8129351 (7.7 MiB)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0


$ ip addr
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host noprefixroute
       valid_lft forever preferred_lft forever
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    link/ether ....MAC............ brd ff:ff:ff:ff:ff:ff
    altname enp0s18
    altname ens18
    inet 192.X.X.X/24 brd X.X.X.255 scope global eth0  <<<<<< eth0
       valid_lft forever preferred_lft forever
    inet6 .....IPV6............./64 scope link proto kernel_ll
       valid_lft forever preferred_lft forever
```

iliaross:
Thanks for the PR. The use case makes sense, but this implementation can’t be merged as-is.

The final `edit_iface.cgi` logic is inverted, i.e. `list_listen_interfaces=0` is labeled as the current active-only, but it calls `net::list_interfaces()`, which returns plain interface names, and then the loop treats those values as hashrefs via `$i->{'fullname'}`. The other branch gets hashrefs from `active_interfaces()` but treats them as strings.

Also, `net::list_interfaces()` only exists in the Linux backend, while this PR adds the new default config to non-Linux platforms too, which would break those DHCP interface pages.

Closing it.

PR #2666, opened 2026-04-11
Make get_webmin_browser_url proxy-aware

swelljoe:
This makes `get_webmin_browser_url` proxy-aware (`X-Forwarded-*` headers). This is needed for a WebSockets hardening PR I'm working on, and I think also fixes a display bug that's been reported a few times in various issues here. As far as I can tell, this has only one caller, so the impact of the change is tiny.

### DIFF
--- a/web-lib-funcs.pl
+++ b/web-lib-funcs.pl
@@ -13482,25 +13482,68 @@ return $url;
 =head2 get_webmin_browser_url([module], [cgi])
 
 Returns the URL for accessing this Webmin system, based on the current browser
-connection.
+connection. Honors X-Forwarded-Proto, X-Forwarded-Host and X-Forwarded-Port
+when Webmin is behind a reverse proxy, so the returned URL is the one the
+user's browser actually used.
 
 =cut
 sub get_webmin_browser_url
 {
 my ($mod, $cgi) = @_;
 
-# Work out the base URL`
-my $host = $ENV{'HTTP_HOST'};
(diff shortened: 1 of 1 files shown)


swelljoe:
Oh, also, I want to note that this function (and other places where X-Forwarded is looked it) explicitly trusts the proxy. That's potentially risky. We should probably add an option to Webmin to specify a trusted proxy (and add it to the proxy configuration documentation), so we can _know_ we're interacting with a trusted proxy and can trust what it tells us.

This specific function is currently only called for a display purpose, so that's not a current security issue.

iliaross:
Thanks for the patch, Joe. Yet, I think we shouldn’t trust forwarded URL headers unconditionally, since this is optional and controlled on the “Webmin ⇾ Webmin Configuration: IP Access Control” page. See this additional patch https://github.com/webmin/webmin/commit/97774b829d41d8cd406ed52ecc897c4df5661772.

swelljoe:
You're not understanding what I said. We _already_ trust them unconditionally in places where it's used. Unless you've addressed the other cases, you're not improving security by excluding this. This patch merely makes the host information display correctly when proxied.

swelljoe:
Oh, but I guess that patch addresses that problem in one case.

iliaross:
> We _already_ trust them unconditionally in places where it's used. Unless you've addressed the other cases, you're not improving security by excluding this.

Check this patch out https://github.com/webmin/webmin/commit/87d8969efbf217f889bd76d3be218b62ba305143.

PR #2676, opened 2026-04-26
Fix ping status PID check to use 16 bits for comparison

mnoorenberghe:
If `/proc/sys/kernel/pid_max` exceeds 16 bits then the ping check could incorrectly fail.

Source: https://man7.org/linux/man-pages/man8/ping.8.html#ID_COLLISIONS


jcameron:
Thanks for the patch!

PR #2691, opened 2026-05-10
Add trusted_proxies config

swelljoe:
Adds a `trusted_proxies` config option to Webmin->IP Access Control, the associated help files, an prevents enabling the "trust proxy" option without providing one or more proxies to trust. I have tested the UI part of it, but I don't have a proxy setup for testing the actual logic, but I've stared at it quite a bit, and it seems correct. This matches how nginx and Apache handle trusting proxies, with `set_real_ip_from` or `RemoteIPTrustedProxy` options, respectively.

Specific changes:

1472-1484 — new opt-in trusted_proxies gate. When set, requests whose direct TCP peer isn't in the list have $headerhost, X-SSL-Client-DN,  and X-SSL-Client-Verify dropped before the existing trust_real_ip block runs. No effect on installs that don't set trusted_proxies (backward-compatible by design).

1739 — /^success/i → /^success$/i. No longer matches "successful", "successive", etc.

1747-1748 — header-DN fallback now requires !($use_ssl && $verified_client). A real mTLS client whose cert doesn't map to a user can no longer pivot via X-SSL-Client-DN.

### DIFF
--- a/miniserv.pl
+++ b/miniserv.pl
@@ -1469,6 +1469,19 @@ if ($headerhost) {
 	$headerhost = undef if (!&check_ipaddress($headerhost) &&
 				!&check_ip6address($headerhost));
 	}
+# If trusted_proxies is configured, header-supplied client IP and SSL
+# client info are only honored when the direct TCP peer is in that list.
+# Otherwise drop them so an attacker reaching miniserv directly cannot
+# spoof X-Forwarded-For or X-SSL-Client-* to bypass auth.
+if ($config{'trust_real_ip'} && $config{'trusted_proxies'} ne '' &&
+    !&ip_match($acptip, $localip,
+	       split(/\s+/, $config{'trusted_proxies'}))) {
+	print DEBUG "handle_request: peer $acptip not in trusted_proxies; ".
+		    "ignoring forwarding and SSL client headers\n";
+	$headerhost = undef;
+	delete $header{'x-ssl-client-dn'};
+	delete $header{'x-ssl-client-verify'};
+	}
 if ($config{'trust_real_ip'}) {
 	$acpthost = $headerhost || $acpthost;
 	if (&check_ipaddress($headerhost) || &check_ip6address($headerhost)) {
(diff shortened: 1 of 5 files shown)


miniserv.pl
```diff
@@ -1723,16 +1736,20 @@ sub handle_request
 my $trust_ssl = $config{'trust_real_ip'} && !$config{'no_trust_ssl'};
 if ($use_ssl && $verified_client ||
     $trust_ssl && $header{'x-ssl-client-dn'} &&
-                  $header{'x-ssl-client-verify'} =~ /^success/i) {
+                  $header{'x-ssl-client-verify'} =~ /^success$/i) {
```
jcameron, line 1739: Isn't it possible for this header to contain some message after SUCCESS? The specs on it's format aren't well documented unfortunately.

jcameron:
This is excellent thanks! I really should have done this long ago!

Is there an RFC for this x-ssl-client-verify header though?

miniserv.pl
swelljoe: So, this is complicated, and seemingly not a standard.

nginx - exactly one of `SUCCESS`, `FAILED:<reason>`, or `NONE`. (ngx_http_ssl_module.c)
Apache - one of `NONE`, `SUCCESS`, `GENEROUS`, or `FAILED:<reason>`. (https://httpd.apache.org/docs/current/mod/mod_ssl.html note SSL_CLIENT_VERIFY values.)
HAProxy - Can be anything? Admins build their own string, could be anything. If we want to document it, we need to document that SUCCESS is the only SUCCESS response allowed.
Others - I dunno.

We only document nginx and Apache, and both will send SUCCESS, unless the user goes out of their way to do something different/unusual.

swelljoe: Oh, found docs for nginx, as well: https://nginx.org/en/docs/http/ngx_http_ssl_module.html (confirms SUCCESS with no trailing info)

jcameron: cool, in that case I think we can trust that the header will just be SUCCESS

PR #2692, opened 2026-05-10
Prefer SHA over MD5 or crypt for session hash, other minor session tweaks

swelljoe:
The biggest change here is preferring SHA for the session hash, and introducing a session key to allow logging out users (by deleting the session key and making a new one).

 `crypt` truncates to 8 bytes, so it's unsafe for session ID storage. Other change is permissions-related...making sure the db isn't umasked to something looser. The original preferred MD5 hash is fine for a session ID, but SHA is better.

This will invalidate existing sessions, so users need to log in again on systems that have Digest::SHA (as we want to upgrade their session), which should be all modern systems. I think even on RHEL, where Perl is broken into a bunch of tiny pieces, excluding several core modules from the `perl` package, I think Digest::SHA gets installed by default when you install Perl.

  Permissions fix (CWE-732):
  - New open_session_db() helper centralises the SDBM/NDBM probe, runs dbmopen under umask 0077, opens with mode 0600 (was 0700), then chmod 0600 on every  ${sessiondb}* file — so pre-existing loose perms get corrected on the first restart after upgrade.
  - All three dbmopen call sites (startup, reload, helper itself) now go through it.

  Session-ID hashing fix:
  - Digest::SHA is probed at startup; if hmac_sha256_hex works, $use_hmac_sha256 = 1.
  - New load_session_secret() reads or creates $var_dir/session.key (32 bytes from /dev/urandom, mode 0600). Idempotent across SIGUSR1 reloads, so sessions survive restarts. Path is overrideable via session_keyfile in config.
  - hash_session_id now returns hmac_sha256_hex($sid, $session_hmac_key) when available, falling back to the old MD5 / unix_crypt paths only on systems without Digest::SHA. Existing two-key lookup in verify/delete already tolerates the new format; pre-upgrade sessions stop validating.

### DIFF
--- a/miniserv.pl
+++ b/miniserv.pl
@@ -6619,8 +6619,7 @@ umask($oldumask);
 # open_session_db()
 # Opens the session DBM with a tight umask, then forces 0600 on the
 # resulting on-disk files so pre-existing loose perms from older installs
-# are corrected. Tries SDBM first, falling back to NDBM (matching the
-# original probe).
+# are corrected. Tries SDBM first, falling back to NDBM.
 sub open_session_db
 {
 my $oldumask = umask(0077);
--- a/miniserv.pl
+++ b/miniserv.pl
@@ -104,6 +104,14 @@ if (&unix_crypt_supports_sha512()) {
 	push(@startup_msg, "Using SHA512 via crypt() function");
 	}
 
+# Check if Digest::SHA with hmac_sha256_hex is available, for keying
+# the session-ID lookup table.
+eval "use Digest::SHA qw(hmac_sha256_hex); hmac_sha256_hex('x', 'y');";
+if (!$@) {
+	$use_hmac_sha256 = 1;
+	push(@startup_msg, "Using HMAC-SHA256 for session ID hashing");
+	}
+
 # Get miniserv's perl path and location
 $miniserv_path = $0;
 open(SOURCE, $miniserv_path);
@@ -605,17 +613,7 @@ if ($config{'logclear'}) {
 
 # Setup the logout time dbm if needed
 if ($config{'session'}) {
-	eval "use SDBM_File";
-	dbmopen(%sessiondb, $config{'sessiondb'}, 0700);
-	eval "\$sessiondb{'1111111111'} = 'foo bar';";
-	if ($@) {
-		dbmclose(%sessiondb);
-		eval "use NDBM_File";
-		dbmopen(%sessiondb, $config{'sessiondb'}, 0700);
-		}
-	else {
-		delete($sessiondb{'1111111111'});
-		}
+	&open_session_db();
 	}
 
 # Run the main loop
@@ -5011,7 +5009,7 @@ print DEBUG "in reload_config_file\n";
 &parse_websockets_config();
 if ($config{'session'}) {
 	dbmclose(%sessiondb);
-	dbmopen(%sessiondb, $config{'sessiondb'}, 0700);
+	&open_session_db();
 	}
 print DEBUG "done reload_config_file\n";
 }
@@ -5095,6 +5093,10 @@ my $var_dir = $1;
 if (!$config{'sessiondb'}) {
 	$config{'sessiondb'} = "$var_dir/sessiondb";
 	}
+if (!$config{'session_keyfile'}) {
+	$config{'session_keyfile'} = "$var_dir/session.key";
+	}
+&load_session_secret();
 if (!$config{'errorlog'}) {
 	$config{'logfile'} =~ /^(.*)\/[^\/]+$/;
 	$config{'errorlog'} = "$1/miniserv.error";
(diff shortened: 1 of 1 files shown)


iliaross:
Thanks, Joe! I think this is a good improvement!

However, I think, you should still consider using existing active sessions, as after the upgrade and Miniserv restart, `sessiondb` entries keyed with the old crypt hash become invisible, effectively logging everyone out right after Webmin restarts, which in the UI can happen while other packages are being updated.

And, one more small nit: your new `load_session_secret()` is called unconditionally from `update_vital_config()`, so this will create `session.key` even when session auth is disabled. It’s mostly harmless, but feels like an unnecessary filesystem side effect for configs that don’t use sessions (which shouldn't be the case nowadays).

At last, as it’s not a critical change, so I think it would be better to schedule it for Webmin 2.650 instead of today’s 2.641 release, since changes to Miniserv should get more thorough testing over a few days or better weeks.

swelljoe:
> this will create session.key even when session auth is disabled.

We support auth without sessions?

> you should still consider using existing active sessions, as after the upgrade and Miniserv restart, sessiondb entries keyed with the old crypt hash become invisible, effectively logging everyone out right after Webmin restarts

That's intentional, though? I mean, we want to upgrade everyone to the new thing.

> which in the UI can happen while other packages are being updated

Which shouldn't be a problem, right?

iliaross:
> We support auth without sessions?

Yeah, though it's not an option on the UI, it's still possible to use.

> That's intentional, though? I mean, we want to upgrade everyone to the new thing.

Yeah, eventually, but not by dropping the connection during a package upgrade. 🙂 

> Which shouldn't be a problem, right?

I think it’s a problem for the user experience, and, most importantly, there’s no real urgency to do it.

jcameron:
I'm actually fine with forcing users to re-login here if it's more secure, since it would be a one-time thing. And the permissions fix is great, thanks!

That said, I forget why Webmin hashes the session ID in the first place!

swelljoe:
Yeah, I guess any (actually) random string could work. AFAIK it needs a good chunk of entropy, something like 64 bits bits. `crypt` can never be more than 32 bits, so that's not safe. MD5 was fine, a long string from urandom would also be fine, I think. So, maybe this could be simplified? We maybe can't count on a source of real randomness on embedded devices? https://cheatsheetseries.owasp.org/cheatsheets/Session_Management_Cheat_Sheet.html

Lemme think on that and do a little more reading. I don't actually know what the right thing is with session IDs, but if we can simply pull a 64 characters from urandom for the session ID, that seems simpler, and simpler is better than a bunch of code and an extra dependency.

swelljoe:
Oh, wait, no. We need to hash it, so it's not stored on disk in a workable form, same as /etc/shadow, etc. The idea being if the db is compromised in some way (via a system backup or something), it doesn't matter, cookies still can't be forged. So, that's why you originally hashed it with MD5 or crypt. MD5 is still probably fine for this particular use case, while crypt is definitely not.

So, yeah, I think this is no more complicated than necessary.

iliaross:
On the other hand, if `sessiondb` is readable only by root on the live system, and an attacker can read it, they’ve probably already won, so hashing session IDs won’t help much?

jcameron:
Now I think about it some more, the best argument for hashing is the same as /etc/shadow - an attacker may somehoe get permissions to read arbitrary files, but not write them. So if sessions are securely hashed, they cannot re-create a valid session cookie even with access to the file.

iliaross:
> So if sessions are securely hashed, they cannot re-create a valid session cookie even with access to the file.

Why would they need a valid session cookie in the first place if they could just create a new one, right? To be clear, I'm not saying we shouldn’t hash the session ID—just questioning the reasoning behind it...

...also, hashing passwords are a different story. A session cookie ID is a random string, while passwords are often reused by users on other websites.

jcameron:
Anyway I am going to merge this as it seems like a good idea regardless

swelljoe:
You can't possibly be arguing against a more secure solution, Ilia! Hashed session IDs in the database is the right thing. There are paths by which the session db could be compromised without the root user being compromised (I mentioned one above). There's a reason `/etc/shadow` is hashed, and there's a reason hashed password storage is required by every security standard. It was right for Jamie to hash them before, and it's right to improve the hashing now.

> I think it’s a problem for the user experience, and, most importantly, there’s no real urgency to do it.

Forcing a new login is not a user experience problem, as long as it's documented. It's common. There's no major sites that allow permanent logins. And, without forcing it, the old and less secure session ID hashes will effectively never go away. We have to purge them and force a new login at _some point_, if we want better security for session cookies (and we do).

We have to force a logout to enable this security enhancement. That's not optional. If we don't do it on first service restart, we have to do something more complicated and I don't think doing something unnecessarily complicated is good for the code.

jcameron:
I plan to merge this once Webmin 2.641 is out

swelljoe:
Oh, I was going to mention I can wrap the session key creation in a check if session auth is enabled, as Ilia's right that creating a file that never gets used is suboptimal.

What are the other ways auth happens? What option decides that?

swelljoe:
OK, I've gated the session key file creation behind `if ($config{'session})`.

So, I think we're probably good to merge.

PR #2695, opened 2026-05-11
Wrap miniserv.pl server loop in unless (caller) so we can require miniserv for unit testing

swelljoe:
Wraps the main part of this script in `unless (caller)`, which means that if the script is started by systemd, an initscript, or from the command line, it will start miniserv.pl normally. If the script is called by another Perl script with `require` or `do` it will define the globals and the subs within the file, so they can be exercised by unit tests.

I lifted some of the globals higher up so they're outside of the block, so they'll all be visible to functions.

Unit tests will have to setup their own `%config` hash and any other state needed by the function being tested, but that's what we want anyway, so we can feed it fake config values...we don't want to read/write from `/etc` during tests.

It is possible there are other things defined in that new block that will be needed by functions when used as a module, but that doesn't effect running it as a command. We'll sort it out when we find a test failing because of it.

Adds `t/miniserv-http_error.t` as an example unit test, that exercises all of the paths, and confirms it's spitting out some reasonable HTML and status. It shows how the config hash can be mocked. Tests run with `prove` in the repo root, assuming Test::More and the other core Perl Test functionality is present.

```
$ prove
t/miniserv-http_error.t .. ok
All tests successful.
Files=1, Tests=32,  0 wallclock secs ( 0.01 usr  0.00 sys +  0.06 cusr  0.00 csys =  0.07 CPU)
Result: PASS
```


iliaross:
Jamie, see this comment where I share my concerns and suggest making these changes some time later:

https://github.com/webmin/webmin/pull/2701#issuecomment-4435333973

iliaross:
See this: https://github.com/virtualmin/virtualmin-gpl/pull/1216#issuecomment-4445684982.

iliaross:
Do we still really need this `unless (caller)` block?

PR #2698, opened 2026-05-12
perlcritic fixes

swelljoe:
Fixes perlcritic errors in nftables.

`return undef` is a bug waiting to happen. It behaves differently depending on caller. e.g. `@varname = nft_version_txt()` would create an array with one element `[ undef ]`, which tests as true in Perl, which is definitely not what you want/expect. Hash context is even weirder. Only scalar context actually works as intended.


jcameron:
Huh, I didn't realize that was best practice! Looks good to me ..

PR #2700, opened 2026-05-12
Revert "perlcritic fixes"

iliaross:
Hello,

I'm reverting this!

I'm begging everyone to stop submitting and merging code that’s actually broken!

image

### DIFF
--- a/nftables/acl_security.pl
+++ b/nftables/acl_security.pl
@@ -3,7 +3,7 @@ use warnings;
 no warnings 'redefine';
 no warnings 'uninitialized';
 
-require './nftables-lib.pl';    ## no critic (Modules::RequireBarewordIncludes)
+require 'nftables-lib.pl';
 our (%in, %text);
 
 # acl_security_form(&options)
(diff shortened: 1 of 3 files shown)


swelljoe:
I tested the module itself after changes, didn't think to check the ACL page.

But, why does this one `require` need to be different from the others? What code makes that path, and why would it need different semantics?

swelljoe:
But, also reverting it undoes necessary fixes to fix one line. Why not just fix that one line?

iliaross:
> But, also reverting it undoes necessary fixes to fix one line. Why not just fix that one line?

Alright, I reverted again this time, keeping the good changes.

And, initially, I reverted everything at once because it was one large commit.

> But, why does this one require need to be different from the others? What code makes that path, and why would it need different semantics?

Because `acl_security.pl` (and also `config_info.pl`) are helper files loaded by another Webmin page (like `acl/`), not run from inside `nftables/`. So Webmin loads `nftables/acl_security.pl`, but the current directory is still the caller’s directory, like `acl/`. Therefore `./nftables-lib.pl` points at the wrong place, because the `acl` module does not have `nftables-lib.pl`. On the other hand, module CGIs like `create_table.cgi` or `setup.cgi` run inside `nftables/`, so `./nftables-lib.pl` works there.

PR #2701, opened 2026-05-12
Make core pl and cli files testable

swelljoe:
Makes the base directory Perl files, and CLI files in bin, testable by making them loadable without executing the code (wraps in a `caller` check). Makes no difference in functionality/behavior, just means we can write tests for the functions in the files. This is the same change I made in #2695 applied to all the other stuff in the base dir and bin. I didn't make the change for any files that don't have any subs. Those just have to be tested via regular command invocation, for now.

Also adds `t/compile.t` to run `perl -c` on all Perl files. The most basic of smoke tests to make sure every Perl file has valid syntax. Doesn't really do much, but we have occasionally had invalid Perl files sneak out into releases, and we have to start somewhere. Skips `pl` files that are config or language or help, as `pl` in those cases means Polish not Perl.

Also adds a README.md to `t` documenting how to run tests, vaguely how to write tests (modules get a `t` directory of their own, where any tests not already covered at the higher level can be placed).

Also excludes `t` and `xt` from all packaging scripts.

### DIFF
--- a/bin/language-manager
+++ b/bin/language-manager
@@ -315,7 +315,7 @@ sub main
 
     return 0;
 }
-main();
+main() if !caller(0);
 
 # The following is used to get correct language files based on old language table.
 # This function, should not be really needed, unless compiling languages with
--- a/changepass.pl
+++ b/changepass.pl
@@ -2,6 +2,8 @@
 # changepass.pl
 # Script for the user to change their webmin password
 
+unless (caller) {
+
 # Get Webmin directory
 my $cwd = $0;
 $cwd =~ s/(.*)\/.*/$1/;
@@ -20,6 +22,8 @@ if ($status != 0) {
 		}
 	}
 exit $status;
+
+} # end of unless (caller)
 
 sub usage
 {
--- a/copyconfig.pl
+++ b/copyconfig.pl
@@ -4,6 +4,8 @@
 # directory. If it is already there, merge in new directives. Called with
 # <osname> <osversion> <install dir> <config dir> <module>+
 
+unless (caller) {
+
 @ARGV >= 4 || die "usage: copyconfig.pl <os>[/real-os] <version>[/real-version] <webmin-dir> <config-dir> [module ...]";
 $os = $ARGV[0];
 $ver = $ARGV[1];
@@ -112,6 +114,8 @@ foreach $m (@mods) {
 		}
 	}
 print join(" ", @newmods),"\n";
+
+} # end of unless (caller)
 
 # read_file(file, &hash, [&order], [lowercase], [split-char])
 # Fill the given hash reference with name=value pairs from a file.
--- a/create-module.pl
+++ b/create-module.pl
@@ -3,6 +3,8 @@
 # Creates a single .wbm file containing multiple modules, possibly with
 # forced versions
 
+unless (caller) {
+
 @ARGV >= 2 || die "usage: create-module.pl [--dir name] <file.wbm> <module>[/version] ..";
 
 my $pwd;
@@ -105,6 +107,8 @@ if ($createsig) {
 	system("gpg ".($keyname ? " --default-key $keyname" : "").
 	       " --armor --output $file-sig.asc --detach-sig $file");
 	}
+
+} # end of unless (caller)
 
 # read_file(file, &assoc, [&order], [lowercase])
 # Fill an associative array with name=value pairs from a file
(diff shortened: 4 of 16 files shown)


iliaross:
I’m worried this is not purely mechanical. `unless (caller)` changes load behavior. When loaded from another Perl file via `require` or `do` caller is true, so the main body is skipped and only the sub definitions are loaded. Perhaps a better way to do it is:

```
sub main {
    ...
}

exit main(@ARGV) unless caller;
1;
```

But there is also a catch, as it introduces a new scope, which can be a regression if helper subs depend on top-level lexicals. Since many of these scripts are old Perl and not always under strict, this can fail silently.

Please, I suggest holding off on these changes for now! Let’s first release Webmin 2.642 and Virtualmin 8.2.0 with the new Virtualmin Podman plugin, make sure everything is stable, and then move on to these riskier changes that need _much_ more testing.

swelljoe:
Where appropriate this PR uses that convention.

I think you're making this more dramatic than it is? Are there actually places where that happens (require/do of a file that has immediately executable stuff in it)? That would be a really ugly usage, having a file both execute things and also include functions and then include it as a library.

swelljoe:
I mean, I'm OK using the other convention, in cases where it's necessary. But, I'm pretty sure I've lifted everything that needs to be lifted into file scope.

We need a test suite so we can make a bunch of changes with confidence. You're trying to push back the very thing we need first. If it causes breakage, we fix it. But, the only way to know if anything causes breakage is to have a good test suite. So, the priority should be on getting the test suite.

swelljoe:
OK, I checked every file.

Verified across both /home/joe/src/webmin and /home/joe/src/virtualmin-gpl: no require or do callers exist that expect side-effect execution of these scripts. The refactor is safe.

  Concretely, for each of miniserv.pl, bin/language-manager, changepass.pl, copyconfig.pl, create-module.pl, extract-module-info-langs.pl, fix-english.pl, makedebian.pl, makedist.pl, makemoduledeb.pl, makemodulerpm.pl,
  oschooser.pl, setup.pl, thirdparty.pl, upload-pod-docs.pl:

  - Every cross-file invocation is a subprocess spawn (perl …, system(...), backticks, or setup.sh) — those are unaffected, since spawning a child interpreter still runs the file normally.
  - The only intra-Perl loads are setup.pl requiring helper libs (setup-pre.pl, setup-post.pl, web-lib-funcs.pl) — none of which were wrapped.
  - The one require that does target a wrapped file is t/miniserv-http_error.t, which is the test that motivated the change and wants definitions only. Working as intended.

  Nothing to fix.

swelljoe:
We actually do have problems using the `unless (caller)` convention in Virtualmin (but nothing in Webmin). The way Virtualmin CLI scripts are run from `json-lib.pl` expects `do` to run the script, which is pretty hairy, and I'm not overly fond of it. Looks like Cloudmin will probably have the same thing.

swelljoe:
Note that `exit main(@ARGV) unless caller;` also doesn't work for this problem, as there is a caller in this case. This is just a weird usage, so, I'll just work around it unless/until it changes. But, not having tests isn't a solution.

swelljoe:
OK, I've spent another hour or so spelunking, trying to find any other callers that could be effected, and nothing shows up. I've also asked every frontier model I have access to (Opus 4.7, Gemini Pro 3.1, GPT 5.4), plus some new near frontier models (Ring 2.6, Qwen 3.6), and all agree that there are no callers that would be broken by this change. Claude found the problem in Virtualmin immediately and that case is indirect and weird, so I think it knows how to find this problem, but I'll ask the other models about Virtualmin, as well, and do some more spelunking on my own.

Anyway, I genuinely believe this is a safe change in Webmin, as is.

swelljoe:
I'll also request a Copilot code review, just to give the models one more chance to find any problems with these changes.

iliaross:
See this: https://github.com/virtualmin/virtualmin-gpl/pull/1216#issuecomment-4445684982.

jcameron:
I wonder if instead of using `unless (caller)` we could extract the perl subs from these files (which I assume is what you want to test?) into a separate `-lib.pl` file, and have it called from the original file? The tests could then just operate on the `-lib.pl` file.

swelljoe:
That would also be fine. And, yes, the point is to be able to unit test all subs. But, there's really no harm in wrapping in a caller check (except for the exceptions noted in the Virtualmin PR).

The goal is to make everything testable with as few changes as possible, then add a gazillion unit tests and static analysis, then we can refactor freely with more confidence we're not breaking anything.

It could also be done with a variable check, like the way I changed the Virtualmin patch to accommodate the `execute_webmin_script` function in `json-lib.pl` which expects side effects of using `do` on the module. That doesn't rely on caller, at all. One has to explicitly set a variable to get different behavior.

Note that if we are refactoring and splitting libs off of executable code, it's a pretty common convention to wrap the main functionality in scripts in a `main` or similar, anyway, so they can also be loaded/mocked and exercised as a library (CLI tools can be tested via a shell callout, but it can be nicer to be able to host it in Perl, so you can set a fake environment and variables before calling main and capture results, etc.

Honestly, though, the primary concern in Webmin was miniserv.pl, as it's a security boundary, and that PR has been merged. All the rest of these are run locally and aren't security sensitive, so it doesn't matter. the Virtualmin changes are more of a concern. So, forget it.

We're wasting a bunch of time arguing about something I assumed would be trivial.

PR #2708, opened 2026-05-17
Exclude t and xt from packages

swelljoe:
Excludes tests in either `t` or `xt` from all package types.


iliaross:
LGTM, merging!

[Jamie](https://github.com/jcameron), quick question for later. In `makemoduledeb.pl`, the `--dsc-file` source tarball is still built from the original module dir, not the cleaned staged tree, so `t/xt` (and `.git`, `RELEASE`, etc.) still end up in it.

Is it intentional, or something we missed because we always build the DEB first and then the RPM overwrites the same `tar.gz` with a clean version and we never noticed?

jcameron:
There certainly shouldn't be any cross-dependency between the deb and RPM package builds.

So yeah, when `--dsc-file` is given, the cleaned staged directory should be used.

iliaross:
Thanks, I've fixed that in the linked commit.

PR #2709, opened 2026-05-17
More miniserv.pl tests

swelljoe:
More test coverage for `miniserv.pl`.

Currently failing due to SHA512 usage bug. Will make another PR to fix that bug and make this failing test pass. But, we don't gate on failing tests yet, so no reason not to merge, as-is.


iliaross:
Thanks!

PR #2712, opened 2026-05-17
Fix encrypt_sha512 so it doesn't always fall back to crypt

swelljoe:
encrypt_sha512 now matches libc crypt() output byte-for-byte for $6$ inputs, and password_crypt's verification round-trip still behaves as callers expect (returns the stored hash on correct password, differs on wrong).

  Diff summary:
  - miniserv.pl:6783 — replaced the regex-extract-and-reassign with the same form used in useradmin/md5-lib.pl:218: pass the salt through to crypt() unmodified, only synthesize a fresh $6$…$ when the caller didn't supply one.
  - No changes needed elsewhere. All five callers within miniserv.pl (one in password_crypt, four password_crypt consumers) use the contract &password_crypt($pass, $stored) eq $stored, which still holds. External-module callers of encrypt_sha512 (acl/, useradmin/, ldap-useradmin/) were already routing to useradmin::encrypt_sha512 and never touched the broken function.

Note, this is duplicated code with the same function in `useradmin/md5-lib.pl`. Not sure if there's an easy/safe way to dedupe.

### DIFF
--- a/miniserv.pl
+++ b/miniserv.pl
@@ -6779,15 +6779,14 @@ return $newhash eq $hash;
 }
 
 # encrypt_sha512(password, [salt])
-# Hashes a password, possibly with the given salt, with SHA512
+# Hashes a password, possibly with the given salt, with SHA512. The salt
+# arg may be a full $6$salt$hash form (verification) or a bare $6$salt$
+# (fresh hashing) — either way it must be passed to crypt() intact so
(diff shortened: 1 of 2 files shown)


iliaross:
It looks fine to me. Though, I’ll let Jamie merge this one.

swelljoe:
Yeah, I want Jamie in the loop on all security boundary merges.

jcameron:
Thanks for this fix!

PR #2719, opened 2026-05-18
Fix to preserve Netplan YAML when saving DNS

iliaross:
Hey, Jamie!

I managed to reproduce the [bug](https://forum.virtualmin.com/t/subject-critical-bug-webmin-network-module-corrupts-netplan-yaml-on-ubuntu-24-04-causing-total-network-interface-loss/137242/3?u=ilia) and fix it.

Now we to preserve existing Netplan indentation when updating DNS settings, validate with `netplan generate` before applying, and surface apply errors to avoid network loss from malformed YAML.

The config I used to repro the bug was:

```yml
# This is the network config written by 'subiquity'
network:
  ethernets:
    enp0s5:
      dhcp4: true
      dhcp6: true
      match:
        macaddress: 00:1c:42:ea:ab:90
      set-name: enp0s5
      nameservers:
        addresses:
          - 1.1.1.1
        search:
          - example.com

    webmin-unused0:
      optional: true
      dhcp4: false
      match:
        macaddress: 02:00:00:00:00:ff
      set-name: webmin-unused0
  version: 2
```

### DIFF
--- a/net/netplan-lib.pl
+++ b/net/netplan-lib.pl
@@ -165,6 +165,30 @@ foreach my $f (glob("$netplan_dir/*.yaml")) {
 return @rv;
 }
 
+# yaml_directive_indent(&directive, default-spaces)
+# Returns the indentation to use for an existing or new YAML directive.
+sub yaml_directive_indent
+{
+my ($yaml, $def) = @_;
+my $indent = $yaml && defined($yaml->{'indent'}) ? $yaml->{'indent'} : $def;
+return " " x $indent;
+}
+
+# yaml_child_indent(&directive, default-spaces)
+# Returns the indentation step used by an existing YAML directive's children.
+sub yaml_child_indent
+{
+my ($yaml, $def) = @_;
+if ($yaml && $yaml->{'members'}) {
+	foreach my $m (@{$yaml->{'members'}}) {
+		if (defined($m->{'indent'}) && $m->{'indent'} > $yaml->{'indent'}) {
+			return $m->{'indent'} - $yaml->{'indent'};
+			}
+		}
+	}
+return $def;
+}
+
 # save_interface(&details, [&all-interfaces])
 # Create or update a boot-time interface
 sub save_interface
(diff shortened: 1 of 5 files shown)


net/netplan-lib.pl
```diff
@@ -610,7 +637,29 @@ sub network_config_files
 # Apply the interface and routing settings
 sub apply_network
 {
-&system_logged("(cd / ; netplan apply) >/dev/null 2>&1");
+my $err = &validate_netplan_config();
+return $err if ($err);
+my $out;
+my $cmd = "(cd / && ".quotemeta(&netplan_command())." apply)";
```
jcameron: There's no need for quotameta here since we contol the command, and also it might contain spaces since it's set in the module config

net/netplan-lib.pl
```diff
@@ -610,7 +637,29 @@ sub network_config_files
...
+my $rv = &execute_command_logged($cmd, undef, \$out, \$out);
+return $rv ? $out || "netplan apply failed" : undef;
+}
+
+# validate_netplan_config()
+# Check that the current YAML can be rendered before applying it.
+sub validate_netplan_config
+{
+my $out;
+my $cmd = "(cd / && ".quotemeta(&netplan_command())." generate)";
```
jcameron: Ditto about the quotemeta

jcameron:
Why did the original code that indents based on the number of parent directives not work though? I am surprised that it is necessary to look at the parent's indent level.

iliaross:
> Why did the original code that indents based on the number of parent directives not work though? I am surprised that it is necessary to look at the parent's indent level.

The old code assumed 4 spaces per level, but `subiquity` and `cloud-init` Netplan files often use 2. As a result, Webmin rewrote only the edited interface at the wrong indentation level breaking the YAML. YAML hierarchy is column-based, not depth-based.

jcameron:
Oh god, that's terrible! I had no idea the indents weren't consistent!

I will approve this PR once you fix those two quotameta issues.

iliaross:
Thanks, Jamie!

I have already dropped them https://github.com/webmin/webmin/pull/2719/commits/7f63875c42eb935a3b08e0b8e45d34a29b0bfd6e.

PR #2720, opened 2026-05-19
Add some acl unit tests

swelljoe:
Tests most of the critical and security-adjacent functions in `acl`.

Note there are two bugs found while testing (not fixed in this PR):

1. parse_webmin_log interpolates $object directly into `<tt>...</tt>` in the delete, modify rename, joingroup, and twofactor branches — XSS if a username (or other object string) ever contains HTML. Usernames are normally restricted, but this is defense-in-depth. One TODO test documents the delete-branch case.
2. get_users_group returns the failed-loop condition ('') instead of undef when no match. Cosmetic but surprising; the test asserts !get_users_group(...) to stay honest about the current contract.

### DIFF
--- a/acl/t/run-tests.t
+++ b/acl/t/run-tests.t
@@ -1,0 +1,1349 @@
+#!/usr/bin/perl
+use strict;
+use warnings;
+use Test::More;
+use Cwd qw(abs_path);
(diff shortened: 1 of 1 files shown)


iliaross:
Thanks, Joe!

PR #2721, opened 2026-05-19
Fixes minor bugs in acl surfaced by tests

swelljoe:
Fixes the two bugs in `acl` surfaced by the unit tests introduced in #2720.

Changes:

 - acl/log_parser.pl — every interpolated user-controlled value now flows through &html_escape. The four branches you flagged (delete, modify rename, joingroup, twofactor) plus acl and reset, which had the same <tt>$object</tt> pattern. The bulk-delete users/groups form (log_delete_users/_groups, where $object is space-joined names) is also escaped now. I extended past your four because leaving an identical-looking pattern unfixed two lines below the fixed ones is the kind of inconsistency that bites later — and it matches the file's existing escape convention.
 - acl/acl-lib.pl — get_users_group now has an explicit return undef;. Both callers (save_user.cgi:106, delete_users.cgi:41) only use the result in boolean tests and hashref dereferences, so the contract change is invisible to them.
 - Tests — dropped the TODO: wrapper, replaced the single delete-branch XSS check with a data-driven loop covering all nine fixed sites (delete-user, delete-users-bulk, modify-rename old/new, joingroup object/group, twofactor object, acl object, reset object), tightened the get_users_group miss to is(..., undef, ...). Removed the comment that documented the broken contract.


acl/acl-lib.pl
```diff
@@ -1809,6 +1809,7 @@ sub get_users_group
 		return $g;
 		}
 	}
+return undef;
```
iliaross: Why `return undef` and not `return`?

swelljoe: lol, bug-compatible implementation! See, this is what I was talking about. The models imitate bad choices if not prevented by good static analysis tools. I'll fix that. (It would also be caught by the incoming perlcritic PR, but no reason to introduce new bug-likes while fixing old ones.)

acl/log_parser.pl
```diff
@@ -16,8 +16,9 @@ sub parse_webmin_log
 my $g = $type eq 'group' ? "_g" : "";
 if ($action eq 'modify') {
 	if ($p->{'old'} ne $p->{'name'}) {
-		return &text('log_rename'.$g, "<tt>$p->{'old'}</tt>",
-					      "<tt>$p->{'name'}</tt>");
+		return &text('log_rename'.$g,
+			     "<tt>".&html_escape($p->{'old'})."</tt>",
+			     "<tt>".&html_escape($p->{'name'})."</tt>");
```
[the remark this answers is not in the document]
swelljoe: @copilot apply changes based on [this feedback](https://github.com/webmin/webmin/pull/2721#discussion_r3269367752)

acl/acl-lib.pl
iliaross: Cool, thanks! 😅

acl/acl-lib.pl
```diff
@@ -1809,7 +1809,7 @@ sub get_users_group
 		return $g;
 		}
 	}
-return undef;
+return;
```
iliaross, line 1812: I think it’s basically the same as having no `return`, so the old logic applies, right?

Though, either way, I personally prefer to explicitly `return` because it looks cleaner.

swelljoe: It is not the same. The old was returning `''` (empty string).

swelljoe: Which, to be clear, has the same kinds of problems as `return undef`. It behaves differently depending on context. It's not truthy in a scalar, but it is truthy in an array/hash.

```
joe@cat:~/src/snippets$ cat return-empty-string.pl
#!/usr/bin/perl
use warnings;
use strict;

sub unexpected {
        return '';
}

my @array_var = unexpected();

if (@array_var) { print "Why are we here?\n" };

my %hash_var = unexpected();

if (%hash_var) { print "Once again, what?\n" };

my $scalar_var = unexpected();

if ($scalar_var) { print "Ah, nothing to do.\n" };
joe@cat:~/src/snippets$ perl return-empty-string.pl
Why are we here?
Odd number of elements in hash assignment at return-empty-string.pl line 13.
Once again, what?
```

And, the corrected `return;`:

```
joe@cat:~/src/snippets$ cat return-nothing.pl
#!/usr/bin/perl
use warnings;
use strict;

sub unexpected {
        return;
}

my @array_var = unexpected();

if (@array_var) { print "Why are we here?\n" };

my %hash_var = unexpected();

if (%hash_var) { print "Once again, what?\n" };

my $scalar_var = unexpected();

if ($scalar_var) { print "Ah, nothing to do.\n" };
joe@cat:~/src/snippets$ perl return-nothing.pl
```

iliaross: > It is not the same. The old was returning `''` (empty string).

Are you sure about that? As far as I know, in Perl, that returns an empty list `()` in list context, or `undef` in scalar context, just like when the sub reaches the end without an explicit `return`.

So adding `return` at the end or having no `return` is effectively the same—that's what I meant. 

However, I'm not sure if the tests are okay with having a sub without an explicit return?

swelljoe:
OK, I think this is ready to merge. I also checked all the callers of the changed `get_users_group` function to make sure none of them are expecting the weird array/hash result, and there aren't any.

So, this is a safe merge, with no changes in behavior from current usage (`return;` and `return '';` and `return undef;` are all falsy in scalar context, while only `return;` is falsy in array/hash context), it just protects against surprises in the future and will be one less thing to fix in the `perlcritic` pass.

iliaross:
Cool, thanks, Joe!

PR #2723, opened 2026-05-20
acl perlcritic test and fixes

swelljoe:
Adds `t/perlcritic.t` to `acl` module, makes all module files pass perlcritic (severity 5), and adds a `.perlcriticrc` to exclude `ProhibitLeadingZeros` (as making that rule happy makes octal usage in directory/file functions uglier, and we have a million of those).

The one maybe suspicious looking change is the removal of the `eval` on 1895 and 1899 in `acl-lib.pl`. There's already a hard error `$use_md5 || &error("No Perl MD5 hashing module found!");` before that, so the `eval` can never do anything useful, and a stringy eval is a `perlcritic` severity 5 error, and if somehow `$use_md5` was something that can't instantiate new, it's still `die`s on usage  in a following line, eval can't prevent an error in this case, and we don't want to prevent the fatal error; there's no way to ignore the error and do something good.

So, the previous code was kind of a bug-like. At least making the error harder to understand should execution ever somehow reach that location without something good in `$use_md5`.

The other eval changes are more verbose, but they enable compile-time syntax checking while the string form doesn't, it's protected against code injection (if a user is able to influence the string being evaled), also faster (string eval is a fresh compilation each time it's hit, block form compiles once), and static analysis can see the dependency.

The most common change is adding `## no critic` for various `require` and other idioms that come from Webmin's unusual module usage. Changing those requires larger code changes than I'd like to make in a simple perlcritic pass.

All callers of altered functions (e.g. the `return undef;` -> `return;` change), except one, have been checked to be sure they don't expect the old semantics; there are no array or hash context users that would be surprised by the change.

The one function (`generate_random_id`) that gets called in a list context has been given `## no critic`. That can be fixed in a future PR by either explicitly forcing to a scalar or by putting the ID in a variable first.

Also a minor change in `twofactor.pl` to use `UNIVERSAL->can` (https://perldoc.perl.org/UNIVERSAL) to satisfy the strict refs requirement while doing the right thing. This also provides a better error than what would happen previously in cases where `$provider` is non-existent.

All unit tests pass, and I've poked around in the module with no evidence of trouble. I believe all changes are safe.

### DIFF
--- a/acl/acl-lib.pl
+++ b/acl/acl-lib.pl
@@ -11,7 +11,7 @@ Library for editing webmin users, passwords and access rights.
 
 =cut
 
-BEGIN { push(@INC, ".."); };
+BEGIN { push(@INC, ".."); };    ## no critic
 use strict;
 use warnings;
 no warnings 'redefine';
@@ -352,7 +352,8 @@ each of which is a hash reference in the same format as their module.info files.
 sub list_module_infos
 {
 my @mods = grep { &check_os_support($_) } &get_all_module_infos();
-return sort { $a->{'desc'} cmp $b->{'desc'} } @mods;
+my @sorted = sort { $a->{'desc'} cmp $b->{'desc'} } @mods;
+return @sorted;
 }
 
 =head2 create_user(&details, [clone])
@@ -1331,12 +1332,12 @@ my ($miniserv) = @_;
 my $sfile = $miniserv->{'sessiondb'} ? $miniserv->{'sessiondb'} :
 	    $miniserv->{'pidfile'} =~ /^(.*)\/[^\/]+$/ ? "$1/sessiondb"
 						     : return;
-eval "use SDBM_File";
+eval { require SDBM_File; SDBM_File->import; 1 };
 dbmopen(%sessiondb, $sfile, 0700);
 eval { $sessiondb{'1111111111'} = 'foo bar' };
 if ($@) {
 	dbmclose(%sessiondb);
-	eval "use NDBM_File";
+	eval { require NDBM_File; NDBM_File->import; 1 };
 	dbmopen(%sessiondb, $sfile, 0700);
 	}
 else {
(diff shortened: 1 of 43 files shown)


acl/acl-lib.pl
```diff
@@ -1331,12 +1332,12 @@ sub open_session_db
 my $sfile = $miniserv->{'sessiondb'} ? $miniserv->{'sessiondb'} :
 	    $miniserv->{'pidfile'} =~ /^(.*)\/[^\/]+$/ ? "$1/sessiondb"
 						     : return;
-eval "use SDBM_File";
+eval { require SDBM_File; SDBM_File->import; 1 };
```
jcameron, line 1335: Won't this fail the script at compile time if SDBM_File is not installed?

acl/acl-lib.pl
```diff
@@ -1331,12 +1332,12 @@ sub open_session_db
...
 dbmopen(%sessiondb, $sfile, 0700);
 eval { $sessiondb{'1111111111'} = 'foo bar' };
 if ($@) {
 	dbmclose(%sessiondb);
-	eval "use NDBM_File";
+	eval { require NDBM_File; NDBM_File->import; 1 };
```
jcameron, line 1340: What does the `->import` do here? Also, why the `;1` at the end?

swelljoe: `use` automatically does `import`, so this makes it behave like the previous code.

acl/acl-lib.pl
swelljoe: No, it behaves the same (well, for our purposes, though it is more efficient), but it makes perlcritic happy and avoids stringy eval.

```
joe@cat:~/src/snippets$ cat eval-require.pl
#!/usr/bin/perl
#

eval { require NonExist; NonExist->import; 1};

print "We made it!\n";

joe@cat:~/src/snippets$ perl eval-require.pl
We made it!
```

jcameron: ok cool!

acl/acl-lib.pl
jcameron: ah ok thanks!

swelljoe: Oh, and the 1 just makes `eval` always return success.

swelljoe: I mean, if it reaches the end. This allows:

```
eval {...} or do {...}
```

For fallback behavior. Try::Tiny implements exceptions this way.

PR #2724, opened 2026-05-20
Add ip, paths, string tests for web-lib-funcs

swelljoe:
First round of unit tests for web-lib-funcs.pl.

There are some bug-likes surfaced by this round of tests:

1. check_ip6address rejects valid CIDR with trailing :: (web-lib-funcs.pl:694). ::1/64 passes, but 2001:db8::/32 is rejected. Cause: split(/:/, ...) only trims trailing empties when the final element is empty; the netmask suffix prevents that, so the empty-block count then exceeds the limit.
2. check_ip6address rejects IPv4-mapped IPv6 (web-lib-funcs.pl:694). ::ffff:10.0.0.1 is rejected because the per-block regex is hex-only. Knock-on: is_non_public_ipaddress has an explicit ::ffff:N.N.N.N branch (:751) that is unreachable — it always falls through and returns 0, so private addresses in IPv4-mapped form are wrongly reported as public.
3. trunc always pops at least one char (web-lib-funcs.pl:622). For trunc("hello world foo", 11) returns "hello worl" even though substr(0,11) cleanly ends on a word boundary. The pop loop's predicate should be while $c IS non-ws (keep chopping the partial word until we hit whitespace), not the current while $c IS ws.
4. split_quoted_string drops everything when input has only leading whitespace (web-lib-funcs.pl:12470). None of the three alternative regexes tolerate a \s* prefix, so "   spaces   between   " returns ().
5. html_escape default-mode double-escapes (web-lib-funcs.pl:297). The smarter behaviour is opt-in via the $nodblamp flag, so any caller that forgets returns &amp;amp; for already-escaped input. Sensible to flip the default eventually but it's a behaviour change. (Saved for later. Not altered in this PR.)
6. quote_path Windows branch doesn't quote-escape (web-lib-funcs.pl:12275). If a Windows path ever contains a ", the resulting string is broken; quotemeta doesn't apply on the Windows branch. Low risk, but worth a note. (Saved for later. Not altered in this PR.)

I'll include those fixes in this PR, absent objections.

### DIFF
--- a/miniserv.pl
+++ b/miniserv.pl
@@ -7027,26 +7027,49 @@ return $_[0] =~ /^(\d+)\.(\d+)\.(\d+)\.(\d+)$/ &&
 }
 
 # Check if some IPv6 address is properly formatted, and returns 1 if so.
+# Kept in sync with web-lib-funcs.pl's copy.
 sub check_ip6address
 {
-  my @blocks = split(/:/, $_[0]);
-  return 0 if (@blocks == 0 || @blocks > 8);
-  my $ib = $#blocks;
-  my $where = index($blocks[$ib],"/");
+  my $addr = $_[0];
   my $m = 0;
-  if ($where != -1) {
-    my $b = substr($blocks[$ib],0,$where);
-    $m = substr($blocks[$ib],$where+1,length($blocks[$ib])-($where+1));
-    $blocks[$ib]=$b;
-  }
-  return 0 if ($m <0 || $m >128); 
-  my $b;
+
+  # Strip an optional /N netmask before splitting. Doing this on the
+  # raw string (rather than from the last split element) keeps split()'s
+  # trailing-empty accounting intact for inputs like "2001:db8::/32",
+  # where the netmask would otherwise hide the trailing "::" shorthand.
+  if ($addr =~ s{/(\d+)\z}{}) {
+    $m = $1;
+    }
+  return 0 if ($m < 0 || $m > 128);
+
+  # Special case for unspecified address (analogous to 0.0.0.0 in IPv4),
+  # both bare and with a netmask.
+  return 1 if ($addr eq "::");
+
+  my @blocks = split(/:/, $addr);
+  return 0 if (@blocks == 0);
+
+  # Accept the IPv4-in-IPv6 forms (RFC 4291 §2.5.5: "::ffff:N.N.N.N"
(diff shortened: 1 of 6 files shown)


swelljoe:
Fixes added for `check_ip6address` (also fixed roughly identical function in `miniserv.pl`), `trunc`, and `split_quoted_string`.

Skipping 5 and 6.

5. has security implications and would need a full audit of usages. It currently seems to be safe with only a cosmetic bug (`&amp` can be rendered in output in cases where double escaping happens)...naive fixes for that without touching every caller could introduce security bugs.
6. Is similarly complicated to fix. There currently may be security implications on Windows, but we don't really try to run on Windows, anymore, so the risk is small or nil.

jcameron:
Possible future improvement - for any functions that are duplicated between miniserv.pl and web-lib-funcs.pl , extract them out into a common file.

iliaross:
Though insignificant, but the new IPv4-in-IPv6 validation accepts incomplete addresses like `1:2:3:4:192.168.0.1`. A valid uncompressed form would be `1:2:3:4:5:6:192.168.0.1`.

A simple fix would be to reject IPv4-tailed addresses when there is no `::` and the effective group count is less than 8. This applies to both `web-lib-funcs.pl` and `miniserv.pl`.

----

And, in fact, it was caught during our code review (you should have it on the email):

### Code Review

#### Findings

`[attention] miniserv.pl​:7051` - The new IPv4-in-IPv6 handling accepts invalid uncompressed addresses with too few groups, such as '1:2:3:4:192.168.0.1', because the IPv4 tail is counted as two groups but the code only enforces the upper bound of 8 groups and does not require '::' for omitted groups.

**Suggested fix:** After computing the effective group count, require either a compressed empty block or exactly 8 groups; for IPv4-tailed forms, reject when there is no '::' and the adjusted count is less than 8.

`[attention] web-lib-funcs.pl​:720` - The new IPv4-in-IPv6 handling accepts invalid uncompressed addresses with too few groups, such as '1:2:3:4:192.168.0.1', because the IPv4 tail is counted as two groups but the code only enforces the upper bound of 8 groups and does not require '::' for omitted groups.

**Suggested fix:** After computing the effective group count, require either a compressed empty block or exactly 8 groups; for IPv4-tailed forms, reject when there is no '::' and the adjusted count is less than 8.

PR #2729, opened 2026-05-21
Add web-lib-funcs data transform tests

swelljoe:
New test files
  - t/web-lib-funcs-encoding.t — 4 subtests covering encode/decode_base64 (RFC 4648 vectors + round-trip), encode/decode_base32 (RFC vectors + padding handling), serialise_variable/unserialise_variable (round-trip + wire-format pinning + Data::Dumper mode), convert_to_json/convert_from_json (round-trip, pretty, relaxed-parser, undef→{} quirk).
  - t/web-lib-funcs-numeric.t — 5 subtests covering is_int, is_float, float (incl. asymmetric + handling), and compare_version_numbers in both numeric and operator forms (Debian-ish quirks: separator equivalence, numeric-not-lexical, ubuntu strip, rcN < release, segment-count asymmetry).

  Bugs fixed in same PR
  - web-lib-funcs.pl:6245 — decode_base32 now strips trailing = padding before the tr, so RFC-conformant padded base32 strings decode correctly. (No effect on existing data. May also fix a bug that would prevent a pasted padded secret from working in 2FA.)
  - web-lib-funcs.pl:13241 — protect against undef values in compare (to make it warnings-clean).

  Quirky behaviours pinned but intentionally not fixed
  - convert_to_json(undef) returns "{}" (the ||= {} default).
  - encode_base32 omits = padding (RFC deviation, decoder is now tolerant either way).
  - float('+5') accepts leading + while is_int/is_float reject it (asymmetric).
  - compare_version_numbers('1.0','1.0.0') = -1 (shorter < longer when prefix matches).


iliaross:
It looks good to me! Thanks!

web-lib-funcs.pl
```diff
@@ -6242,6 +6242,8 @@ sub decode_base32
 {
 $_ = shift;
```
jcameron, line 6243: Ideallt we should replace `$_` here with a proper variable for the function param, but that can be fixed later.

PR #2730, opened 2026-05-22
xterm unit tests and minor refactor for testing

swelljoe:
Several minor refactors in this one to make functions testable, but I've tested and I think it's solid. Since this module has all sorts of security edges, we really want it to be fully testable.

  New tests (xterm/t/):
  - perlcritic.t — severity-5 lint gate (mirrors nftables/t/perlcritic.t)
  - run-tests.t — 7 subtests / 50 assertions covering config_pre_load, verify_websocket_key, parse_resize_message, resolve_shell_user, acl_security_save, shipped ACL templates,
   and require-time side-effect freedom

  Refactors to make security code testable:
  - xterm-lib.pl — added verify_websocket_key(), parse_resize_message(), resolve_shell_user() as pure functions. Behavior of resolve_shell_user is verbatim from the previous inline version.
  - shellserver.pl — wrapped the main body in unless (caller) { ... } (the require-and-stub pattern documented in t/README.md), so tests can load it without binding sockets.
  The inline WebSocket-key check and resize regex now call the extracted helpers.
  - index.cgi — replaced the inline user-resolution block with resolve_shell_user(\%access, $remote_user, \%in, \%config).

  Bug fixes (small, isolated):
  1. index.cgi typo: %termjs_opts (undeclared, autovivified) is now %term_opts (the actually-declared lexical). Would have been a compile error under strict.
  2. shellserver.pl was logging the base64-encoded session_id to $module_var_directory/websocket-connection-$port.out on every auth-failure handshake. Replaced with a value-free log line — both the client key and the session id are session-bearing secrets.
  3. xterm-lib.pl and index.cgi no longer raise uninit warnings on unset HTTP_X_REQUESTED_WITH / unset $config{size} (defensive || '').
  4. Perlcritic violations across all four files cleaned up (added use strict; use warnings; no warnings 'uninitialized';, declared package globals via our, marked the require './foo' and stringy eval "use ..." lines converted to block form).

### DIFF
--- a/xterm/acl_security.pl
+++ b/xterm/acl_security.pl
@@ -1,5 +1,9 @@
+use strict;
+use warnings;
+no warnings 'uninitialized';
 
-require 'xterm-lib.pl';
+require 'xterm-lib.pl';    ## no critic
+our (%in, %text);
 
 # acl_security_form(&options)
 # Output HTML for editing security options for the xterm module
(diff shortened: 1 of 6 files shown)


iliaross:
LGTM!

PR #2732, opened 2026-05-22
Add some ui-lib tests and fix quote escapes

swelljoe:
New file: t/ui-lib-escaping.t — 21 tests covering attribute-breakout resistance in 13 input-handling functions (the safe-contract regression set) plus the 3 patched XSS
  sites.

  Patches to ui-lib.pl:

  - ui_help (:58-64) — wrap title in html_escape after html_strip. Output shape unchanged; attribute value now escapes ", &, etc.
  - ui_img (:82-90) — pass src, alt, title through quote_escape. Output shape unchanged; the empty-title branch now uses $title ne '' instead of truthiness so quote_escape's
  empty-string handling doesn't drift.
  - js_redirect (:2916-2938) — quote_escape the URL, then neutralise `</script>` and `<!--` (the two HTML parser state transitions that can escape the script element); pin $timeout
   to int(); pin $window to `^[\w.]+$`.
 - ui_select - Line 1110: option label now `&html_escape($o->[1] || $o->[0], 1)` (covers explicit labels and the value-as-label fallback). Line 1116: add-if-missing label now `&html_escape($s, 1)`, keeping the literal `&nbsp;` for the empty case.

  The HTML shape stays identical for benign inputs.

Also, updates two callers of `ui_select` that never could have done anything good (HTML is invalid in a select option, but these two sites could send `<br>`, from `nice_destination_file` or `nice_remotes`). All callers were checked exhaustively, including in Virtualmin. I didn't check third-party modules, but it's such a rare thing (2 in all of Webmin's hundreds of callers) I'll be shocked if we see it, and the change is harmless. HTML in these places is not valid and would either be rendered as plain text or not at all, so this is just a security in depth change, not a breaking change.

### DIFF
--- a/t/ui-lib-escaping.t
+++ b/t/ui-lib-escaping.t
@@ -1,0 +1,135 @@
+#!/usr/bin/perl
+# Tests for ui-lib.pl attribute escaping and XSS resistance.
+#
+# These cover the *default* (non-theme) code path. Each function checks for
+# a theme override first and returns early if one is defined; when ui-lib.pl
+# is loaded as a library, no theme is present, so the default path runs.
+#
+# The core invariant: caller-supplied data interpolated into HTML must not
+# escape its attribute context. We check this by stripping all quoted
+# attribute values from the output and asserting that no event-handler
+# attribute (on*=) survives. If the input broke out of an attribute, the
+# handler appears outside any quoted region and the assertion fails.
+
+use strict;
+use warnings;
+use Test::More;
+use File::Basename qw(dirname);
+use File::Spec;
+
+my $root = File::Spec->rel2abs(File::Spec->catfile(dirname(__FILE__), '..'));
+require File::Spec->catfile($root, 'web-lib-funcs.pl');
+require File::Spec->catfile($root, 'ui-lib.pl');
+
+# Strip all "..." and '...' quoted regions from $html so that what remains
+# is tag scaffolding only. Any attribute introduced by an attribute-quote
+# breakout will appear in the stripped output.
+sub strip_attr_values {
+	my ($html) = @_;
+	$html =~ s/"[^"]*"//g;
+	$html =~ s/'[^']*'//g;
+	return $html;
+}
+
+sub assert_no_handler_injection {
+	my ($html, $label) = @_;
+	my $bare = strip_attr_values($html);
+	unlike($bare, qr/\bon[a-z]+\s*=/i,
+		"$label: no event-handler attribute leaks out of attribute value");
+}
+
+# Attribute-breakout payloads. The first is for double-quoted attrs, the
+# second for single-quoted; we use whichever matches what the function
+# emits (or both, when uncertain).
+my $xss_dq = q{x"onmouseover="alert(1)};
+my $xss_sq = q{x' onmouseover='alert(1)};
+
+# ---- safe-contract regression tests ---------------------------------------
+# These functions already escape correctly today. The tests lock that down.
+
+assert_no_handler_injection(main::ui_textbox('field', $xss_dq, 20),
+	'ui_textbox value');
+assert_no_handler_injection(main::ui_textbox($xss_dq, 'safe', 20),
+	'ui_textbox name');
+
+# ui_textarea must escape < so a value cannot inject </textarea>.
+{
+	my $payload = q{</textarea><script>alert(1)</script>};
+	my $html = main::ui_textarea('field', $payload, 5, 40);
+	unlike($html, qr{</textarea><script}i,
+		'ui_textarea value cannot close the textarea');
+	like($html, qr{&lt;/textarea}, 'ui_textarea value html-escapes <');
+}
+
+assert_no_handler_injection(main::ui_hidden($xss_dq, $xss_dq),
+	'ui_hidden name+value');
+assert_no_handler_injection(
+	main::ui_select('field', '', [[$xss_dq, 'label']]),
+	'ui_select option value');
+assert_no_handler_injection(main::ui_select($xss_dq, '', [['v','label']]),
+	'ui_select name');
+assert_no_handler_injection(main::ui_checkbox($xss_dq, 'v', 'label', 0),
+	'ui_checkbox name');
+assert_no_handler_injection(main::ui_oneradio($xss_dq, 'v', 'label', 0),
+	'ui_oneradio name');
+assert_no_handler_injection(main::ui_password('field', $xss_dq, 20),
+	'ui_password value');
+assert_no_handler_injection(main::ui_submit($xss_dq),
+	'ui_submit label');
+assert_no_handler_injection(main::ui_button($xss_dq),
+	'ui_button label');
+
+# ui_tag_start --- attribute values go through quote_escape.
+assert_no_handler_injection(
+	main::ui_tag_start('div', { 'data-foo' => $xss_dq }),
+	'ui_tag_start data attribute');
+
+# ---- the three XSS sites we are about to fix ------------------------------
+
+# ui_help: title goes into aria-label="...". Single-quote breakout doesn't
+# apply (attr is double-quoted), but double-quote breakout does.
+assert_no_handler_injection(main::ui_help($xss_dq),
+	'ui_help title');
+
+# ui_img: src/alt/title all in single-quoted attrs today.
+assert_no_handler_injection(main::ui_img('img.png', $xss_sq),
+	'ui_img alt');
+assert_no_handler_injection(main::ui_img('img.png', 'safe', $xss_sq),
+	'ui_img title');
+assert_no_handler_injection(
+	main::ui_img(q{img.png' onerror='alert(1)}, 'safe'),
+	'ui_img src');
+
+# js_redirect: url must not be able to close the script tag, and the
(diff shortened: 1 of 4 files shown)


ui-lib.pl
```diff
@@ -83,7 +83,10 @@ sub ui_img
 {
 return &theme_ui_img(@_) if (defined(&theme_ui_img));
 my ($src, $alt, $title, $class, $tags) = @_;
-return ("<img src='".$src."' class='ui_img".($class ? " ".$class : "")."' alt='$alt' ".($title ? "title='$title'" : "").($tags ? " ".$tags : "").">");
+my $esrc   = &quote_escape($src);
+my $ealt   = &quote_escape($alt);
+my $etitle = &quote_escape($title);
```
iliaross, line 88: Why do we escape both quotes here instead of just the one `'` we need, e.g.:

```perl
my $esrc   = &quote_escape($src, "'");
my $ealt   = &quote_escape($alt, "'");
my $etitle = &quote_escape($title, "'");
```

jcameron: Surely it's safer to always escape everything?

iliaross: Surely not. Overthinking is just as bad as underthinking.

swelljoe: Why would you want to be more specific about a security guardrail? That doesn't make sense. What good could that possibly do?

swelljoe: Anyway, security in depth is the goal, not "what can prevent the one specific problem that immediately springs to mind?". It should be "how would a very clever attacker work around the mitigation I just implemented?" The more general guardrails are, the more likely they are to catch the imaginative attacker. We _just_ had an issue in help...I don't know details, but leaving filenames and paths that users might be able to manipulate able to inject code is how you get exploits.

If you want to do something less, the onus is on you to prove doing it the right way breaks something (and prove we can't fix whatever it breaks, though I don't think it breaks anything).

iliaross: You don't see how double quotes could be used in a title, for example?

The rule is to escape only what’s necessary, not everything and “just in case”. If those vars like `$etitle` are inside an attribute with single-quoted content, the only thing to escape is a single quote, right?

swelljoe: Is there any such case anywhere in Webmin? (And, no, I don't think we should allow it. This is not a case where creativity in titles is a useful feature.)

swelljoe: Oh, hell, it doesn't even cause a problem. Escaping double-quotes in titles isn't harmful, even if there were such a case, and there isn't any such caller in all of Webmin.

You're trying to protect something that doesn't exist and wouldn't even be effected if it did.

There's no reason to be specific here.

iliaross: > Is there any such case anywhere in Webmin?

Trust me, we've run into issues multiple times with overly escaped data. And, I'm also fairly sure there were a few cases where we used double quotes in the title attributes, for example:

```
Visit "Limits and Validation ⇾ Validate Virtual Servers" page to validate this domain
```

> This is not a case where creativity in titles is a useful feature.

I don’t think this is about allowing “creativity” in titles. It’s about using the correct escaping for the each context.

For a single-quoted HTML attribute, a literal double quote does not terminate the attribute. Escaping it **is not required** for that context. 

Using the wrong escaping in the wrong context does not add security. It only makes the code look safer than it is, and makes it confusing to the reader.

swelljoe: You're making up a problem. There is no user of `ui_img` that does that, and, as I mentioned, it is not user-visible. HTML entities are decoded by the browser when rendering or displaying a tool tip for an image.

I stg you'll argue about anything.

Look!

```
$ cat quote-escape-demo.html
<!DOCTYPE html>
<html>
<head><title>img attribute entity decoding</title></head>
<body>
<h1>HTML entities in <code>img</code> attributes are decoded on render</h1>

<p>Hover each image to see the tooltip. Inspect the DOM to see the
decoded values. Source has the raw markup.</p>

<h2>1. Escaped title with both quote types</h2>
<p>Raw source: <code>title='He said &amp;quot;hi&amp;quot; &amp; she said &amp;#39;bye&amp;#39;'</code></p>
<img src="data:image/gif;base64,R0lGODlhAQABAAAAACw="
     width="32" height="32" style="border:1px solid #888"
     alt="alt &quot;text&quot; &#39;here&#39;"
     title='He said &quot;hi&quot; &amp; she said &#39;bye&#39;'>

<h2>2. Single-quoted attribute containing literal double quote</h2>
<p>Raw source: <code>title='Say "hi"'</code> — also valid; <code>"</code> is fine inside <code>'...'</code>.</p>
<img src="data:image/gif;base64,R0lGODlhAQABAAAAACw="
     width="32" height="32" style="border:1px solid #888"
     alt='Say "hi"'
     title='Say "hi"'>

<h2>3. Verify decoded values via JS</h2>
<pre id="out"></pre>
<script>
const imgs = document.querySelectorAll('img');
const out  = document.getElementById('out');
imgs.forEach((img, i) => {
  out.textContent +=
    `img[${i}] title = ${JSON.stringify(img.title)}\n` +
    `img[${i}] alt   = ${JSON.stringify(img.alt)}\n`;
});
</script>

<p>Expected: image 1's <code>title</code> property is
<code>He said "hi" &amp; she said 'bye'</code> — entities decoded,
even though the markup contained <code>&amp;quot;</code> and
<code>&amp;#39;</code>.</p>
</body>
</html>
```

image

iliaross: > HTML entities are decoded by the browser when rendering or displaying a tool tip for an image.

I know, thanks, Joe! Though, is this simple HTML output all we have to deal with?

> You're trying to protect something that doesn't exist and wouldn't even be effected if it did.

All I'm suggesting is to do things right, and in a simple way possible.

And also, I’m fairly sure that if you use JavaScript to query an element’s `title` attribute, it will return something like:

```
Visit &quote;Limits and Validation ⇾ Validate Virtual Servers&quote; page to validate this domain
```

This is messy and adds unnecessary clutter. Yet, if you’d rather not listen, you can submit it as is.

swelljoe: > All I'm suggesting is to do things right, and in a simple way possible.

That's exactly what this does.

I included JavaScript in the example showing this is correct. You're inventing a problem.

I am submitting it as-is, because it is correct as-is.

iliaross: > I am submitting it as-is, because it is correct as-is.

Can you explain why do you need to escape double quotes in a single-quoted HTML attribute?

I find this to be redundant output bloat! Moreover, it only expands every `"` from 1 byte to 6 bytes in the generated HTML making the page unnecessarily heavier. I really don’t get why you’re arguing with me about it.

And, by the way, using inline HTML in the UI library functions and elsewhere in the code should be considered obsolete. Nowadays, we should use `ui_tag` to do things safely and without that kind of overhead.

swelljoe: I've already said, security in depth.

Also, this change doesn't effect _any_ code. It does not make HTML larger because there is no caller that has a title or alt tag, much less one with double-quotes.

You are fighting like mad about a nonexistent problem.

There is no reason to limit this quoting to only single quotes.

> And, by the way, using inline HTML in the UI library functions and elsewhere in the code should be considered obsolete. Nowadays, we should use ui_tag to do things safely and without that kind of overhead.

I am not doing major refactors at this time. This PR adds tests and closes XSS risks with minor no-risk tweaks. Refactoring comes after tests.

iliaross: > I am not doing major refactors at this time.

Alright, agreed, as I also think we should focus only on closing XSS risks now. Thanks.

ui-lib.pl
```diff
@@ -2919,15 +2922,21 @@ sub js_redirect
...
 if ($url =~ /^\//) {
 	# If the URL is like /foo , add webprefix
 	$url = &get_webprefix().$url;
 	}
+$url = &quote_escape($url);
+# Prevent the URL from terminating or destabilising the enclosing
+# script element. </script> would close it outright; <!-- would push
+# the parser into script-data-escaped state.
+$url =~ s|</script|<\\/script|gi;
+$url =~ s|<!--|<\\!--|g;
```
iliaross: A better general fix would probably be:

```
$url =~ s/</\\u003C/g;
```

swelljoe: Yeah, that's fine, too. I've changed it.

t/ui-lib-escaping.t
```diff
@@ -0,0 +1,159 @@
...
+	my ($html, $label) = @_;
+	my $bare = strip_attr_values($html);
+	unlike($bare, qr/\bon[a-z]+\s*=/i,
+		"$label: no event-handler attribute leaks out of attribute value");
+}
+
+# Attribute-breakout payloads. The first is for double-quoted attrs, the
+# second for single-quoted; we use whichever matches what the function
+# emits (or both, when uncertain).
+my $xss_dq = q{x"onmouseover="alert(1)};
```
iliaross, line 44: Why no space here after `x"`?

swelljoe: No reason. It works the same either way. I see there is an inconsistency with the line after it, but that also doesn't matter. They perform identically.

Neither is wrong, both do what's intended.

PR #2737, opened 2026-05-23
Add bind8 tests, perlcritic, minor fixes

swelljoe:
This touches a lot of files, but is mostly non-invasive.

There are a couple of bugs surfaced but not fixed, plus a couple of unchecked values that have been fixed.

  New tests in bind8/t/:
  - perlcritic.t (129 files, severity 5) — same template as nftables/t/perlcritic.t; explicitly skips config.info.pl and module.info.pl because they're Webmin data files in Polish-translation-shaped names.
  - run-tests.t (128 assertions) covering: IPv4/IPv6 reverse helpers and round-trips, expand_ip6/expandall_ip6, email/dotted SOA RNAME round-trip, valdnsname (incl. the underscore-by-type rule), valemail, check_net_ip, compute_serial for all three soa_style modes (incl. counter rollover), make_record/record_id/find_record_by_id/join_record_values, SPF and DMARC parse/format round-trips, extract_time_units, version_atleast, wrap_lines, convert_to_absolute, make_reverse_name (incl. partial reverse delegation), dnssec_size_range/list_dnssec_algorithms, can_edit_zone/can_edit_view with allow/deny/view ACL shapes, full read_config_file → directive_lines → read_config_file round-trip of a realistic named.conf, and read_zone_file parsing across $TTL, SOA, NS, A, CNAME, MX, and the TXT→DMARC reclassification path, plus is_raw_format_records.

  Perlcritic fixes (severity 5, 0 remaining):
  - 121 CGIs and .pl files: annotated require './bind8-lib.pl'; with ## no critic (Webmin's idiomatic convention, matching nftables/).
  - bind8-lib.pl: stringy eval "..." for the Net::DNS::SEC::Tools probe rewritten to block-eval require + ->import; two opendir(ZONEDIR, ...) rewritten to lexical handles; three no strict "subs" blocks (intentional, for conditionally loaded rollmgr_sendcmd constants) annotated with ## no critic (ProhibitNoStrict); ~30 return undef; → return; (preserved wantarray semantics in find_value).
  - records-lib.pl: 5 return undef; → return; (all scalar-only callers).
  - log_parser.pl, system_info.pl, stop.cgi, mass_update_form.cgi, cpan_modules.pl: added use strict; use warnings;, lexicalized loop iterators and unscoped globals (log_parser.pl had a local (...) = @_; that should always have been my).
  - edit_soa.cgi: my $ttl = $defttl->{'defttl'} if ($defttl); → ternary (the my…if idiom has long-standing undefined semantics).

  Bugs surfaced — not fixed in this PR:
  1. valemail accepts garbage prefix: valemail('bad email@example.com') returns true because the `(\S*)\@(\S*)` regex isn't anchored. Easy fix (anchor ^…$, require \S+), but touches CGI input validation across record-edit flows, so worth its own PR with care taken about backwards compatibility for edge cases like leading dots.
  2. The test suite also documents (via test design, not asserts) that the ACL deny-list convention uses ! a b c with ! as a separate token: a single !banned.com token is silently treated as allow-all because it never matches any zone name. Construction code in acl_security.pl already does this correctly, but if anyone hand-edits an ACL the failure mode is silent.

  Bugs fixed:

In `create_dnssec_key` `opendir` is checked for success. Now returns an error message on failure. All callers checked, and should be fine.

In `get_dnssec_key` the `opendir` on line 3846 now has an out of band warn to log an error. Previously silently failed. Behavior should otherwise be the same.

### DIFF
--- a/bind8/acl_security.pl
+++ b/bind8/acl_security.pl
@@ -3,7 +3,7 @@ use warnings;
 no warnings 'redefine';
 no warnings 'uninitialized';
 
-require 'bind8-lib.pl';
+require 'bind8-lib.pl';    ## no critic
 # Globals from bind8-lib.pl
 our (%config, %text, %in);
 
--- a/bind8/backup_config.pl
+++ b/bind8/backup_config.pl
@@ -45,21 +45,21 @@ return map { &make_chroot($_) } &unique(@rv);
 # Called before the files are actually read
 sub pre_backup
 {
-return undef;
+return;
 }
 
 # post_backup(&files)
 # Called after the files are actually read
 sub post_backup
 {
-return undef;
+return;
 }
 
 # pre_restore(&files)
 # Called before the files are restored from a backup
 sub pre_restore
 {
-return undef;
+return;
 }
 
 # post_restore(&files)
@@ -71,7 +71,7 @@ my $pidfile = &get_pid_file();
 if (&check_pid_file(&make_chroot($pidfile, 1))) {
 	return &restart_bind();
 	}
-return undef;
+return;
 }
 
 1;
--- a/bind8/bind8-lib.pl
+++ b/bind8/bind8-lib.pl
@@ -3470,7 +3470,8 @@ $fn || return "Could not work out keys directory!";
 my $dom = $z->{'members'} ? $z->{'values'}->[0] : $z->{'name'};
 
 # Remove all keys for the same zone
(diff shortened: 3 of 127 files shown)


iliaross:
Thanks, Joe! The PR looks good and safe to me.

jcameron:
Thanks for these code cleanups!

PR #2741, opened 2026-05-23
Postfix tests, perlcritic, strict/warnings, bugfixes

swelljoe:
Another one that touches a lot of files, but is mostly low-risk.

  Added postfix/t/:
  - run-tests.t — 57 assertions exercising the pure/parsing logic in postfix-lib.pl: the main.cf parser (get_current_value incl. continuation lines and foo:bar sub-parameters), resolve_current_value's compatibility-level conditionals (all six operators), the master.cf parser + master_line serializer round-trip, get_maps_types_files/get_maps_files, file_map_type, is_table_comment/make_table_comment, in_props, get_ldap_key, make_map_ldap_dn, get_backend_config, and load-time globals. It seeds a fake module config so loading never shells out to postconf.
  - perlcritic.t — severity-5.
  
  All 127 tests pass; compile.t passes for all 80 postfix files.

  perlcritic / use strict

  Brought the lib and all 69 module-owned .cgi/.pl files to use strict; use warnings; at severity 5 (237 violations → 0), following the established convention: ## no critic on require '...' lines, our(...) for real globals, local→my for function-locals, lexical loop iterators, return undef→return, stringy eval "use M"→ block eval { require M; ... }, and 2-arg/bareword opens → 3-arg lexical opens.

  Bugs found

  Fixed (small, module-contained, behavior-preserving or clearly correct):
  - mailq_search.cgi — reflected XSS. The search term $in{'match'} was interpolated into <tt>…</tt> and passed to &text() (which does not HTML-escape). Now wrapped in &html_escape.
  - list_queue (postfix-lib.pl:1485) — $q->{'file'} = $file referenced an undefined $file; corrected to $path (the value never recorded; unused downstream).
  - supports_map_type — cleared @supports_map_type while the cache is @supports_map_type_cache (dead/wrong line); corrected.
  - reload_postfix — referenced an always-empty undeclared $out in its error return; dropped the dead $out ||.

  Surfaced, not changed (pre-existing, behavioral):
  - postfinger.cgi has a broken guard at line ~72: ("x…" != "x") compares two strings with the numeric !=, so it's always false and the "don't run in a group/world-writable dir" check never triggers. It also system("rm postfinger.*.d …") and writes postfinger.$$.{d,n} into the module CWD, and prints system-command output (hostname/uname/postconf/master.cf/ls) into HTML unescaped. None are web-user-controlled and the page is gated by  $access{'postfinger'} + readonly mode, but the logic is dubious. Fixing the != changes behavior, so I left it.
  - mailq_search.cgi strips ! from $in{'field'} on line 18, making the =~ /^!/ test on line 22 always false (the search_results3 message is dead). Cosmetic.
  - The symlinked edit_afile/ffile/rfile.cgi (owned by sendmail) interpolate $in{'file'} into HTML unescaped — an XSS, but it's another module → separate PR.

### DIFF
--- a/postfix/acl_security.pl
+++ b/postfix/acl_security.pl
@@ -12,7 +12,6 @@ our (@acl_pages, %in, %text);
 # Print the form for security options of postfix module
 sub acl_security_form
 {
-my $o;
 foreach my $o (@acl_pages) {
 	print &ui_table_row($text{'acl_'.$o} || $o,
 			    &ui_yesno_radio($o, $_[0]->{$o} ? 1 : 0));
@@ -26,7 +25,6 @@ print &ui_table_row($text{'acl_dir'},
 # Parse the form for security options for the postfix module
 sub acl_security_save
 {
-my $o;
 foreach my $o (@acl_pages) {
 	$_[0]->{$o} = $in{$o};
 	}
--- a/postfix/acl_security.pl
+++ b/postfix/acl_security.pl
@@ -1,5 +1,8 @@
 
-require 'postfix-lib.pl';
+require 'postfix-lib.pl';    ## no critic
+use strict;
+use warnings;
+our (@acl_pages, %in, %text);
 @acl_pages = ("resource", "address_rewriting", "aliases", "general",
 	      "canonical", "virtual", "transport", "relocated", "header",
 	      "body", "bcc", "dependent", "sni", "local_delivery", "smtpd",
@@ -9,8 +12,8 @@ require 'postfix-lib.pl';
 # Print the form for security options of postfix module
 sub acl_security_form
 {
-local $o;
-foreach $o (@acl_pages) {
+my $o;
+foreach my $o (@acl_pages) {
 	print &ui_table_row($text{'acl_'.$o} || $o,
 			    &ui_yesno_radio($o, $_[0]->{$o} ? 1 : 0));
 	}
@@ -23,8 +26,8 @@ print &ui_table_row($text{'acl_dir'},
 # Parse the form for security options for the postfix module
 sub acl_security_save
 {
-local $o;
-foreach $o (@acl_pages) {
+my $o;
+foreach my $o (@acl_pages) {
 	$_[0]->{$o} = $in{$o};
 	}
 $_[0]->{'dir'} = $in{'dir'};
--- a/postfix/address_rewriting.cgi
+++ b/postfix/address_rewriting.cgi
@@ -7,7 +7,10 @@
 #
 # << Here are all options seen in Postfix sample-rewrite.cf >>
 
-require './postfix-lib.pl';
+require './postfix-lib.pl';    ## no critic
+use strict;
+use warnings;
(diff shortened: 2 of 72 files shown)


jcameron:
Let me know when this is ready for review

swelljoe:
The code review comments from the github actions bot (GPT 5.5) are incorrect.

Re: the first of the "[attention]" comments:

The claim rests on a mistaken model of `open(FH, "-|", $string)`. Perl treats a single-scalar command the same way system/exec do: metacharacters → /bin/sh -c, otherwise split-and-exec. Since every converted command in postfinger.cgi carries 2>/dev/null (and most also carry backticks), they all take the shell path — byte-for-byte the same behavior as the original "... |" two-arg form. Redirections and backticked sub-commands are honored, as the test proved.

The reviewer's suggested fix (`open($mailq, '-|', '/bin/hostname'`) as separate list args) would actually be worse here: the list form bypasses the shell entirely, which would break the 2>/dev/null redirection and the backtick expansion that these commands depend on. So adopting it would introduce the very failure the comment claims to prevent.

No edit to make. The safe litmus test for these -| findings is the same as the $a one — run it. open(my $f,"-|","ls -l /x 2>&1"); print <$f> demonstrates the shell path in two lines.

And, the second:

Same misunderstanding — the command runs fine, no fix needed.

With the configured command being /usr/sbin/postconf (no shell metacharacters), the string is /usr/sbin/postconf -m. That hits Perl's metachar-free path: it splits on whitespace and execs /usr/sbin/postconf with argument -m directly. This is exactly my earlier test case 2 (ls -d /tmp → printed /tmp) — executable + space-separated arg, no metachars, runs correctly. The map types are not "always unavailable."

  Behavior is preserved versus the original in every case:
  - Normal value (/usr/sbin/postconf): original ran it via shell; mine execs directly — identical output.
  - Value containing metachars: the whole string has a metachar → mine routes to /bin/sh -c, matching the original.
  - Path containing a space (e.g. /opt/my dir/postconf): broken under both forms — the original shell would split on the space too. Pre-existing, not introduced here.

  One difference from the postfinger.cgi case worth noting: here the reviewer's suggested list-form fix — open(my $postconffh, '-|', $config{'postfix_config_command'}, '-m') — would actually work, because this command has no redirection or backticks to preserve. It's a defensible optional tweak (slightly more robust for an exotic space-in-path config), but it isn't required for correctness, and it would diverge from the original's behavior in the metachar-config edge case. The current code is correct as-is.

  So: both this finding and the postfinger.cgi one stem from the same incorrect premise that open(FH, "-|", $string) always treats the string as a lone executable name. It doesn't.

swelljoe:
Wait, hang on. Found a bug.

swelljoe:
OK, bugs fixed. I think we're good to go on this one.

jcameron:
Thanks for the code cleanups!

iliaross:
Jamie, after merging a PR, I suggest to click the "Delete branch" button above to keep things clean.

PR #2743, opened 2026-05-28
Add GRUB 2 boot loader module

iliaross:
Hello Jamie,

This PR introduces a new Webmin GRUB 2 module for managing boot entries, editing defaults, creating custom entries, handling themes, configuring password protection, managing BLS-aware kernel options, safely regenerating menus, installing the boot loader, manually editing configuration files, reporting status, managing ACLs, performing backups, handling logging, and running tests.

This PR is primarily to highlight the update and does not require a review. I plan to merge it tomorrow unless there are any questions.

For the best UI experience, please make sure you’re using the latest Webmin build along with the latest theme.

Looking forward to your feedback!

Screenshots:

Screenshot 2026-05-28 at 02 06 33
Screenshot 2026-05-28 at 02 06 44
Screenshot 2026-05-28 at 02 07 41
Screenshot 2026-05-28 at 02 11 35
Screenshot 2026-05-28 at 02 11 53
Screenshot 2026-05-28 at 02 12 04
Screenshot 2026-05-28 at 02 15 24
Screenshot 2026-05-28 at 02 15 46
Screenshot 2026-05-28 at 02 16 12
Screenshot 2026-05-28 at 02 16 23


gray-theme/unauthenticated/gray-theme.css
jcameron: Is this change part of the GRUB 2 module?

grub2/grub2-lib.pl
```diff
@@ -0,0 +1,3347 @@
...
+# grub2_acl_keys()
+# Returns the supported GRUB 2 ACL capabilities.
+sub grub2_acl_keys
+{
+return qw(view edit security apply runtime manual install backup);
+}
+
+# grub2_effective_acl([&raw-acl])
+# Returns normalized ACL settings for the current Webmin user.
+sub grub2_effective_acl
```
jcameron: What's this function for? It seems like an unusual pattern versus just calling get_module_acl like other modules do

grub2/acl_security.pl
```diff
@@ -0,0 +1,48 @@
...
+# acl_security_form(&options)
+# Outputs HTML for editing security options for the GRUB 2 module.
+sub acl_security_form
+{
+my ($o) = @_;
+print &ui_table_span(&ui_tag('b', &html_escape($text{'acl_section_view'})));
+# View is separated because it controls whether the module can be entered.
+foreach my $a (qw(view)) {
+	print &ui_table_row($text{'acl_'.$a},
+			    &ui_yesno_radio($a, &grub2_check_acl($a, $o)), 3);
```
jcameron: Why not just use $o->{$a} ?

grub2/acl_security.pl
```diff
@@ -0,0 +1,48 @@
...
+our (%in, %text);
+
+# acl_security_form(&options)
+# Outputs HTML for editing security options for the GRUB 2 module.
+sub acl_security_form
+{
+my ($o) = @_;
+print &ui_table_span(&ui_tag('b', &html_escape($text{'acl_section_view'})));
+# View is separated because it controls whether the module can be entered.
+foreach my $a (qw(view)) {
```
jcameron: Why a loop over one element?

gray-theme/unauthenticated/gray-theme.css
iliaross: Not really. I submitted it as a separate change, but it accidentally slipped in here. I've reverted it.

grub2/grub2-lib.pl
iliaross: It’s a small abstraction to centralize ACL normalization so _page_ checks, the ACL _editor_, and _tests_ (since we’ve all agreed to have tests) all use the same capabilities.

It also follows the newer module style, like `kea-dhcp`. And, thinking of this more, we might eventually want to turn this pattern into a shared Webmin ACL helper API so new modules don’t keep reinventing the same logic. I can check this after the merge.

grub2/acl_security.pl
iliaross: For the same reason, to keep the ACL editor aligned with runtime ACL checks and make the behavior easier to test and change.

grub2/acl_security.pl
iliaross: No practical reason, except to make the three ACL sections visually consistent and readable.

jcameron: I would recommend dropping this then, it actually makes the code less understandable

iliaross: Yes, already did.

grub2/grub2-lib.pl
jcameron: I still don't understand why this is needed. The ACL editor is already supplied the fields to edit as a parameter, so it doesn't need to do any normalization. In fact, doing so may break some cases like Webmin ACL groups. Please undo this, and just do it the way all the older Webmin modules have.

grub2/acl_security.pl
jcameron: Please don't, as I mentioned above the ACL passed into acl_security.pl must be treated differently to what's returned by get_module_acl

iliaross: Can you first explain why in more detail please?

iliaross: Alright, I’ve made it more in line with Webmin conventions.

Though, to be perfectly honest, I still think it’s somewhat a downgrade since we’ve lost the abstraction for repeated boilerplate and now have two places with the ACL key list...

PR #2744, opened 2026-05-29
Fix stale mailbox entries after deleted or moved

iliaross:
Hey Jamie!

This PR fixes the bug discussed in https://github.com/webmin/usermin/issues/134.

It refreshes stale Maildir data and sorts mailbox indexes when messages disappear, avoids rendering missing messages, and keeps IMAP sort indexes in sync with mailbox count changes.

### DIFF
--- a/mailboxes/boxes-lib.pl
+++ b/mailboxes/boxes-lib.pl
@@ -2207,6 +2207,32 @@ else {
 					# Flagged as deleted by IMAP .. skip
 					next;
 					}
+				# Skip entries that cannot be read as messages.
+				local $path = "$_[0]/$d/$f";
+				local @fst = stat($path);
+				if (!@fst) {
+					&error_stderr("Skipping Maildir file ".
+						      "$path : stat failed : ".
+						      "$!");
+					next;
+					}
+				if (!$fst[7]) {
+					&error_stderr("Skipping Maildir file ".
+						      "$path : file is zero ".
+						      "bytes");
+					next;
+					}
+				if (!-r _) {
+					my $m = sprintf("%04o",$fst[2] & 07777);
+					my $o = getpwuid($fst[4]) || $fst[4];
+					my $g = getgrgid($fst[5]) || $fst[5];
+					&error_stderr("Skipping Maildir file ".
+						      "$path : not readable by".
+						      " current user, owner=".
+						      "$o($fst[4]):$g($fst[5])".
+						      " mode=$m");
+					next;
+					}
 				push(@shorts, "$d/$f")
 				}
 			closedir(DIR);
@@ -2682,9 +2708,11 @@ my ($file, $headersonly) = @_;
 # Open and read the mail file
 &open_as_mail_user(MAIL, $file) || return undef;
 my $mail = &read_mail_fh(MAIL, 0, $headersonly);
-$mail->{'file'} = $file;
 close(MAIL);
+
 local @st = stat($file);
+return undef if (@st && !$st[7]);
+$mail->{'file'} = $file;
 $mail->{'size'} = $st[7];
 $mail->{'time'} = $st[9];
 $mail->{'ctime'} = $st[10];
--- a/mailboxes/boxes-lib.pl
+++ b/mailboxes/boxes-lib.pl
@@ -2706,12 +2706,12 @@ sub read_mail_file
 my ($file, $headersonly) = @_;
 
 # Open and read the mail file
+local @st = stat($file);
+return undef if (@st && !$st[7]);
 &open_as_mail_user(MAIL, $file) || return undef;
 my $mail = &read_mail_fh(MAIL, 0, $headersonly);
 close(MAIL);
 
-local @st = stat($file);
-return undef if (@st && !$st[7]);
 $mail->{'file'} = $file;
 $mail->{'size'} = $st[7];
 $mail->{'time'} = $st[9];
--- a/mailboxes/boxes-lib.pl
+++ b/mailboxes/boxes-lib.pl
@@ -2087,10 +2087,20 @@ foreach $f (@files) {
 		$i++;
 		next;
 		}
+	local $idx = $i++;
 	local $mail = &read_mail_file($f, $_[3]);
-	$mail->{'idx'} = $i++;
-	$mail->{'id'} = $f;	# ID is relative path, like cur/4535534
-	$mail->{'id'} = substr($mail->{'id'}, length($_[0])+1);
+	if (!$mail && !$_[4]) {
+		# The cached Maildir file list can be stale if another client
+		# deleted or moved a message. Re-read it once before returning
+		# blank entries to the caller.
+		&flush_maildir_cachefile($_[0]);
+		return &list_maildir($_[0], $_[1], $_[2], $_[3], 1);
+		}
+	if ($mail) {
+		$mail->{'idx'} = $idx;
+		$mail->{'id'} = $f;	# ID is relative path, like cur/4535534
+		$mail->{'id'} = substr($mail->{'id'}, length($_[0])+1);
+		}
 	push(@rv, $mail);
 	}
 return @rv;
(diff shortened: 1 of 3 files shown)


mailboxes/boxes-lib.pl
```diff
@@ -2188,6 +2207,32 @@ sub get_maildir_files
 					# Flagged as deleted by IMAP .. skip
 					next;
 					}
+				# Skip entries that cannot be read as messages.
+				local $path = "$_[0]/$d/$f";
+				local @fst = stat($path);
+				if (!@fst) {
+					&error_stderr("Skipping Maildir file ".
```
jcameron, line 2214: Will this call to `error_stderr` cause the whole script to exit at this point?

iliaross: No, it just prints an error to `miniserv.error`. It doesn't exit the script.

jcameron: ok cool!

mailboxes/boxes-lib.pl
```diff
@@ -2167,7 +2186,7 @@ sub get_maildir_files
 	# Check the on-disk cache file
 	local $cachefile = &get_maildir_cachefile($_[0]);
 	local @cst = $cachefile ? stat($cachefile) : ( );
-	if ($cst[9] >= $newest) {
+	if ($cst[9] > $newest) {
```
jcameron, line 2189: Why this change?

mailboxes/boxes-lib.pl
```diff
@@ -2663,9 +2708,11 @@ sub read_mail_file
 # Open and read the mail file
 &open_as_mail_user(MAIL, $file) || return undef;
 my $mail = &read_mail_fh(MAIL, 0, $headersonly);
-$mail->{'file'} = $file;
 close(MAIL);
+
 local @st = stat($file);
+return undef if (@st && !$st[7]);
```
jcameron: Shouldn't this check be done before open_as_mail_user above?

mailboxes/boxes-lib.pl
```diff
@@ -2188,6 +2207,32 @@ sub get_maildir_files
 					# Flagged as deleted by IMAP .. skip
 					next;
 					}
+				# Skip entries that cannot be read as messages.
+				local $path = "$_[0]/$d/$f";
+				local @fst = stat($path);
```
jcameron, line 2212: This is going to make listing large mailboxes **much** slower if we have to call stat on every file. Are you sure this is a good idea?

mailboxes/boxes-lib.pl
```diff
@@ -2087,10 +2087,20 @@ sub list_maildir
...
 	local $mail = &read_mail_file($f, $_[3]);
-	$mail->{'idx'} = $i++;
-	$mail->{'id'} = $f;	# ID is relative path, like cur/4535534
-	$mail->{'id'} = substr($mail->{'id'}, length($_[0])+1);
+	if (!$mail && !$_[4]) {
+		# The cached Maildir file list can be stale if another client
+		# deleted or moved a message. Re-read it once before returning
+		# blank entries to the caller.
+		&flush_maildir_cachefile($_[0]);
+		return &list_maildir($_[0], $_[1], $_[2], $_[3], 1);
```
jcameron, line 2097: Couldn't this cause an infinite loop?

mailboxes/boxes-lib.pl
iliaross: If an email is deleted (or moved) and the cache file and Maildir directory land in the same time, the cache might still be stale.

mailboxes/boxes-lib.pl
iliaross: Yes, agreed. Fix now. If the file disappears or becomes unreadable after `stat`, `open_as_mail_user()` already returns `undef`, so the race is still handled...

mailboxes/boxes-lib.pl
iliaross: Yeah, that’s fine. This is the only spot where we can block unusable entries from getting into the mailbox list.

The performance hit happens only during cache rebuilds, and it’s minimal. In a 100k-file test, it added about 0.111s for the stat check.

mailboxes/boxes-lib.pl
iliaross: No, because the fifth param (i.e., `$_[4]`) being set to `1` prevents this block from being entered again.

mailboxes/boxes-lib.pl
iliaross: And, I just ran the same 100k-file benchmark on an **old low-grade** spinning HDD system, and the cost is much higher there. Plain scan and sort was about 0.175s, while scan and sort with stat checks was about 1.5s. That is roughly 1.34s extra per 100k files, or about 13s extra for 1M files during a Maildir filename cache rebuild.

It isn’t nothing, but still, correctness, as someone once said, is more important. 🙂

iliaross: All that said, for comparison, that old HDD case is roughly a 760% increase during cache rebuilds, versus about 320% on faster storage.

The relative percentage looks scary, but the absolute cost is still bounded to cache rebuilds only. 

And, for very large Maildirs on low-grade spinning disks this can be noticeable, but that kind of mailbox size is already a poor fit for that class of storage.

iliaross:
So, are we good with this one?

mailboxes/boxes-lib.pl
jcameron: Ah ok thanks!

mailboxes/boxes-lib.pl
jcameron: Ok but wouldn't this change potentially cause an invalid cache to be used if it was generated in the same second as the Maildir update? That could happen in the case when Webmin itself is writing to the Maildir.

mailboxes/boxes-lib.pl
jcameron: Ok .. so is the specific failure mode you're worried about the case when a new file has been created in the maildir but is zero size?

mailboxes/boxes-lib.pl
```diff
@@ -2188,6 +2207,32 @@ sub get_maildir_files
...
+						      "$!");
+					next;
+					}
+				if (!$fst[7]) {
+					&error_stderr("Skipping Maildir file ".
+						      "$path : file is zero ".
+						      "bytes");
+					next;
+					}
+				if (!-r _) {
```
jcameron, line 2225: This seems like a really rare case with regular mail delivery, unless there are situations where new mail files are written as root and then renamed?

mailboxes/boxes-lib.pl
iliaross: > Ok but wouldn't this change potentially cause an invalid cache to be used if it was generated in the same second as the Maildir update? That could happen in the case when Webmin itself is writing to the Maildir.

Yeah, that’s correct, and that’s why we need this change. 🙂 With the new `$cst[9] > $newest` check, same-second timestamps are not trusted, right??

mailboxes/boxes-lib.pl
iliaross: Any guard against Maildir entries that exist on disk but cannot become valid messages. zero-byte files or files owned by another user like `root`—there were plenty reports over years where users had ghost rows like None, 12/31/1969.

mailboxes/boxes-lib.pl
iliaross: Yes, it can happen when mail files are owned by another user, for example, after manual copies, third-party restores, or migration scripts and etc. A fairly ordinary scenario for many users.

Now, instead of making it look like something is broken on our side, we will print an error to stdout and ignore the message.

jcameron: Ok, seems like a rare case, but since you're using `-r _` the overhead is small.

jcameron:
Thanks for this patch and the explanations! I have also occasionally notice those "None" messages, so hopefully this fixes them.

iliaross:
Thanks!

> I have also occasionally notice those "None" messages, so hopefully this fixes them.

I’d be really surprised if it doesn’t.

Issue #2559, opened 2025-09-26
Raspberry OS Bookworm - Webmin - Network Configuration mismatch


pieter-groeneweg:
Installing Webmin on Raspberry OS Bookworm in a Pi4 (maybe Pi5 too) showed the networking -> network configurations settings not being respected on edit and save. 

It appears that since Bookworm, Raspberry OS is relying on dhcpcd and not ifup. 

My solution was to uninstall and purge dhcpcd5 then install ifupdown (keyboard and monitor attached)
add sudo if you need.  

```
apt purge dhpcd5
systemctl disable dhcpcd
systemctl stop dhcpcd
```

Then reconnect temporary by:

```
ip addr add 192.168.1.100/24 dev eth0
ip link set eth0 up
ip route add default via 192.168.1.1
echo "nameserver 8.8.8.8" | tee /etc/resolv.conf
```

use your own IP and network interface if not eth0. you have connection again to get to the repositories.

```
apt update
apt install ifupdown
```

Go to webmin networking -> network configuration in any browser, login and edit/save the network settings then reboot.

This requires the PI's to connect to a monitor and keyboard.. something I usually not do.
Best thing of course would be if Webmin would recognize what network management is used.
But meanwhile, does anyone know a solution that could be done ssh (headless)?

Tx

jcameron:
Does Bookworm use a network config file format similiar to any other distro, like Debian or CentOS? Or does it have it's own custom format?

pieter-groeneweg:
Bookworm is a Debian version (Debian 12)
But for raspberry it uses dhpcd (etc/dhcpd.conf) apparently and not ifupdown (etc/network/interfaces)

The networking->network configurations -> network interfaces (tab Activated at Boot) are saved by webmin in etc/network/interfaces if available. 
If not, the settings in webmin are not respected and after boot there is no connection at all anymore. Only to be corrected with monitor and keyboard after the above steps.
Since this is a raspberry issue most probably, you may leave this just as a comment / remark for other users that stumble upon this issue.

I did not have this issue on a regular small server also using same Debian 12 Bookworm.

jcameron:
Hmm, so do you mean that with Bookworm there is no way to configure a static IP address, and it only uses DHCP?

pieter-groeneweg:
No it uses dhpcd.conf.... and when I change configuration through Webmin panel, it is not repected an reboot. After reboot there is no connection at all till I connect a keyboard, mouse and monitor to configure some locally.

[...]

PR #2747, opened 2026-06-01
Fix to detect NetworkManager networking on Debian

iliaross:
Hey Jamie!

As discussed [here](https://github.com/webmin/webmin/issues/2559), this PR prefers Netplan when Debian uses a Netplan YAML config. Otherwise, it selects the existing NetworkManager backend for Debian systems with saved NM connection profiles.

### DIFF
--- a/net/net-lib.pl
+++ b/net/net-lib.pl
@@ -6,6 +6,9 @@ use WebminCore;
 &init_config();
 %access = &get_module_acl();
 $access{'ipnodes'} = $access{'hosts'};
+do "net-detect.pl";
+
+$auto_net_mode = &net_auto_backend($gconfig{'os_type'});
 
 if (-r "$module_root_directory/$gconfig{'os_type'}-$gconfig{'os_version'}-lib.pl") {
 	do "$gconfig{'os_type'}-$gconfig{'os_version'}-lib.pl";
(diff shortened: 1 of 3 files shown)


net/net-lib.pl
```diff
@@ -6,6 +6,9 @@
 &init_config();
 %access = &get_module_acl();
 $access{'ipnodes'} = $access{'hosts'};
+do "net-detect.pl";
```
jcameron, line 9: Is this a separate library for testability purposes?

jcameron:
Please test this on as many Linux distros and versions as possible, as there's a lot of odd mixes of configs out there!

net/net-lib.pl
iliaross: Yes, added in this PR.

iliaross:
> Please test this on as many Linux distros and versions as possible, as there's a lot of odd mixes of configs out there!

I'm not entirely sure what needs to be tested. We're already familiar with the default settings for Ubuntu, Debian, and EL. 

If an administrator on an EL system chooses to use `netplan` or `nm` as on Debian/Ubuntu, we’ll support it.

Please review the test logic in the new `net-detect.pl`. To my mind, it doesn't change much. It just adds the assumption that Debian and Ubuntu may also use `nm`.

If you have any questions let me know.

jcameron:
Excellent, thanks for the patch!

Issue #2734, opened 2026-05-22
Systemd improvements


dogarethebest:
When making changes to a service, the interface redirects back to the full service list instead of returning to the specific service being edited, which makes troubleshooting more difficult. Additionally, the service creation flow should expose all available configuration options such as `WorkingDirectory` so they do not need to be manually added afterward.

iliaross:
Hello,

Thanks for the heads-up. I've started working on this, and it will be done together with the issue report at https://github.com/webmin/webmin/issues/2733.

iliaross:
Implemented in this https://github.com/webmin/webmin/pull/2748 PR.

Anyone who wants to try it right now can run:

```
webmin patch https://github.com/webmin/webmin/pull/2748.patch
```

Issue #2733, opened 2026-05-22
[Feature Request] Support linger and systemd user services (systemctl --user)


rubenferreira97:
Webmin already supports system-wide systemd services in:
```text
System → Bootup and Shutdown
```
It would be great if this module could also support systemd user services (`systemctl --user`) for users with `linger` enabled.
Right now this already works manually from the shell, for example:
```bash
loginctl enable-linger userexample
```
Then the user can create services in:
```bash
~/.config/systemd/user/
```
and manage them with:
```bash
systemctl --user
```
Example:
```bash
systemctl --user daemon-reload
systemctl --user enable myapp.service
systemctl --user restart myapp.service
systemctl --user status myapp.service
journalctl --user -u myapp.service
```
### Webmin (root/admin)
Webmin could manage linger state:
```bash
loginctl enable-linger userexample
loginctl disable-linger userexample
```
and expose a new section inside Bootup and Shutdown:
```text
Bootup and Shutdown
├── System Services
└── User Services
```
### Virtualmin / user scope
If linger is enabled for a user, Virtualmin could expose a UI for managing services in:
```bash
~/.config/systemd/user/
```
including:
* Create/edit `.service`
* daemon-reload
* enable/disable
* start/stop/restart
* status
* logs (`journalctl --user`)

This would fit very naturally into the existing Bootup and Shutdown module and would make Virtualmin much nicer for hosting modern apps (Node.js, Python, Bun, workers, etc.) using isolated per-user services instead of global root-owned services.

jcameron:
Interesting, but what does "linger" mode do exactly?

rubenferreira97:
@jcameron In systemd, “linger” means the user’s systemd user manager is allowed to keep running even when the user has no active login session.

Without linger, `systemctl --user` services are tied to a login session, so they normally only work while the user is logged in.

With linger enabled:

```bash
loginctl enable-linger username
```

systemd can start that user’s `user@UID.service` at boot, and the user’s units in:

```bash
~/.config/systemd/user/
```

can run like normal background services without requiring the user to be logged in.

The important bit is that the services still run as that user, not as root. For Virtualmin, this would be especially useful because a specific account owner/client could be allowed to manage their own app services inside their own user scope, without touching global system services or affecting other users on the server.

So instead of creating root-owned system services for every hosted app, each Virtualmin user could have isolated per-user services managed through their own account.

[...]

PR #2748, opened 2026-06-02
Add advanced systemd unit and user unit management

iliaross:
Howdy, Jamie!

This PR expands the "Bootup and Shutdown" module’s systemd support with creation and management for services, timers, sockets, paths and targets. It also adds user-scoped systemd units, linger controls, status and log actions, tabbed unit listings by type, and contextual help for the new options.

For user-scoped units, the implementation includes several safety guards because unit files live under user-controlled home directories:

- User accounts are validated with system account data before any user-unit operation is attempted.
- User unit names are restricted to known systemd unit suffixes and safe filename characters.
- User unit files are limited to direct children of `~/.config/systemd/user`.
- Symlinked `.config`, `.config/systemd`, and `.config/systemd/user` paths are rejected.
- User unit reads, writes, directory creation, and deletes are performed after dropping privileges to the target Unix user.
- File operations re-check paths close to the actual read/write/delete operation to reduce symlink race exposure.
- User unit create failures roll back half-created files when daemon reload fails.
- User-provided unit names, owners, paths, command output, and logs are HTML-escaped before display.
- systemctl, journalctl, and loginctl command arguments are shell-quoted before execution.
- User services omit `User=` and `Group=` directives because they already run under the selected user’s systemd manager.

Together, all these changes will now allow Webmin admin to manage both system and user systemd units while keeping user-controlled home-directory paths from becoming root-level file read/write/delete exploits.

Implemented in response to these two issue requests  https://github.com/webmin/webmin/issues/2733 and https://github.com/webmin/webmin/issues/2734

Screenshot 2026-06-02 at 20 34 56
Screenshot 2026-06-02 at 20 35 07
Screenshot 2026-06-02 at 20 35 38
Screenshot 2026-06-02 at 20 35 55
Screenshot 2026-06-02 at 20 36 33


rubenferreira97:
@iliaross Awesome work, much appreciated! This looks like a great addition, and the security measures around user-scoped units seem very well thought out. I'll definitely give it a try and do some testing.

I do have one question regarding user units: if a user logs into Virtualmin (for example, a domain owner or a user created through Virtualmin), would they be able to manage their own user units?

iliaross:
> I do have one question regarding user units: if a user logs into Virtualmin (for example, a domain owner or a user created through Virtualmin), would they be able to manage their own user units?

I expected this question to come up. Well, right now, it's only meant for master administrators.

Support for domain owners may be added in future updates if there's enough interest.

rubenferreira97:
@iliaross Should I create a new issue to track this since my original one was closed? I think it would be helpful for tracking interest.

init/log_parser.pl
```diff
@@ -8,23 +8,47 @@
 sub parse_webmin_log
 {
 local ($user, $script, $action, $type, $object, $p) = @_;
-if ($action eq 'modify') {
+
+# User unit names and owners can come from user-controlled home directories, so
+# escape them before inserting into the log viewer HTML.
```
jcameron: Why not just escape everything?

jcameron:
Very impressive! Did you consider if it would make sense to have a separate systemd module for all this new functionality though?

init/log_parser.pl
iliaross: We need to track `log_user_*` and the other actions separately. Still, I agree the comment was a bit confusing. I fixed it.

iliaross:
> @rubenferreira97 Should I create a new issue to track this since my original one was closed? I think it would be helpful for tracking interest.

Yeah, open a new issue for this if you think it’s really needed.

> @jcameron Very impressive!

Thanks! 

> @jcameron Did you consider if it would make sense to have a separate systemd module for all this new functionality though?

I remember we discussed this before, and I still think the systemd logic should stay in the `init` module.

I don’t see much benefit in creating a separate module for it. We don’t have separate modules for upstart, init, rc, and other init systems, even though systemd is broader in scope. In practice, a system only uses one init system, and that’s what the Bootup and Shutdown module is meant for.

Also, other modules, like `virtual-server`, already rely on `init`, so splitting systemd out would add extra work and compatibility concerns without a clear upside.

init/edit_systemd.cgi
```diff
@@ -1,88 +1,470 @@
...
+					"style='display:none'" ]);
 
 	# Start at boot?
-	print &ui_table_row($text{'systemd_boot'},
+	print &ui_table_row(&hlink($text{'systemd_boot'}, "systemd_boot"),
 			    &ui_yesno_radio("boot", 1));
+
+	# User service controls
+	my $default_unituser = $unituser ||
+		($remote_user && $remote_user ne "root" ? $remote_user : undef);
```
jcameron: We can't safely assume that the root user is called "root" - for example in a install from setup.sh, the default Webmin user is called "admin". You'd need to find a way to use Webmin ACL system to treat non-root users differently.

init/index.cgi
```diff
@@ -329,70 +330,40 @@ elsif ($init_mode eq "upstart" && $access{'bootup'}) {
...
-			  &ui_text_color("$text{'no'}", 'warn'),
-			$u->{'status'} == 1 ? &ui_text_color("$text{'yes'}", 'success') :
-			  $u->{'status'} == 0 ?
-			  &ui_text_color("$text{'no'}", 'warn') :
-			  "<i>$text{'index_unknown'}</i>",
-			]);
+	# Show systemd units by type; keep user units on their own tab.
+	# Query parameters only choose the tab and optional user context.
+	print &systemd_index_style();
+	my $scope = $in{'scope'} eq 'user' ? 'user' : '';
```
jcameron, line 343: Be very very careful about using `$in` in `index.cgi` scripts, as they do no referrer check and are subject to XSS attacks (as we saw in other modules recently)

init/index.cgi
```diff
@@ -450,3 +421,252 @@ print &ui_buttons_end();
...
 &ui_print_footer("/", $text{'index'});
 
+# index_boot_system_title()
+# Returns the boot-system label shown in the page header.  For systemd
+# systems, prefer the exact systemctl version string so the title reflects
+# what is installed on the host.
+sub index_boot_system_title
+{
+return $text{'mode_'.$init_mode} if ($init_mode ne "systemd");
+my $systemctl = &has_command("systemctl");
```
jcameron, line 438: Can systemd exist without the `systemctl` command?

init/edit_systemd.cgi
```diff
@@ -1,88 +1,470 @@
...
+			    1, undef, \@service_row);
+	print &ui_table_row(&hlink($text{'systemd_stoppost'}, "systemd_stoppost"),
+			    &ui_textarea("stoppost", undef, 3, 80),
+			    1, undef, \@service_row);
+	print &ui_table_row(&hlink($text{'systemd_reload'}, "systemd_reload"),
+			    &ui_textarea("reload", undef, 3, 80),
+			    1, undef, \@service_row);
+
+	print &ui_hidden_table_end("advanced");
+	my $systemd_js = <<'EOF';
```
jcameron, line 244: Please don't use JS to show and hide fields like this, instead just add different links on index.cgi for creating different unit types. In fact, if they are so different, why not create edit_timer.cgi , edit_socket.cgi , etc? The code would be much simpler, and more like what we do elsewhere in Webmin.

init/edit_systemd.cgi
```diff
@@ -1,88 +1,470 @@
...
 
 	# Config file and contents
-	print &ui_table_row($text{'systemd_file'},
-			    "<tt>$u->{'file'}</tt>");
+	print &ui_table_row(&hlink($text{'systemd_file'}, "systemd_file"),
+			    &ui_tag('tt', &html_escape($u->{'file'})));
+
+	# User unit files are read through the privilege-dropping helper so a
+	# user-controlled path cannot make root follow symlinks in the home tree.
+	$conf = $edit_user_scope ?
```
jcameron, line 405: Maybe the root user shouldn't be seeing user units at all? That would avoid the need for all this privilege dropping code where on bug could open the root user up to a stored XSS attack.

init/mass_systemd.cgi
```diff
@@ -1,104 +1,269 @@
...
+	if ($u->{'user_scope'}) {
+		push(@{$users{$u->{'user'}}}, $u->{'name'});
+		}
+	else {
+		push(@system, $u->{'name'});
+		}
+	}
+&webmin_log($action, 'systemd', join(" ", @system)) if (@system);
+foreach my $user (sort keys %users) {
+	&webmin_log($action, 'systemd-user', join(" ", @{$users{$user}}),
```
jcameron, line 266: Why the different log action type for user units, instead of just adding the user name as a param..

init/edit_systemd.cgi
iliaross: Agreed. That was not correct. Fixed here https://github.com/webmin/webmin/pull/2748/changes/750940e2f5437be5fc246430ee6563efe8719c25.

init/index.cgi
iliaross: Yes, I remember that clearly when I implemented it. I’m pretty sure all the data was escaped before being displayed, so there’s no reflected `%in` value being injected into the HTML.

init/index.cgi
iliaross: No, `systemctl` is always present on _systemd_ systems.

init/edit_systemd.cgi
iliaross: We could, yet, I'm sure don’t think having five separate links like `edit_service.cgi`, `edit_timer.cgi`, `edit_socket.cgi`, `edit_path.cgi`, and `edit_target.cgi` would make things simpler. For this module's use case, it would only make things more complicated, unlike other modules. These unit types **share** most of the same workflow, and splitting them into separate CGIs would **duplicate a lot of code**.

We also need some JS anyway for system vs user mode. User units should hide `User=` and `Group=`, show the owner and linger fields, and adjust the install target. Otherwise, we’d need separate system/user create pages for each unit type too, ending up with ten links and pages of duplicated code.

To be clear, the server still validates everything, and JS just keeps the form readable.

And, the index page already creates context-specific links by passing `unittype`, so clicking from the Timers tab opens the form with Timer selected.

init/edit_systemd.cgi
iliaross: The root user should still see user units too. Managing user-scoped systemd units is part of the admin role here, especially for the Virtualmin-style use case.

I also don’t believe this exposes stored XSS in its current shape. User-controlled values are validated before use where needed, URL-encoded in links, and HTML-escaped before display.

And, more broadly, each module has its own workflow. I don’t think we should force this or any other module to match another module’s CGI layout if that makes the user experience worse.

In this case, keeping one create and edit flow gives us a cleaner UI and one shared server-side validation path, while JS only hides fields that are not relevant to the selected mode. It couldn’t be better or simpler in this scenario, in fact.

init/mass_systemd.cgi
iliaross: `systemd-user` is used because scope and owner are different things. The log type says this was a user-scoped systemd unit action, while the `user` param says which Unix user owned the unit. This keeps parsing and wording simpler, especially for mixed mass actions...

jcameron:
> I don’t see much benefit in creating a separate module for it. We don’t have separate modules for upstart, init, rc, and other init systems, even though systemd is broader in scope. In practice, a system only uses one init system, and that’s what the Bootup and Shutdown module is meant for.

Looking at this some more, I really don't think it makes sense to include all this other systemd functionality in the Bootup and Shutdown module.

The original purpose of that module (in the UI and API) was to provide a mostly consistent UI for controlling which services start at boot time, regardless of the underlying system. By adding timers and sockets and so on, we're complicating the UI away from the original use case, which I think leads to over-complex code as I mentioned in my comments.

I think what would make a lot more sense is to move the other systemd functionality for timers/etc into a separate module. And make it focused on actions that the user wants to take, not the underlying features of systemd. For example, users don't really care that systemd is behind all these unit types, then just want to control which services start a boot time, and maybe create scheduled cron-like jobs using timers.

iliaross:
> Looking at this some more, I really don't think it makes sense to include all this other systemd functionality in the Bootup and Shutdown module.
> 
> The original purpose of that module (in the UI and API) was to provide a mostly consistent UI for controlling which services start at boot time, regardless of the underlying system.

Thanks, Jamie. I see your point, and if we were designing this from scratch, I’d fully agree that a separate systemd-focused module could be cleaner.

But in the current Webmin layout reality, I think keeping this in the existing `init` module is still the better _trade-off_.

The `init` module is **already** the central place for service management in Webmin. It already has the systemd backend, and many other Webmin modules already rely on it to control services.

Moving the extra systemd functionality into a new module would either duplicate quite a bit of existing code, split closely related service logic between two modules, or worse, make both modules depend on each other. That will make things much harder to maintain and also less clear for users.

> I think what would make a lot more sense is to move the other systemd functionality for timers/etc into a separate module. And make it focused on actions that the user wants to take, not the underlying features of systemd. For example, users don't really care that systemd is behind all these unit types, then just want to control which services start a boot time, and maybe create scheduled cron-like jobs using timers.

It doesn’t seem to reflect the actual situation, as the request came from users who wanted better systemd support in the existing module. On modern Linux systems, systemd is the common case, while the older init systems are mostly there for compatibility and niche/older systems.

So maybe the better compromise is to keep the same module and API, but adjust the UI on systemd systems specifically.

For example, we could rename the visible module title from “Bootup and Shutdown” to something broader, like "Boot and Shutdown Services".

Then the module can still focus on practical actions, without pretending it only manages old-style boot scripts. It would also be easy to let users hide non-`.service` units in the module config, making it configurable for the rare cases where people do not want to see other unit types.

> By adding timers and sockets and so on, we're complicating the UI away from the original use case, which I think leads to over-complex code as I mentioned in my comments.

Being quite familiar with systemd, it is hard for me to fully agree with this point, because timers and sockets also fit this reasonably well. They can be **enabled at boot**, **started and stopped** like other units, and sockets can activate services on demand. So they are **not completely separate from boot** and service management, even if they are not "boot scripts" in the old sense.

And, also, notice how clean and still simple it looks compared to the old UI, while providing much more functionality to meet advanced users expectations:

Before:

image

After:

image

----

It basically looks and feels the same, which means the same group of users can keep using the old functionality as before, while more advanced users can get the extra controls they need.

Creating a new module does not seem very clear to me either. What role would the existing "Bootup and Shutdown" module play then? Would we list services in both modules, keep services in the old module and move only timers, sockets, and other unit types into the new one, or retire the old module completely?

All of those options feel messy and unnecessary, especially considering Webmin already has a lot of systemd-related code in the existing `init` module.

That is my reasoning, and my preference would be to keep this in the existing `init` module, but give it a better name like "Boot and Shutdown Services" on systemd systems.

I would also clean up the module config so it does not show options that are unrelated to systemd and visa versa. Luckily, Webmin’s config API already makes that easy to do. I’ll submit a follow-up commit for that in this PR.

Please let me know if my explanation sounds reasonable and helps clarify my thinking a bit.

rubenferreira97:
As a user, I would like to add that I really like the direction of this PR.

I personally read the module/menu purpose as something closer to **“Boot and Shutdown Services”**, so I think that rename makes sense and better reflects what the module is already doing on modern systemd-based systems.

I also want to reiterate that while I think this PR is a good improvement on its own, I would still love to see Virtualmin user-unit management considered as a follow-up.

In particular, I still believe Virtualmin users, for example, a domain owner or a user created through Virtualmin, should eventually be able to manage their own user units. Without that, several important use cases are lost.

For example, I may want to give a project account, such as `project1`, to a developer or a client so they can configure and manage their own application stack using user units. Giving them root access is not a solution in that scenario.

So overall, I support the direction of this PR, I think renaming the module on systemd systems to something like **“Boot and Shutdown Services”** makes sense, and I still think user-unit management for Virtualmin users should be considered an important follow-up feature because it enables real-world delegation without requiring root access.

iliaross:
As suggested above in my previous comment I have updated module config to display only entries for selected boot system to keep it clean and not confusing for the user, e.g.:

image
image
image

----

When, for example, "SysV init" is selected, the module config page will have config options loaded accordingly:
image
That works without any JS ad uses only Webmin config API.

----

And, I also changed the module title to make it more suitable for all systems as suggested in my previous comment as well:

image

iliaross:
> I also want to reiterate that while I think this PR is a good improvement on its own, I would still love to see Virtualmin user-unit management considered as a follow-up.
> In particular, I still believe Virtualmin users, for example, a domain owner or a user created through Virtualmin, should eventually be able to manage their own user units. Without that, several important use cases are lost.
> I still think user-unit management for Virtualmin users should be considered an important follow-up feature because it enables real-world delegation without requiring root access.

Yes, sure. Once Jamie and I agree on the final direction, hopefully merge this PR, I’ll open a new one for it. It should be straightforward since the systemd user model is already well covered here.

jcameron:
So I still think that this new functionality is awesome, but it should really be split up into separate modules.

Consider this - we have a separate modules already for xinetd , crontab and at. Even though systemd is taking over all that functionality with sockets and timers, users don't care about the underlying technology that's implementing them. A user who wants to run commands on a regular basis shouldn't need to know what systemd is to find the module, and certainly wouldn't be looking for a module called Bootup and Shutdown.

My strongly preferred approach would be to leave the support for systemd services in the Bootup and Shutdown module, and create new modules named something like "Scheduled Timers" and "Network Services" that manage timer and socket units. I don't mind if the Perl APIs for systemd support remains in the `init` module which is then called by these new modules, but they should be clearly separate modules with their own UIs.

iliaross:
> Consider this - we have a separate modules already for xinetd , crontab and at.

Yeah, new users looking for "run this every hour" probably **won’t naturally** search for systemd timers, and users looking for "socket activation" may not think of "Boot and Shutdown". I agree with that concern.

However, systemd services, timers, sockets, paths, targets, user units, linger, logs, enable/disable state, status, dependencies, and paired units are **all part of one systemd** control plane. If we split them now, we'll just create fragmented unusable thin modules that mostly duplicate code or deep-link back into the same `init` module.

Also, `crontab` is not going away from Linux any time soon, realistically not for decades, if ever. systemd timers are useful for some operational workflows, but they do not replace the simplicity, portability, and per-user convenience of cron. So "Scheduled Cron Jobs" should remain its own module, while "Scheduled Timers" could be confusing because systemd timers are not a separate Linux scheduling subsystem. They are **_systemd unit_** types.

The `xinetd` and `crontab` analogy is not quite the same. Those are separate technologies with their own configuration models. systemd timer and socket units, by contrast, only make sense as _**systemd units**_ and **often depend** on paired service units, shared enablement state, logs, status, and dependency handling.

The tabbed unit view and contextual create links already make this understandable as users can manage Services, Timers, Sockets, Paths, Targets, and User units in one place without scattering systemd state across multiple modules.

> Even though systemd is taking over all that functionality with sockets and timers, users don't care about the underlying technology that's implementing them.

I think there is a contradiction here. The concern is that users may not discover timers and sockets under the current module name, but splitting them into separate modules would make systemd state **even more fragmented**.

In reality, users have not asked for separate modules for each systemd unit type. If anything, they expected this as added functionality in the existing boot/services module, where systemd service management already lives. 

A good Webmin precedent is Postfix. We do not split aliases, virtual maps, transport maps, restrictions, queues, and service controls into separate modules just because they are different operational tasks. They stay together because they are one Postfix configuration and runtime domain. Same with Users and Groups.

systemd is the same kind of case, where timers, sockets, paths, targets, and services are not separate products. They are different unit types managed by the same daemon, commands, logs, enablement model, dependency graph, and user/system scope.

Splitting them would **_hide those relationships_** and make common workflows harder, especially for **_paired units_** like `foo.timer` + `foo.service` or `foo.socket` + `foo.service`—these are real concepts that should not and cannot be split, and are too important to ignore. For example:

 - `apt-daily.timer` + `apt-daily.service`: package metadata refresh on Debian/Ubuntu.
 - `dnf-automatic.timer` + `dnf-automatic.service`: automatic update checks/installs on Fedora/RHEL-like systems.
 - `logrotate.timer` + `logrotate.service`: periodic log rotation on systems using the systemd timer.
 - `certbot.timer` + `certbot.service`: renew TLS certificates.
 
That said, common units like `logrotate.timer`, `dnf-automatic.timer`, `apt-daily.timer`, `certbot.timer`, or `etckeeper.timer` are only **half the workflow**, as each triggers a paired `.service`, so managing `.timers` separately from services would hide the relationship users often need to inspect or edit.

Though I agree that neither "Bootup and Shutdown" nor "Boot and Shutdown Services" title really fit the modern systemd setup. I’d rename the module to "**Systemd Services and Units**," and keep "Bootup and Shutdown" for other systems.

Lastly, I also checked some other server admin tools, and it uses the same pattern with one systemd services section with tabs for services, timers, sockets, paths, and targets, not separate modules. I sent more details in DM.

iliaross:
> Though I agree that neither "Bootup and Shutdown" nor "Boot and Shutdown Services" title really fit the modern systemd setup. I’d rename the module to "**Systemd Services and Units**," and keep "Bootup and Shutdown" for other systems.

I’ve worked on this and found a way to do it without overhead. The latest simple commit https://github.com/webmin/webmin/commit/c11d67d989d8555b48b575647ef2c10aef294806 to the master branch adds support for alternative module descriptions using `module.info`. And, this patch https://github.com/webmin/webmin/pull/2748/commits/14f90afb89b6aba2c09ae0389be8609ba4f5d041 will display the correct module title in the UI for all systemd systems.

image

I think this is the best way forward, considering all the reasoning I gave in my previous comment.

Take a fresh look please.

jcameron:
Sorry no, it really doesn't make. sense to jam all this unrelated functionality into a single module just because systemd is doing so. The fact that we have to dynamically rename the "Bootup and Shutdown" module because these extra features are in there should be a sign that it's trying to do too much. 

Please undo the patches https://github.com/webmin/webmin/commit/c11d67d989d8555b48b575647ef2c10aef294806 and https://github.com/webmin/webmin/commit/14f90afb89b6aba2c09ae0389be8609ba4f5d041 , which won't be needed if you just simply create separate modules for the other systemd functionality and leave the `init` module to do what it has always done, which is control which servers are running and which are started at boot time.

Honestly I blame `systemd` itself for this mess, because it goes against the Unix practice of single tools that do one thing well. They have attempted to merge init.d, cron, at and inetd into one huge service, despite all these being different functions. But we don't have to make the same mistake in Webmin!

iliaross:
> The fact that we have to dynamically rename the "Bootup and Shutdown" module because these extra features are in there should be a sign that it's trying to do too much.

Jamie, then you shouldn’t have added _systemd_ support to "Bootup and Shutdown" in the first place. You should have created a separate _systemd_ module instead, since _systemd_, starting from version 1, introduced the **inseparable concept** of services, timers, sockets, paths and targets.

> Please undo the patches https://github.com/webmin/webmin/commit/c11d67d989d8555b48b575647ef2c10aef294806 and https://github.com/webmin/webmin/commit/14f90afb89b6aba2c09ae0389be8609ba4f5d041 , which won't be needed if you just simply create separate modules

Can do, yet only the first patch is in master and needs to be reverted, since the other one is the part of this PR.

And, it’s very unlikely that I’ll be able to help create separate modules for _systemd_ timers, sockets, paths, and targets for both system and user units. That’s the wrong approach, because that’s not what _systemd_ is or what users expect from it.

_Systemd_ has worked the same way for the past 16 years, with services, timers, sockets, paths, and targets all following one consistent inseparable model.

> Honestly I blame systemd itself for this mess, because it goes against the Unix practice of single tools that do one thing well. They have attempted to merge init.d, cron, at and inetd into one huge service, despite all these being different functions. But we don't have to make the same mistake in Webmin!

It would be a huge mistake to try to force Linux users to think our way, especially since _systemd_ is used on about 90% of Linux systems today. That would be a much bigger loss instead.

And, I already gave clear reasons and examples showing why the concept of units cannot and must not be separated in _systemd_. This should be considered and discussed here, because it is a core concept of _systemd_, and going against it, no matter what we wish it to be, creates more problems than it solves.

----

I could close this PR, as I don't disagree that rewiring existing `init` module to squeeze in proper _systemd_ support is suboptimal. Yet, it isn't as bad as trying to work around past mistakes, than actually solving the problem properly.

Although still, if we look at [systemd](https://en.wikipedia.org/wiki/Systemd) with an unbiased perspective, it is still an init system:

> systemd is a [software suite](https://en.wikipedia.org/wiki/Software_suite) for system and service management on [Linux](https://en.wikipedia.org/wiki/Linux)[[7]](https://en.wikipedia.org/wiki/Systemd#cite_note-7) built to unify service configuration and behavior across [Linux distributions](https://en.wikipedia.org/wiki/Linux_distribution).[[8]](https://en.wikipedia.org/wiki/Systemd#cite_note-8) **Its main component is an [init](https://en.wikipedia.org/wiki/Init) system** used to [bootstrap](https://en.wikipedia.org/wiki/Bootstrapping_(computing)) [user space](https://en.wikipedia.org/wiki/User_space) and manage [user processes](https://en.wikipedia.org/wiki/Process_(computing)).

That means sockets, timers, and other unit types **participate in the boot process** and clearly belong to the system boot concept, and therefore to the "Bootup and Shutdown" module! Red Hat sees it the same way, and I proved that to you in yesterday’s DM.

All that said, and something I could likely do optionally is building a fully functional Webmin _systemd_ module _but_ for all unit types that _systemd_ is designed to manage.

Essentially, it would keep the same UI style we can see now, which users and you’ve mentioned several times is cool and awesome.

In that case, `init` module would share some existing API with the new `systemd` module but "Bootup and Shutdown" would be hidden from the UI as the new `systemd` module named "Systemd Services and Units" would do all the work for all unit types.

However, if you want me to pretend and portray _systemd_ as something it isn’t by design or by how millions of people understand it, that’s entirely your choice and I respect that. But then I don’t think I'd be able to participate in it. Unless, of course you can reasonably convince me otherwise.

iliaross:
After a long but productive call, Jamie and I agreed on the plan: we’ll create a new, more advanced Webmin `systemd` module and keep the current `init` module as a simpler, service-only version.

I’ll close this PR and open a new one soon for the new Webmin `systemd` module, plus another with updates to the existing `init` module.

iliaross:
Alright, the new "Systemd Services and Units" module has been fully implemented in this PR: https://github.com/webmin/webmin/pull/2760. It also includes support for Virtualmin and Webmin safe users.

PR #2749, opened 2026-06-03
Fix Webmin temp directory setup in Advanced Options

iliaross:
Hey Jamie!

This PR fixes saving a custom Webmin temp directory from "Webmin Configuration → Advanced Options".

Previously, setting a path like /`var/webmin/tmp` failed if the directory did not already exist. Users had to create it manually, and it was easy to end up with a [bad parent](https://forum.virtualmin.com/t/error-while-installing-phpmyadmin/137362/4?u=ilia) directory such as `/var/webmin` with `0700`, which made the saved temp path unusable.

This change makes Webmin handle the safe parts automatically:

- Creates missing temp directories and parents as `0755`

- Validates existing parent directories are traversable by group/other

- Requires the final Webmin temp directory to be root-owned with mode `0755`

- Allows shared temp dirs like /var/tmp when root-owned and `1777`

- Shows a clear error when existing permissions must be fixed manually


web-lib-funcs.pl
```diff
@@ -538,9 +538,12 @@ sub tempname
...
 		while($tries++ < 10) {
 			my @st = lstat($tmp_dir);
-			last if ($st[4] == $< && (-d _) &&
-				 ($st[2] & 0777) == 0755);
 			if (@st) {
+				my $mode = $st[2] & 07777;
+				# Accept Webmin-private dirs and standard shared
+				last if ($st[4] == $< && (-d _) &&
+					 ((($mode & 0777) == 0755) ||
+					  $mode == 01777));
```
jcameron: What does this do exactly? We have to be very very careful with the permissions on the temp dir to avoid the risk of an attacker pre-creating /tmp/.webmin as a link or with the wrong ownership.

webmin/change_advanced.cgi
```diff
@@ -141,3 +143,114 @@ sub allowed_temp_dir
 my ($t) = @_;
 return $t eq "/" || $t =~ /^\/[^\/]+\/?$/ ? 0 : 1;
 }
+
+# Create a configured Webmin temp directory if needed, and validate existing
+# path components without changing permissions on directories the user made.
+sub setup_advanced_temp_dir
```
jcameron: This is a lot of code just to set a temp dir! Can you explain what it is doing and what it is checking?

web-lib-funcs.pl
iliaross: It keeps the existing safety checks and only expands the allowed permission modes. This line accepts the temp directory only if all of these are true:

- it is owned by the current Webmin process user, i.e. `root`
- it is a real directory, not a symlink
- it has either Webmin’s private temp-dir mode `0755` or standard shared temp-dir mode `01777`

The only behavior change is that root-owned sticky temp dirs like `/var/tmp` are now treated as valid instead of being “fixed” back to `0755`.

webmin/change_advanced.cgi
iliaross: Yeah, it looks like a lot because it’s replacing a very small but too-naive check.

Previously we only asked: “does this directory exist?” That failed for new paths like `/var/webmin/tmp`.

Now it does the practical setup work:

- If the directory path is missing, create it.
- If parent directories are missing, create those too.
- If a parent already exists, make sure Webmin can actually get through it. For example, `/var/webmin` cannot be `0700` (as in the [ticket ](https://forum.virtualmin.com/t/error-while-installing-phpmyadmin/137362/4?u=ilia)).
- If the final temp directory already exists, make sure it is safe: owned by root and either `0755` or normal shared-temp `1777`.
- If something already exists in the path but is not a directory, stop with an error...

...and, the reason it does not just `chmod` everything is that existing directories may belong to the admin’s setup, and silently widening permissions on them could be surprising.

So Webmin creates what is missing, but only validates what already exists.

I think this is honestly the smartest and best way to handle it...

webmin/change_advanced.cgi
```diff
@@ -141,3 +143,114 @@ sub allowed_temp_dir
...
+			     &advanced_temp_dir_perms_text(), "$!"));
+	}
+&advanced_temp_dir_perms_ok($dir) ||
+	&error(&text('advanced_etempperms', $dir,
+		     &advanced_temp_dir_allowed_perms_text()));
+return $dir;
+}
+
+# Permissions used for newly created Webmin temp directories.
+sub advanced_temp_dir_perms
```
jcameron: Couldn't this just be a variable, not a sub?

webmin/change_advanced.cgi
```diff
@@ -141,3 +143,114 @@ sub allowed_temp_dir
...
+# Octal text form used in error messages.
+sub advanced_temp_dir_perms_text
+{
+return sprintf("%04o", &advanced_temp_dir_perms());
+}
+
+# Existing final temp dirs may be Webmin-private, or standard shared temp dirs.
+sub advanced_temp_dir_allowed_perms_text
+{
+return "<tt>".&advanced_temp_dir_perms_text()."</tt> or <tt>1777</tt>";
```
jcameron: Shouldn't the word "or" be translatable? Also, is it necessary to have a sub just to return one string? Seems like it could be done in a `text` call.

webmin/change_advanced.cgi
```diff
@@ -141,3 +143,114 @@ sub allowed_temp_dir
...
+# Permissions used for newly created Webmin temp directories.
+sub advanced_temp_dir_perms
+{
+return 0755;
+}
+
+# Octal text form used in error messages.
+sub advanced_temp_dir_perms_text
+{
+return sprintf("%04o", &advanced_temp_dir_perms());
```
jcameron: Same, this could just be a string

webmin/change_advanced.cgi
jcameron: I think this should be simplified to not create the custom temp dir, because if the user makes a typo and enters `/varr/tmp/webmin` it will create that path. Instead just keep it simple and only check that the permissions are correct.

web-lib-funcs.pl
jcameron: Do we really want to allow a temp dir with 0777 permissions? That seems dangerous, because any non-root user could pre-create a malicious symlink that gets used as a temp file by another process running as root.

webmin/change_advanced.cgi
iliaross: Yes, sure. Fixed that on the latest commit https://github.com/webmin/webmin/pull/2749/commits/e434d0b1389edc6a1e9ad507c6475ad073c03acc.

webmin/change_advanced.cgi
iliaross: Yes, fixed that too on the same commit https://github.com/webmin/webmin/pull/2749/commits/e434d0b1389edc6a1e9ad507c6475ad073c03acc.

webmin/change_advanced.cgi
iliaross: Yes, agreed. Fixed here https://github.com/webmin/webmin/pull/2749/commits/501bddabc85e341a8c8b31b9440588fa6043e72b.

webmin/change_advanced.cgi
iliaross: Well, I get the typo concern, however, I believe only validating keeps the exact problem alive.

This already happened in the [forum report](https://forum.virtualmin.com/t/error-while-installing-phpmyadmin/137362/4?u=ilia): `/var/webmin/tmp` itself looked correct (`0755`), but its parent `/var/webmin` was effectively not searchable for the process that needed to enter it, so phpMyAdmin failed with:

```
cd: /var/webmin/tmp/...: Permission denied
```

The goal of this PR is to avoid sending users to the shell to manually create or fix these paths, since that’s how we end up with parent directories like `0700` in the first place.

Also, having to go to the shell to create a directory defeats the purpose of having a Webmin UI too.

And, the code is not blindly chmodding existing admin-created directories. It **only creates** missing path components with **known permissions**. If something already exists, it validates it and **errors if the ownership or permissions are unsafe**. That means Webmin takes responsibility for what it creates, but does not unexpectedly widen permissions on directories the admin already had.

For typos like `/varr/tmp/webmin`, yes, that can create the wrong path. So what should we do about it? I can only suggest to pay attention.

As the user explicitly entered and saved a custom temp directory, and creating the configured directory is the behavior that prevents this recurring support issue.

If we want an extra guard, I’d rather add a warning or confirmation when creating a **new top-level path component**, but I don’t think we should drop auto-creation entirely.

web-lib-funcs.pl
iliaross: This check is not allowing plain `0777`. It allows exact `01777`, meaning world-writable **with the sticky bit**, which is **the normal/default mode** for shared temp directories like `/tmp` and `/var/tmp` on contemporary Linux systems.

So this is specifically to avoid breaking valid directories like `/var/tmp`, not to accept unsafe `0777` temp dirs. A plain `0777` directory without sticky bit would still fail this check.

To be clear, these are normal defaults for all shared temp directories on contemporary Linux systems:

```
root@rocky9-gpl:~# ls -ld /tmp
drwxrwxrwt. 17 root root 4096 Jun  5 11:29 /tmp
root@rocky9-gpl:~# ls -ld /var/tmp
drwxrwxrwt. 12 root root 4096 Jun  5 00:00 /var/tmp
root@rocky9-gpl:~# 
```

And as for a pre-created malicious symlink, that’s exactly why this check is strict. `01777` is the normal mode for `/tmp` and `/var/tmp` on modern Linux, but we only accept it for an existing path that is owned by root and is a real directory, not a symlink. Plain `0777` is not accepted, and files inside still use Webmin’s random temp name creation instead of blindly reusing attacker-created paths.

jcameron: I don't think we should be using just /var/tmp as a Webmin root directory at all though - there's a good reason why the default is /tmp/.webmin and not just /tmp . Making /tmp/.webmin mode 0755 protects against a lot of potential attacks via malicious file creation. Also, there is other code in Webmin that cleans up leftover files in /tmp/.webmin , which would be dangerous if the temp dir is shared by other users.

webmin/change_advanced.cgi
jcameron: See my comment above about the danger of using a temp dir that already exists. Changing the temp dir is a pretty niche feature that few if any Webmin users will need to do, so instead of adding a lot of code to try to automate that, we should just check that the permissions are reasonable and that the directory exists.

web-lib-funcs.pl
iliaross: > I don't think we should be using just /var/tmp as a Webmin root directory at all though - there's a good reason why the default is /tmp/.webmin and not just /tmp . Making /tmp/.webmin mode 0755 protects against a lot of potential attacks via malicious file creation. Also, there is other code in Webmin that cleans up leftover files in /tmp/.webmin , which would be dangerous if the temp dir is shared by other users.

Jamie, I know, and I agree with that. Webmin should not use `/tmp` or `/var/tmp` directly as its temp root. I never meant that, and this patch isn’t suggesting it.

The case we’re fixing is something like `/var/tmp-webmin/tmp`, or any similar path a user enters in the UI. The final Webmin temp dir is private, but one of its parents may already exist with bad permissions, like `/var/tmp-webmin` being `0700`. Users can enter paths like `/var/my-temps/webmin-temp/tmp`, so we need to make sure the path is either created correctly or already has usable permissions if was created by the user. Otherwise we end up with the same failure as in the forum report.

If the concern is that the current check also accepts `01777` for the final configured temp dir, I can tighten that. The final Webmin temp dir can be root-owned `0755` only, while parent dirs only need to be searchable, meaning at least `0711`.

webmin/change_advanced.cgi
iliaross: > See my comment above about the danger of using a temp dir that already exists.

See my comment to your comment.

> Changing the temp dir is a pretty niche feature that few if any Webmin users will need to do

Not anymore. We now show a warning about `ramfs` being used for `/tmp/.webmin` as the default temp directory, so users are more likely to change it.

They’ve already done that and got the permissions wrong, as in the forum report. If we leave this as a manual setup step, it will keep happening.

> We should just check that the permissions are reasonable and that the directory exists.

Yes, and this patch is doing that, and also handles the part users are most likely to get wrong, which is **creating the path** with usable permissions.

It still checks that existing directories are reasonable. It does not blindly `chmod` what the admin already created. It only creates missing directories correctly, so we don’t end up with `/var/tmp-webmin` as `0700` and a private temp dir inside it that Webmin can’t actually use.

I don't know how I can be more clear about it.

Yet, I’ll add a patch to harden to `/var/tmp` so users don’t end up using it as their final private temp directory.

web-lib-funcs.pl
iliaross: > I don't think we should be using just /var/tmp as a Webmin root directory at all though

Now it will be just forbidden to use any temporary system directory https://github.com/webmin/webmin/pull/2749/commits/9577737aebc78da16c32426d0d048579a8950640.

iliaross:
Alright, this should be _all_ good now.

To summarize once again, this patch prevents users from ending up with incorrect permissions on a custom Webmin temporary directory, like in the forum report, and should help avoid the same type of issues in the future.

webmin/change_advanced.cgi
```diff
@@ -6,14 +6,21 @@ require './webmin-lib.pl';
...
+
 # Save global temp dir setting
 if ($in{'tempdir_def'}) {
 	delete($gconfig{'tempdir'});
 	}
 else {
-	-d $in{'tempdir'} || &error($text{'advanced_etemp'});
 	&allowed_temp_dir($in{'tempdir'}) ||
 		&error(&text('advanced_etempallowed', $in{'tempdir'}));
+	$in{'tempdir'} = &setup_advanced_temp_dir(
```
jcameron: Doesn't calling `setup_advanced_temp_dir` here mean that the temp directory will be created even if `error` is called later on in this CGI? Best practice in Webmin is to not take any actions until all validation on the form is complete.

webmin/change_advanced.cgi
```diff
@@ -33,9 +40,10 @@ for($i=0; defined($tmod = $in{'tmod_'.$i}); $i++) {
 	next if (!$tmod);
 	$tdir = $in{'tdir_'.$i};
 	%minfo = &get_module_info($tmod);
-	-d $tdir || &error(&text('advanced_etdir', $minfo{'desc'}));
 	&allowed_temp_dir($tdir) ||
-		&error(&text('advanced_etempallowed', $in{'tempdir'}));
+		&error(&text('advanced_etempallowed', $tdir));
+	$tdir = &setup_advanced_temp_dir(
```
jcameron: Same issue here

webmin/change_advanced.cgi
jcameron: Actually why does the final directory need to be created at all when the `tempname` function already does that? I think this code could all be simplified drastically as follows : 

1- **Check** but do not change permissions on all the parent directories, and show an error if they are too tight or too loose. But there is no needed to change any permissions or create directories!

2 - If the directory the user enters does not exist, that's fine. It will be created in `tempname` , just like `/tmp/.webmin` is.

3 - If the directory does exist, then just check that the permissions are the same as what `tempname` accepts, and throw an error if not.

Since this is a niche use case, it's much safer and simpler to just warn the user if their path is non-existant or un-usable than to go around changing permissions. For example, what if the admin entered /etc/.webmin as the path ... would we really want to be changing permissions on /etc ??

iliaross:
> Actually why does the final directory need to be created at all when the `tempname` function already does that?

It can only create a target directory, it won't work with `/var/tmp/1/2/3/4/5/.webmin`. This is the reasons to have this change in the first place.

> Check but do not change permissions on all the parent directories

We are not going to change permissions on existing directories. We only validated that a parent directory inside `/tmp` or `/var/tmp`, such as `/var/tmp/custom1/custom2`, has at least 711 permissions. The final Webmin temporary directory, like `/var/tmp/custom1/custom2/.webmin`, will be created with the correct permissions.

> Doesn't calling setup_advanced_temp_dir here mean that the temp directory will be created even if error is called later on in this CGI? Best practice in Webmin is to not take any actions until all validation on the form is complete.

Yes, you're right, and this has been addressed in the latest patch. Thanks for pointing this out!!

webmin/change_advanced.cgi
jcameron: I don't think you really addressed this properly - it looks like the current code is using `advanced_temp_dirs_error` to clean up if validation fails later on in the CGI. The correct thing is to do **ALL** validation before taking any actions.

jcameron:
Please apply the changes we discussed today, and let me know. In particular we ideally should always create a `.webmin` sub-directory under whatever path they enter, to ensure that there is no danger of messing with an existing directory. This way we can avoid the need to check for a bunch of system directories.

webmin/change_advanced.cgi
mckaygerhard: > 1- **Check** but do not change permissions on all the parent directories, and show an error if they are too tight or too loose. But there is no needed to change any permissions or create directories!

@jcameron  This is great but! you should raise a big alert if such directory cannot be used or created by any external reason.. by exampel suddently read-only filesystem or similar.. 

@iliaross , he is right! he is on correct user and admin of the OS must know what it does! the software cannot handle such responsabilitie! and  the right way is warn and notify to the ADMIN (not user., @jcameron , cos the user only uses the system) if their path is non-existant or un-usable

iliaross:
> Please apply the changes we discussed today, and let me know. In particular we ideally should always create a `.webmin` sub-directory under whatever path they enter, to ensure that there is no danger of messing with an existing directory. This way we can avoid the need to check for a bunch of system directories.

Yeah, I’ll do that right after I submit the new Systemd Services and Units module.

jcameron:
Ok cool, let me know when this is ready for a re-review..

iliaross:
> Ok cool, let me know when this is ready for a re-review..

All done and ready now. It can be merged now. Please check again.

jcameron:
Looks good thanks!!!

iliaross:
You’re welcome, my pleasure!

PR #2752, opened 2026-06-05
Fix LE renewal to schedule by elapsed interval

iliaross:
Howdy, Jamie!

This turned out to be a bug after [this](https://forum.virtualmin.com/t/letsencrypt-auto-renew-in-webmin/137371/12?u=ilia) comment on the Virtualmin forum.

The Webmin SSL LE renewal setting is labeled as "Months between automatic renewal", but it was previously saved as a calendar-style cron month expression like `*/N`.

That is not the same as an elapsed renewal interval. Webmin’s cron matcher evaluates month schedules against calendar month numbers, so values like `*/5`, `*/12`, or values above `12` do not reliably mean “renew every N months”. This could cause uneven or dangerously late renewal timing.

This changes the renewal job to use Webmin cron’s elapsed `interval` support instead of calendar-month matching.

- Saves automatic renewal as `renew * 30 * 24 * 60 * 60` seconds.
- Clears the cron time fields so the scheduler uses the interval path only.
- Keeps `months => '*/N'` so the SSL UI can continue to display the saved renewal value.
- Resets the renewal timer only after a newly issued certificate.
- Preserves the existing renewal timer for settings-only saves.
- Migrates existing month-based Let's Encrypt renewal jobs during postinstall.

### DIFF
--- a/webmin/letsencrypt.cgi
+++ b/webmin/letsencrypt.cgi
@@ -219,11 +219,10 @@ if (&foreign_check("webmincron")) {
 		&webmincron::delete_webmin_cron($job) if ($job);
 		}
 	else {
-		# When a cert was just issued, delete and re-create the job so
-		# it gets a fresh ID and its elapsed-interval countdown restarts
-		# from now. Re-using the ID could keep a stale last-run time
-		# that is already past the interval, firing a renewal
-		# immediately and wasting a Let's Encrypt issuance
+		# A manual cert request does not update Webmin cron's last-run
+		# state, so re-create the job to start the elapsed renewal
+		# interval from now instead of immediately running an overdue
+		# job with the old ID
 		if ($job && $reset_renewal_time) {
 			&webmincron::delete_webmin_cron($job);
 			$job = undef;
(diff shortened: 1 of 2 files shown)


webmin/letsencrypt.cgi
```diff
@@ -219,14 +219,25 @@ if (&foreign_check("webmincron")) {
 		&webmincron::delete_webmin_cron($job) if ($job);
 		}
 	else {
-		my @tm = localtime(time() - 60);
+		# When a cert was just issued, delete and re-create the job so
+		# it gets a fresh ID and its elapsed-interval countdown restarts
+		# from now. Re-using the ID could keep a stale last-run time
+		# that is already past the interval, firing a renewal
+		# immediately and wasting a Let's Encrypt issuance
+		if ($job && $reset_renewal_time) {
```
jcameron, line 226: Are you 100% sure this is necessary? It seems like a deeper bug in Webmin if somehow the last-run time isn't being updated.

jcameron:
Mostly looks good, thanks!

webmin/letsencrypt.cgi
iliaross: I don't think it's a Webmin bug as manual renewal is not a cron run, so it expectedly doesn’t update the cron last-run timestamp. Recreating the job is the simplest way to restart the renewal timer without poking Miniserv’s internals.

I think it's fine.

jcameron: Oh sorry I missed that this was for a manual renewal! Ok yes this makes perfect sense, thanks..

PR #2753, opened 2026-06-06
Add minimal Alpine Linux compatibility

iliaross:
Hello, Jamie!

This PR adds baseline Alpine Linux support in Webmin with OS detection, APK package and update backend, OpenRC boot integration, ifupdown-style networking support and DHCPD defaults.

https://github.com/webmin/webmin/issues/2353


webmin-openrc-init
```diff
@@ -5,11 +5,7 @@
 
 
 name="webmin"
-# config file must be in sync with setup script, so i will not touch cos process will be the same as webmin-systemd
-# like calling updateboot.pl with a new procedure method defined as openrc and not just assumed sysvinit if not systemd is present
-# and also atboot.pl with new procedure for openrc the alpine linux init system at the moment
-#conf_file="WEBMIN_CONFIG/miniserv.conf"
-conf_file="/etc/webmin/miniserv.conf"
+conf_file="WEBMIN_CONFIG/miniserv.conf"
```
jcameron, line 8: Should this change to the gentoo init file be part of this PR if it's for Alpine Linux?

jcameron:
Mostly looks good, thanks!

webmin-openrc-init
iliaross: I see the confusion. I think it's better to just rename the file. Check this patch https://github.com/webmin/webmin/pull/2753/commits/3215c0d0a3563e5b358d34b19ed9aff418240659.

mckaygerhard:
I will test it in few hours..  in a baremetal or virtual machine.. cos i currently only use alpine into containers.. but it seems looks good..  the renaming of the openrc script is a right one cos geento seems uses other init system now!

makedebian.pl
```diff
@@ -81,7 +81,7 @@
 system("cd $usr_dir && rm -f mount/openbsd-mounts*");
```
mckaygerhard, line 81: I dont know but why you use the file "makedebian.pl" ??  what about a change in sourcered file not in sync with alpine?

iliaross:
> the renaming of the openrc script is a right one cos geento seems uses other init system now!

The other init system? Which one?

mckaygerhard:
> The other init system? Which one?

As https://wiki.gentoo.org/wiki/Init_system OpenRC still is default BUT soome desktops as like Gnomepuach it depends on shitstemd as mandatory.. so some parts are included etc etc etc.. 

My comment means that the renaming from your PR works for any openrc based linux and was planned(asked in fact) in my issue! thanks for your work.. I will post feedback later

mckaygerhard:
Oh found that the openrc init script was not referenced or packaged in any place in repo.. so the the CI build fails!

PR #2760, opened 2026-06-12
Add Systemd Services and Units module

iliaross:
Hello Jamie!

This PR adds a standalone Systemd Services and Units module for managing systemd units across system and user scopes.

The module keeps systemd-specific behavior separate from the legacy Bootup and Shutdown module and is implemented as standalone `strict`/`warnings` Perl code rather than depending on its existing init helpers. Those helpers intentionally smooth over multiple init systems, while this module keeps systemd-specific file handling, user-manager behavior, ACL checks, and control operations explicit, scoped, and easier to audit.

It includes:

- Tabbed views for services, timers, sockets, paths, targets, storage, resources, devices, and user units
- Guided creation and editing for common unit types, with contextual fields, validation, and help
- User-scoped unit management with linger support and safe handling of home-directory unit files
- Runtime actions for start, stop, restart, enable, disable, status, logs, properties, dependencies, and system-unit mask/unmask
- Drop-in override inventory plus create, edit, and delete flows
- Manual unit-file editing with daemon reload reminders and actions
- Configurable module behavior, visible tabs, display options, and post-create navigation
- Comprehensive ACL controls for system/user scopes, actions, manual edits, drop-ins, linger, reload, backup, and user filters
- Safe Webmin user support through a scoped safe ACL preset
- Virtualmin integration for granting domain owners access to their own systemd user units
- Tests for unit generation, safety checks, ACL behavior, user-unit handling, backup coverage, and Perl::Critic compatibility

A companion [Virtualmin PR](https://github.com/virtualmin/virtualmin-gpl/pull/1234) adds template integration so domain owners can be granted scoped access to their own systemd user units when this module is installed.

Screenshots: 
<table>
  <tr>
    <td width="50%"><a target="_blank" href="https://github.com/user-attachments/assets/eaa189a5-a13d-4ea9-8ed1-9a475ae4bff4"></a></td>
    <td width="50%"><a target="_blank" href="https://github.com/user-attachments/assets/45c7bc8c-5112-4119-a83e-ac1d26acc774"></a></td>
  </tr>
  <tr>
    <td width="50%"><a target="_blank" href="https://github.com/user-attachments/assets/068fe39a-0bb2-4414-97c9-3c32667dbf38"></a></td>
    <td width="50%"><a target="_blank" href="https://github.com/user-attachments/assets/431d6efe-d9ee-42e1-a6d1-5f6d8c64cffa"></a></td>
  </tr>
  <tr>
    <td width="50%"><a target="_blank" href="https://github.com/user-attachments/assets/cf43c4e7-2648-466d-a6e4-4ea22592690f"></a></td>
    <td width="50%"><a target="_blank" href="https://github.com/user-attachments/assets/6319becf-291e-4de7-8ebf-01455f606ca9"></a></td>
  </tr>
  <tr>
    <td width="50%"><a target="_blank" href="https://github.com/user-attachments/assets/367f7593-56bd-432f-a3c3-b6b36c388d15"></a></td>
    <td width="50%"><a target="_blank" href="https://github.com/user-attachments/assets/a7be9b9f-6eb4-4917-8ab9-43dfdce727a2"></a></td>
  </tr>
  <tr>
    <td width="50%"><a target="_blank" href="https://github.com/user-attachments/assets/552ecf2c-35a6-4a4f-bf37-ad784125b637"></a></td>
    <td width="50%"><a target="_blank" href="https://github.com/user-attachments/assets/764a0a07-ac05-41c8-b877-5b4e47ad6fbc"></a></td>
  </tr>
  <tr>
    <td width="50%"><a target="_blank" href="https://github.com/user-attachments/assets/6345871d-7cf2-4cf4-b832-9c80161264a6"></a></td>
    <td width="50%"><a target="_blank" href="https://github.com/user-attachments/assets/4c587817-480d-4629-9997-8474f29a6cc9"></a></td>
  </tr>
  <tr>
    <td width="50%"><a target="_blank" href="https://github.com/user-attachments/assets/0281836f-09a9-42fa-baa8-9f6d79734ec6"></a></td>
    <td width="50%"><a target="_blank" href="https://github.com/user-attachments/assets/e600b190-1846-4545-8614-641b8114b001"></a></td>
  </tr>
  <tr>
    <td width="50%"><a target="_blank" href="https://github.com/user-attachments/assets/28ac9dd3-d7d6-4bc6-a7a3-5f2adbfe5f85"></a></td>
    <td width="50%"><a target="_blank" href="https://github.com/user-attachments/assets/b86f065a-f681-4f13-bfd4-c59d41cd3459"></a></td>
  </tr>
  <tr>
    <td width="50%"><a target="_blank" href="https://github.com/user-attachments/assets/471d1706-fef1-466d-9d3b-a151b7515629"></a></td>
    <td width="50%"><a target="_blank" href="https://github.com/user-attachments/assets/eb237d27-5767-4fa9-b1df-3c7f92527153"></a></td>
  </tr>
  <tr>
    <td width="50%"><a target="_blank" href="https://github.com/user-attachments/assets/5c2c92fe-5a62-4a00-9618-44e9e7822352"></a></td>
    <td width="50%"><a target="_blank" href="https://github.com/user-attachments/assets/4b923d41-e2ee-4e57-8f9d-0624c82fcb1c"></a></td>
  </tr>
  <tr>
    <td width="50%"><a target="_blank" href="https://github.com/user-attachments/assets/1fd6e843-b2e7-46e1-b5b7-895ffd1d97cf"></a></td>
    <td width="50%"><a target="_blank" href="https://github.com/user-attachments/assets/456d76d1-77d8-48b2-a03f-4a7406d82b84"></a></td>
  </tr>
  <tr>
    <td width="50%"><a target="_blank" href="https://github.com/user-attachments/assets/f4bd52c1-d23d-4d61-8995-b56b3f9645d1"></a></td>
    <td width="50%"><a target="_blank" href="https://github.com/user-attachments/assets/5c7f4a7a-b2c8-4cdd-9662-de8765a97a44"></a></td>
  </tr>
  <tr>
    <td width="50%"><a target="_blank" href="https://github.com/user-attachments/assets/99408c93-1177-491b-ac8b-58d0d544a30e"></a></td>
    <td width="50%"><a target="_blank" href="https://github.com/user-attachments/assets/b8719eb8-a280-4bad-8624-aea0f5dffaa3"></a></td>
  </tr>
  <tr>
    <td width="50%"><a target="_blank" href="https://github.com/user-attachments/assets/c73deb65-d89b-4868-a314-831b2ec4cbc9"></a></td>
    <td width="50%"><a target="_blank" href="https://github.com/user-attachments/assets/831a048f-34c0-4188-afb6-3a56cdafc4bb"></a></td>
  </tr>
  <tr>
    <td width="50%"><a target="_blank" href="https://github.com/user-attachments/assets/f35b3eeb-7116-4cf6-9712-8d45a12291db"></a></td>
    <td width="50%"></td>
  </tr>
</table>


swelljoe:
Claude has the following concerns:

🔴 HIGH — Confirmed stored XSS via mount Where= (fix recommended)

  edit_unit.cgi:253-255 builds an automount dropdown option label from a mount unit's Where= path and passes it to ui_select. Core ui_select (ui-lib.pl:1233) emits option labels raw (only the  value is escaped). The Where= value comes from a user-controlled .mount file and is filtered only by valid_path (systemd-lib.pl:1017), which blocks whitespace/=/\r\n\0 but allows < > / ".

  Exploit: a low-privilege operator who can write their own user .mount unit sets  Where=/srv</option></select><script>alert(document.domain)</script>  (no spaces, no second =, so it passes valid_path). When a higher-privilege admin opens the create-automount form scoped to that user, the script runs in the admin's session → privilege escalation.

  Fix: escape the label before handing it to ui_select:
  push(@automount_mounts, [ $mu->{'name'}, html_escape($label) ]);
  The Where= text path field at edit_unit.cgi:423 should be spot-checked too, though ui_textbox/ui_filebox value-escape.

🟡 LOW / hardening (no confirmed exploit)

  1. valid_path is a loose validator (systemd-lib.pl:949) — permits HTML and shell metacharacters; safe today only because every caller quotemetas (shell) or should html_escape (HTML). The XSS above is what happens when that invariant slips. Recommend tightening it to also reject <>"', as defense-in-depth for future call sites.
  2. backup ACL is not scope-split (backup_config.pl:14) — one boolean covers system unit-file contents. The shipped safeacl sets backup=0, so default non-root profiles are safe, but a hand-edited restricted profile with backup=1 would leak /etc/systemd/system config. Consider splitting into backup/backup_user.
  3. Unused edit_unit() library API (systemd-lib.pl:3861) writes override.conf from caller-supplied dir/filename with no validation or symlink guard — the one function not following the module's hardening pattern. Not reachable from any CGI today (no callers). Harden or remove before anything wires it up.
  4. State changes reachable via GET — set_linger.cgi and restart_user.cgi are GET links; no CGI checks REQUEST_METHOD. Protected by Webmin's Referer check and consistent with Cron's convention, so not exploitable by default. Optional: POST-gate them.

iliaross:
> Fix: escape the label before handing it to ui_select:

Thanks, Joe! Fixed that!

systemd/acl_security.pl
```diff
@@ -0,0 +1,81 @@
...
+		    $m == 5 ? scalar(getgrgid($o->{'users'})) : "", 13)." ".
+		group_chooser_button("gid", 0)."<br>" ],
+	   [ 4, $text{'acl_uid'}." ".
+		ui_textbox("uidmin", $o->{'uidmin'}, 6)." - ".
+		ui_textbox("uidmax", $o->{'uidmax'}, 6)."<br>" ],
+	 ]), 3, undef, undef, 1);
+print ui_table_hr();
+
+print ui_table_span(ui_tag('b', html_escape($text{'acl_section_view'})));
+foreach my $a (qw(view view_user status status_user logs logs_user)) {
```
jcameron, line 42: Instead of qw, just use a regular array like : 

"view", "view_user", "status", "status_user", "logs", "logs_user"

systemd/acl_security.pl
```diff
@@ -0,0 +1,81 @@
...
+our (%in, %text);
+
+# acl_security_form(options)
+# Outputs ACL controls for granting access to systemd unit management.
+sub acl_security_form
+{
+my ($o) = @_;
+my $m = $o->{'mode'} || 0;
+
+print ui_table_span(ui_tag('b', html_escape($text{'acl_section_users'})));
```
jcameron, line 17: Some comments on what each section configures would be nice

systemd/acl_security.pl
```diff
@@ -0,0 +1,81 @@
...
+print ui_table_span(ui_tag('b', html_escape($text{'acl_section_runtime'})));
+foreach my $a (qw(start start_user stop stop_user restart restart_user
+		  boot boot_user mask mask_user reload linger)) {
+	print ui_table_row($text{'acl_'.$a},
+			   ui_yesno_radio($a, $o->{$a}), 3);
+	}
+print ui_table_hr();
+
+print ui_table_span(ui_tag('b', html_escape($text{'acl_section_change'})));
+foreach my $a (qw(create create_user edit edit_user delete delete_user
```
jcameron, line 59: Same here about qw. In fact, drop it everywhere .. I have no idea why Perl has this wierd construct for creating arrays by splitting strings on spaces that no other programming language does.

systemd/systemd-lib.pl
```diff
@@ -0,0 +1,3938 @@
...
+	   dropin dropin_user manual manual_user reload linger backup));
+}
+
+=head2 systemd_user_unit_acl(user)
+
+Returns a safe ACL hash for managing one Unix user's systemd user units,
+without granting any system-unit access.
+
+=cut
+sub systemd_user_unit_acl
```
jcameron: Is this function used anywhere other than tests?

systemd/systemd-lib.pl
```diff
@@ -0,0 +1,3938 @@
...
+}
+
+=head2 systemd_can_enter_module(&acl)
+
+Returns 1 if the ACL allows any interactive access to this module.
+
+=cut
+sub systemd_can_enter_module
+{
+my ($acl) = @_;
```
jcameron: Why pass in the ACL as a parameter when %access is available as a global?

systemd/systemd-lib.pl
```diff
@@ -0,0 +1,3938 @@
...
+$acl{'uidmax'} = "";
+return %acl;
+}
+
+=head2 systemd_acl_bool(&acl, key)
+
+Returns a boolean ACL value.
+
+=cut
+sub systemd_acl_bool
```
jcameron, line 103: Couldn't this whole function be replaced with just `$access{$key}` ?

systemd/index.cgi
```diff
@@ -0,0 +1,522 @@
...
+# Display systemd system and user units grouped by unit type.
+
+use strict;
+use warnings;
+
+require './systemd-lib.pl'; ## no critic
+
+our (%access, %config, %in, %text);
+
+ReadParse();
```
jcameron, line 11: Strongly recommend against using `$in` in `index.cgi` scripts as they are not protected from XSS attacks

systemd/set_linger.cgi
```diff
@@ -0,0 +1,50 @@
...
+# This page is reached from the index linger toggle links.
+error_setup($text{'systemd_linger_err'});
+ReadParse();
+
+# Validate the requested user and linger state before calling loginctl.
+my $user = clean_unit_value($in{'user'});
+my $enabled = $in{'enabled'};
+get_user_details($user) || error($text{'systemd_euser'});
+systemd_can_linger(\%access, $user) || systemd_acl_error('plinger');
+if (!defined($enabled) || $enabled !~ /^[01]$/) {
```
jcameron: Better to use int() instead of a regexp to check in `$enabled` is a number or not

systemd/acl_security.pl
iliaross: I’d keep `qw(...)`. I know you’re probably asking because we don’t really use `qw(...)` much in other modules, but in this case it’s idiomatic Perl for fixed lists of simple ACL keys, and it keeps the code less noisy by avoiding repeated quotes and commas around every item.

systemd/acl_security.pl
iliaross: Agreed, good suggestion. I just added it. Thanks.

systemd/acl_security.pl
iliaross: See the same comment of mine https://github.com/webmin/webmin/pull/2760#discussion_r3413011310.

systemd/systemd-lib.pl
iliaross: In this repo it is only covered by tests, but it is intentionally exported for Virtualmin integration. I saw your suggestion https://github.com/virtualmin/virtualmin-gpl/pull/1234#pullrequestreview-4493938108 to improve its name. I agree—renamed here https://github.com/webmin/webmin/pull/2760/commits/9fb08dd8f04b832f7cda0339e0e02dca69816e55.

systemd/systemd-lib.pl
iliaross: It's better coding practice. Even though `%access` is available globally, I pass the `%access` hash explicitly so permission checks stay testable and reusable.

systemd/systemd-lib.pl
iliaross: No, not really. It’s a small normalizer, not just a hash lookup. It lets callers check an explicit ACL hash instead of relying on global `%access`, and it handles Webmin ACL values consistently:

```perl
return $acl->{$key} && $acl->{$key} ne '0' ? 1 : 0;
```

So it avoids scattered repeated checks, keeps this check in one place, makes tests easy to write, and avoids accidentally treating the string `"0"` as allowed...

systemd/index.cgi
iliaross: I am aware of it! And I agree that raw `$in` use on index pages is risky, but here it is needed only to restore tab/user context on return links. I have updated the comment https://github.com/webmin/webmin/pull/2760/commits/cbebe22a1459722ef6628d408caebdfe7727d463 to reflect that and that it is safe.

systemd/set_linger.cgi
iliaross: > Better to use int() instead of a regexp to check in `$enabled` is a number or not

Honestly, I’d keep the regexp here. This is not checking for "any number". It is validating a strict boolean request value to be only `0` or `1`.

`int()` would coerce bad input instead of rejecting it, e.g.:

```perl
int("abc")  # 0
int("1foo") # 1
```

That said, we can make this clearer without a regexp by explicitly allowing only string `0` or `1`, then normalizing it after validation if that's what you like more. See this https://github.com/webmin/webmin/pull/2760/commits/a1c88df428aa1e04ad2b0e3fa40e5a82dcfe208e patch.

systemd/acl_security.pl
jcameron: Yeah I know it's used a lot in Perl, so I suppose it's OK. Still an unusual language choice in my opinion though!

systemd/systemd-lib.pl
jcameron: Ok makes sense now, thanks!

systemd/systemd-lib.pl
jcameron: I really don't think it is better coding practice unless there is some situation where a different hash is used. And we don't do it this way anywhere else in Webmin.

systemd/systemd-lib.pl
jcameron: Again, when is a hash other than %access passed in?

systemd/index.cgi
jcameron: Ok sounds good

systemd/set_linger.cgi
jcameron: Yes that's cleaner, thanks! Regexps are nice, but I prefer not to use them when a simple equality comparison can be used.

systemd/systemd-lib.pl
iliaross: > I really don't think it is better coding practice unless there is some situation where a different hash is used. And we don't do it this way anywhere else in Webmin.

Yeah, we generally don’t do this elsewhere, but I think that’s mostly old Webmin style rather than a strong reason to keep doing it here.

And passing the ACL in makes these helpers easier to understand and test, because they don’t secretly depend on global `%access`. For example, `systemd_can_linger()` and _many_ other `systemd_can_*()` helpers are tested with custom ACL hashes.

That would be much messier if every helper always read `%access` directly...

systemd/acl_security.pl
iliaross: Cool, thanks!

systemd/systemd-lib.pl
iliaross: Quite a lot. `systemd_acl_bool()` is tested directly with custom hashes, and most of the `systemd_can_*()` helpers are tested that way too. That’s why the ACL is passed in, to let these functions be be as pure permission checks instead of having to mutate global `%access` for every case!

systemd/set_linger.cgi
iliaross: Yeah, and regexes are usually much slower, though mostly they’re much easier and shorter to write.

systemd/systemd-lib.pl
jcameron: Oh so the use case for this is just for writing tests of the various acl functions?

systemd/systemd-lib.pl
jcameron: Consistency across the codebase is important, as it reducing repetitive use of `%access` which is already guaranteed to be available by the Webmin API.

systemd/systemd-lib.pl
iliaross: > Oh so the use case for this is just for writing tests of the various acl functions?

Yes, this is the main reason.

systemd/systemd-lib.pl
iliaross: > Consistency across the codebase is important, as it reducing repetitive use of `%access` which is already guaranteed to be available by the Webmin API.

Yes, but such designed isn't great for tests or function portability.

iliaross:
I think we can merge it?

systemd/systemd-lib.pl
jcameron: OK then in that case, can we at least make the ACL parameter optional and it not set these various ACL functions can default to using `%access` . That way it doesn't have to be repeated in all the ACL function calls..

iliaross: Yes, if that's what you want https://github.com/webmin/webmin/pull/2760/commits/f2fe6c930f2eeed58c625386be34d2923e499ee0.

jcameron: Awesome, that is much cleaner!

systemd/systemd-lib.pl
jcameron: Actually you resolved this in the way I suggested in another comment, so we're good

PR #2761, opened 2026-06-12
Fix init module

iliaross:
Hi Jamie!

This PR cleans up the legacy Bootup and Shutdown module now that a dedicated Systemd Services and Units module exists.

It includes:

- Shows only configuration options relevant to the selected/detected boot system

- Keeps the legacy systemd view focused on basic `.service` units by avoiding exposing broader systemd unit types in the init module, since full systemd management belongs in the dedicated module

- Adds an optional config switch to hide the legacy systemd services table and show a link to the dedicated Systemd module instead if chosen or might be configured via Virtualmin Config

- Preserves existing Bootup and Shutdown naming and actions

This keeps the init module simpler and safer while giving admins a clear path to the new full systemd module when available.


jcameron:
Excellent thanks!

PR #2762, opened 2026-06-13
Don't fork. systemd prefers one process.

swelljoe:
systemd does not want services to fork.

New versions of systemd have a race condition when fork happens, because of more locked down PID checking. Restart loop and error is:

```
Jun 13 18:27:47 fqdn systemd[1]: Starting usermin.service - Usermin server daemon...
░░ Subject: A start job for unit usermin.service has begun execution
░░ Defined-By: systemd
░░ Support: https://wiki.almalinux.org/Help-and-Support
░░ 
░░ A start job for unit usermin.service has begun execution.
░░ 
░░ The job identifier is 303.
Jun 13 18:27:48 fqdn systemd[1]: usermin.service: Can't open PID file '/var/usermin/miniserv.pid' (yet?) after start: Permission denied
Jun 13 18:28:02 fqdn systemd[1]: usermin.service: start operation timed out. Terminating.
Jun 13 18:28:02 fqdn systemd[1]: usermin.service: Failed with result 'timeout'.
░░ Subject: Unit failed
░░ Defined-By: systemd
░░ Support: https://wiki.almalinux.org/Help-and-Support
░░ 
░░ The unit usermin.service has entered the 'failed' state with result 'timeout'.
Jun 13 18:28:02 fqdn systemd[1]: Failed to start usermin.service - Usermin server daemon.
░░ Subject: A start job for unit usermin.service has failed
░░ Defined-By: systemd
░░ Support: https://wiki.almalinux.org/Help-and-Support
░░ 
░░ A start job for unit usermin.service has finished with a failure.
```

There's no reason to do the fork dance, systemd handles all that (and can capture STDOUT to the journal, if we don't redirect it).

The changes and their reasons:

Type=simple - No longer a `forking` service. systemd directly controls the process.
ExecStop removed - systemd manages the process, it can stop it., we don't need `kill`.
Restart=on-failure - preferred over "always", which can cause a restart loop that can be destructive...if it's failing because memory is tight it can cause cascading failures or failures in other services, and repeated failure/start loops always consume resources. on-failure is the more correct behavior for a non-critical service.
TimeoutSec removed - No need for this. Never was a need for this to be custom.

### DIFF
--- a/webmin-systemd
+++ b/webmin-systemd
@@ -10,7 +10,7 @@ ExecStart=WEBMIN_LIBDIR/miniserv.pl --nofork WEBMIN_CONFIG/miniserv.conf
 ExecReload=WEBMIN_KILLCMD -HUP $MAINPID
 Restart=on-failure
 RestartSec=2s
-TimeoutStopSec=120s
+TimeoutStopSec=300s
 
 [Install]
 WantedBy=multi-user.target
--- a/webmin-systemd
+++ b/webmin-systemd
@@ -5,11 +5,10 @@ After=network.target network-online.target
 
 [Service]
 Environment="PERLLIB=WEBMIN_LIBDIR"
-ExecStart=WEBMIN_LIBDIR/miniserv.pl WEBMIN_CONFIG/miniserv.conf
+ExecStart=WEBMIN_LIBDIR/miniserv.pl --nofork WEBMIN_CONFIG/miniserv.conf
 ExecStop=WEBMIN_KILLCMD $MAINPID
 ExecReload=WEBMIN_KILLCMD -HUP $MAINPID
 PIDFile=WEBMIN_VAR/miniserv.pid
-Type=forking
 Restart=always
 RestartSec=2s
 TimeoutSec=15s
--- a/webmin-systemd
+++ b/webmin-systemd
@@ -6,13 +6,10 @@ After=network.target network-online.target
 [Service]
 Environment="PERLLIB=WEBMIN_LIBDIR"
 ExecStart=WEBMIN_LIBDIR/miniserv.pl --nofork WEBMIN_CONFIG/miniserv.conf
-ExecStop=WEBMIN_KILLCMD $MAINPID
 ExecReload=WEBMIN_KILLCMD -HUP $MAINPID
-PIDFile=WEBMIN_VAR/miniserv.pid
-Restart=always
+Restart=on-failure
 RestartSec=2s
-TimeoutSec=15s
-TimeoutStopSec=300s
+TimeoutStopSec=120s
 
 [Install]
 WantedBy=multi-user.target
--- a/webmin-systemd
+++ b/webmin-systemd
@@ -4,6 +4,7 @@ Wants=network-online.target
 After=network.target network-online.target
 
 [Service]
+Type=simple
 Environment="PERLLIB=WEBMIN_LIBDIR"
 ExecStart=WEBMIN_LIBDIR/miniserv.pl --nofork WEBMIN_CONFIG/miniserv.conf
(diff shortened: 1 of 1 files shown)


webmin-systemd
```diff
@@ -5,11 +5,10 @@ After=network.target network-online.target
 
 [Service]
 Environment="PERLLIB=WEBMIN_LIBDIR"
-ExecStart=WEBMIN_LIBDIR/miniserv.pl WEBMIN_CONFIG/miniserv.conf
+ExecStart=WEBMIN_LIBDIR/miniserv.pl --nofork WEBMIN_CONFIG/miniserv.conf
 ExecStop=WEBMIN_KILLCMD $MAINPID
 ExecReload=WEBMIN_KILLCMD -HUP $MAINPID
 PIDFile=WEBMIN_VAR/miniserv.pid
-Type=forking
```
iliaross, line 12: I wouldn't drop `Type` and made it explicitly set to `simple`.

webmin-systemd
```diff
@@ -5,11 +5,10 @@ After=network.target network-online.target
 
 [Service]
 Environment="PERLLIB=WEBMIN_LIBDIR"
-ExecStart=WEBMIN_LIBDIR/miniserv.pl WEBMIN_CONFIG/miniserv.conf
+ExecStart=WEBMIN_LIBDIR/miniserv.pl --nofork WEBMIN_CONFIG/miniserv.conf
 ExecStop=WEBMIN_KILLCMD $MAINPID
 ExecReload=WEBMIN_KILLCMD -HUP $MAINPID
 PIDFile=WEBMIN_VAR/miniserv.pid
```
iliaross: Do we still need `PIDFile` or we can drop it now?

iliaross:
Thanks, Joe! Overall, it looks good to me. I’ve left a few comments though.

webmin-systemd
swelljoe: Now that I'm awake, I realize we don't need a lot of stuff. But `Type=simple` isn't the right type, I don't think.

I'll make it nice.

iliaross: Thanks!

swelljoe: Oh, but, I guess the newer `exec` type doesn't work on old systemd versions which we have to keep supporting.

swelljoe:
OK, I've updated the PR description to explain all the changes.

One question for your @iliaross : Why do we have such a long TimeoutStopSec? 5 minutes feels crazy?

iliaross:
Small wording nit: systemd can capture stdout/stderr for foreground services, but miniserv.pl still redirects stdout to /dev/null and stderr to its configured error log after startup. The unit change still looks right; that part of the PR description just isn't quite accurate.

iliaross:
... though we will let Jamie decide on it.

iliaross:
> One question for your @iliaross : Why do we have such a long TimeoutStopSec? 5 minutes feels crazy?

To reiterate private chat: I initially set it to 5 minutes, but it could probably be longer to accommodate systems running large/lengthy backup jobs. In practice, it shutdowns quickly, yet 5 minutes might still interrupt a backup job, and 90 seconds is even more likely to do so...

swelljoe:
I've updated the PR description. I think we should, at some point, let it log to the journal rather than redirecting, unless explicitly configured not to (it could do both, as well, for a transition period). Everyone on modern systems is going to look there first for "why isn't this starting?" and it automatically gets ingested into whatever log aggregation tools the user is using. Logging to the journal is being a good citizen.

swelljoe:
So, yeah, we'll leave the stop timeout to 5 minutes for now. We can address the root cause of that problem independently.

PR #2763, opened 2026-06-15
xmlrpc.cgi strict, warnings, critic, tests, and security audit

swelljoe:
Beyond the obvious changes to make strict/warnings/perlcritic happy:

1. Replaced the precedence-sensitive ACL expression with webmin_user_can_rpc() — same decisions for every rpc value (verified 0/1/2/3/undef), but the policy now lives in one place instead of a hand-rolled &&/|| chain inside the security gate, and it's no longer fragile to a misread of operator precedence.
2. Hardened the body read against a missing CONTENT_LENGTH (fail-closed, no warning noise).
3. Dropped the now-unused $base_remote_user declaration.

Net effect on the security surface is small because the surface is small, but it's simpler/less code. The one change with real value is centralizing the authorization decision in the shared helper.

I'm not sure there are still any real users of XML-RPC with Webmin? But, we might as well have unit tests, and strict/warnings clean code, if it's going to stick around. It is a giant security landmine since RPC privilege is total control of the system, but any root user already has that, so it's kind of moot. If it is still widely used, I can add fine-grained ACLs and maybe privilege separation to a todo list.

### DIFF
--- a/t/xmlrpc.t
+++ b/t/xmlrpc.t
@@ -1,0 +1,216 @@
+#!/usr/bin/perl
+# Unit tests for xmlrpc.cgi helper subs.
+#
+# xmlrpc.cgi is loaded as a module; its top-level body (ACL check, reading
+# the request, dispatching the call, emitting the response) is skipped by
+# the `unless (caller) { ... }` guard, so requiring it only defines the
+# subs plus loads WebminCore.
+#
+# The four subs under test are the XML <-> Perl marshalling layer:
+#   encode_xml_value  - Perl scalar/hashref/arrayref -> XML-RPC <value> body
+#   parse_xml_value   - parsed XML <value> node      -> Perl scalar/ref
+#   find_xmls         - recursive element search over an XML::Parser tree
+#   make_error_xml    - faultCode/faultString -> <methodResponse> fault doc
+#
+# Assertions target the contract (type selection, round-trip identity,
+# structural balance, escaping), not exact whitespace or attribute order.
+
+use strict;
+use warnings;
+use Test::More;
+use File::Basename qw(dirname);
+use File::Spec;
+
+# WebminCore loads web-lib.pl / ui-lib.pl with relative `do`, so the repo
+# root must be the cwd (the script's BEGIN block adds "." to @INC).
+my $root = File::Spec->rel2abs(
+	File::Spec->catdir(dirname(__FILE__), '..'));
+chdir($root) or die "chdir $root: $!";
+
+my $script = File::Spec->catfile($root, 'xmlrpc.cgi');
+require $script;
+
+# XML::Parser is only needed to build the parsed-tree inputs for
+# parse_xml_value and the round-trip tests. Probe for it once.
+my $have_parser = eval { require XML::Parser; 1 };
+
+# Build the parsed-tree node that parse_xml_value expects from an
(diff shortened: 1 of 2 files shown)


iliaross:
> * and wrap top-level execution in `unless (caller)` to enable unit testing.

I forgot. Did we decide not to use `unless` and instead break the code into small, testable functions?

swelljoe:
We did not. And, there's no reason to demand a refactor before adding tests, strict, warnings, and perlcritic, and in fact, having tests makes it safer to refactor.

This is not a dangerous change, and this is a part of the code that isn't used by anyone I'm aware of (and it's kinda scary, since it bypasses all ACLs if the user can RPC).

iliaross:
Thanks for the provided explanation.

> This is not a dangerous change, and this is a part of the code that isn't used by anyone I'm aware of (and it's kinda scary, since it bypasses all ACLs if the user can RPC).

I will leave it to Jamie to review and decide.

iliaross:
> And, there's no reason to demand a refactor before adding tests, strict, warnings, and perlcritic, and in fact, having tests makes it safer to refactor.

To be clear, I think adding tests is absolutely fine. Yet I would have done the same and added tests first too. But once the tests are in place, and working, I’d refactor the code so the production path doesn’t feel shaped around the test harness...

jcameron:
I don't love the "unless caller" thing because it makes the code hard to read with that random trailing `}` that doesn't follow normal indents, but if it allows testing then that's an improvement. Long-term I'd prefer if we could split functions we want to test into a separate file though, or maybe just move them into web-lib-funcs.pl if they are suitably generic.

iliaross:
This PR should not have been merged as is. We’re approaching the problem from the wrong side, and, worse, it breaks working software. That’s very annoying.

`miniserv.pl` runs Perl CGIs with `do $file`, and inside `do`, `caller` is _true_. So the new `unless (caller)` guard skipped the whole `xmlrpc.cgi` request handler, printed no CGI headers, and made Miniserv log `Missing Headers`.

Please create new tests if you want to use `unless`, make sure they return the results you expect, then break the code into small testable functions and get rid of the `unless (caller)` mess.

I fixed it here https://github.com/webmin/webmin/commit/74fd0ca12dc55b282881160523cecbbc8f73a464. This is still ugly but at least it works again.

I suggest we revert that change for `unless (caller)` in `shellserver.pl` and `miniserv.pl` and do it properly. I didn't find yet that anything was broken in those two though but please let's not add new `unless (caller)`.

swelljoe:
I'm fine with subs in cgis becoming proper libraries or moved to `web-lib-funcs.pl` (though that feels heavy, but I guess it's always shared memory with all the other users of `web-lib-funcs.pl`, which is just about everybody). I don't care how it happens, but I'd like to see good unit test coverage, and strict/warnings/perlcritic so bigger refactoring is safer, fuzzing for security and profiling for performance can be done on individual units of functionality.

swelljoe:
But, I'll stop making these test/strict/warnings PRs for CGIs until we sort that out, I guess.

jcameron:
Oh yeah I'm strongly in favor of better tests!

Although in this case, I feel like we could drop xmlrpc.cgi entirely because I doubt it is used by anyone.

iliaross:
I’m afraid we cannot know that for sure.

PR #2766, opened 2026-06-16
Register SNI callback on all SSL contexts, not just the default

bob7123:
On servers with multiple domains sharing a dedicated IP, or where the server hostname resolves to an IP that has a different domain's ipcert entry, Webmin serves the wrong SSL certificate. The SNI hostname sent by the client is ignored for connections to dedicated IPs.

**Cause:** `setup_ssl_contexts()` in miniserv.pl registers `CTX_set_tlsext_servername_callback` only on the default (`*`) context. Per-IP contexts created from `ipcert` entries do not get the callback. When a client connects to a dedicated IP, miniserv uses the per-IP context directly, the SNI callback never fires, and the client receives whatever certificate is assigned to that IP regardless of the requested hostname.

**Fix:** Register the same SNI callback on every context in `%ssl_contexts`. The callback function is unchanged. Clients without SNI still receive the per-IP certificate (unchanged behavior). Clients with SNI get the correct certificate matched by hostname.

Related: https://github.com/virtualmin/virtualmin-gpl/pull/1229

If there are any further concerns, please do not hesitate to let me know.

### DIFF
--- a/miniserv.pl
+++ b/miniserv.pl
@@ -4827,18 +4827,18 @@ foreach my $ip (keys %ssl_contexts) {
 # Ref: https://github.com/virtualmin/virtualmin-gpl/pull/1229
 if (defined(&Net::SSLeay::CTX_set_tlsext_servername_callback)) {
 	foreach my $ctx_key (keys %ssl_contexts) {
-	Net::SSLeay::CTX_set_tlsext_servername_callback(
-	    $ssl_contexts{$ctx_key}->{'ctx'},
-	    sub {
-		my $ssl = shift;
-		my $h = Net::SSLeay::get_servername($ssl);
-		my $c = $ssl_contexts{$h} ||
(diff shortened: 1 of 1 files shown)


miniserv.pl
```diff
@@ -4823,10 +4823,12 @@ sub setup_ssl_contexts
 		}
 	}
 
-# Setup per-hostname SSL contexts on the main IP
+# Setup per-hostname SSL contexts on all IPs, not just the default
+# Ref: https://github.com/virtualmin/virtualmin-gpl/pull/1229
 if (defined(&Net::SSLeay::CTX_set_tlsext_servername_callback)) {
+	foreach my $ctx_key (keys %ssl_contexts) {
```
jcameron, line 4829: This is good, but can you indent the code inside this new foreach loop?

bob7123: Sure, no problem. My apologies for not catching it before sending it to you.

jcameron: thanks!

Issue #1607, opened 2022-03-06
No network interfaces shown on Pi OS 11 (Debian Bullseye)?


Jibun-no-Kage:
No network interfaces shown on Pi OS 11 (Debian Bullseye)?  Does WebMin rely on /etc/network/interfaces file?  If so, than any Debian 11 based distribution that has moved to using the dhcpcd service for dhcp IP Address and ethernet interface configuration specifics will not how network configuration in webmin correctly.  Default configuration for most Debian 11 based distributions is now defaulting to dhcpcd service use.  Webmin is blind to this apparently?

jcameron:
Webmin will use /etc/network/interfaces if it exists, or files in /etc/netplan otherwise.

Do you have either of those on Debian 11?

Jibun-no-Kage:
No, not by default.  So webmin shows active interfaces, but no interfaces as configured for startup, which is not true, if dhcpcd is active and that is the default for debian 11 (Pi OS 11 for example).  Debian has been moving to using dhcpcd service as core since 9, and standard in 10, now 11.

jcameron:
So on Debian 11, in which file would you configure a static IP address in? I assume that there's some config in which you specify that you want to use DHCP for an interface?

Jibun-no-Kage:
Look at the DHCPCD service manual information.  You can configure either DHCP or static under DHCPCD, as well as a failback to static if DHCP fails.  I could post examples, but figure you will want to review the detailed information yourself.

jcameron:
DHCPCD controls the assignment of addresses to other machines - what I was wondering is how on Debian 11 you select to use DHCP on an interface. But it sounds like this is set in files under  /etc/network/interfaces.d ?

Jibun-no-Kage:
"DHCP is NOT DHCPCD... note the DIFFERENCE.  DHCPCD is a service that provides reference to DEFINED INTERFACES, which maybe use DHCP based, STATIC or BOTH when a failback configuration is used.  Please DO NOT confuse DHCP with DHCPCD.  This is why I suggested reviewing the documentation of DHCPCD.

Webmin does not show any interfaces configured for startup when 'interfaces' or 'interfaces.d' or 'netplan' is not used for configuration of interfaces at startup.  There another source for network configuration, i.e. DHCPCD,  and so it can be a source for such information.  It is misleading to not show defined (at start) networking information in the webmin UI, because it currently ignores how DHCPCD can quality the start up configuration of interfaces/adapters, no?

[...]

PR #2769, opened 2026-06-19
Add dhcpcd network backend for Debian and Raspberry Pi OS

iliaross:
Hi Jamie,

This PR adds dhcpcd backend support for Debian and Raspberry Pi OS network configuration as requested in this https://github.com/webmin/webmin/issues/1607 feature report a while ago. It detects dhcpcd only as a final fallback after Netplan, NetworkManager, and ifupdown, preventing Webmin from incorrectly falling back to `/etc/network/interfaces` on dhcpcd-managed systems.

The new backend reads and writes `/etc/dhcpcd.conf`, including DHCP and static IPv4/IPv6 configuration, gateways, static routes, DNS servers, search domains, MTU, and virtual IPv4 aliases. It also supports implicit DHCP-managed interfaces for default dhcpcd setups with no explicit interface blocks, and handles `allowinterfaces` / `denyinterfaces` behavior.

This PR also fixes apply/delete flows for dhcpcd-managed interfaces and virtual aliases, avoids rewriting generated `/etc/resolv.conf`, preserves spacing/comments in touched hosts and nsswitch files, and tightens Active Now handling so virtual aliases are treated as IP addresses rather than independent links.


jcameron:
When you say "dhcpd" here, do you mean a DHCP server like the one Webmin already configures in the "dhcpd" module? Normally when interfaces are configured via DHCP, a separate machine on the network controls IP address, DNS and route configuration.

Also, for fixes unrelated to PiOS support, could you split them into a separate PR? That's a lot easier to review and merge..

iliaross:
> When you say "dhcpd" here, do you mean a DHCP server like the one Webmin already configures in the "dhcpd" module?

No, `dhcpcd` is not the same as `dhcp`. `dhcp` is the protocol. `dhcpcd` is a client program that manages network configuration.

> Also, for fixes unrelated to PiOS support, could you split them into a separate PR? That's a lot easier to review and merge..

Ah, yeah — sorry. I had meant to commit those directly at first, but I ended up getting pretty tired after working on these changes and submitting everything in this PR. On top of that, things weren’t behaving properly because of those existing bugs. In any case, I reverted it here https://github.com/webmin/webmin/pull/2769/commits/090aed60dfaab8aa7d6e87072e768f623bbf6793 and here https://github.com/webmin/webmin/pull/2769/commits/f2782073ce3e39415f98a636f9422c4b2fcc592d and opened a separate PR here https://github.com/webmin/webmin/pull/2770, with each commit having a clear, detailed explanation.

Remember, if you want to test it and have everything work flawlessly, you need to apply changes from both PRs.

jcameron:
Thanks for that separate PR ... I've approved it!

And I see now the difference between `dhcpcd` and `dhcpd` :-)

net/delete_bifcs.cgi
```diff
@@ -14,27 +14,40 @@ foreach $d (reverse(@d)) {
...
+		if ($in{'deleteapply'} && $act &&
+		    !defined(&unapply_interface_after_delete)) {
+			# De-activate first for legacy immediate-action backends.
+			if (defined(&unapply_interface)) {
+				$err = &unapply_interface($act);
+				$err && &error("<pre>$err</pre>");
+				}
+			else {
+				&deactivate_interface($act);
+				if(&iface_type($b->{'name'}) eq 'Bonded'){
```
jcameron: Add a space after the `if`

net/delete_bifcs.cgi
```diff
@@ -14,27 +14,40 @@ foreach $d (reverse(@d)) {
...
+					if (($gconfig{'os_type'} eq 'debian-linux') && ($gconfig{'os_version'} >= 5)) {}
```
jcameron: No need for brackets around `$gconfig{'os_type'} eq 'debian-linux'` since it's a simple comparison and there are no precedence issues

net/net-detect.pl
```diff
@@ -17,15 +21,128 @@ sub net_has_netplan_config
        -d $dir;
 }
 
+# net_has_ifupdown_config([interfaces-file], [&seen-files])
+# Returns true if ifupdown has any non-loopback iface stanzas
+sub net_has_ifupdown_config
```
jcameron, line 26: Why is this function called `net_has_ifupdown_config` but it's actually parsing a config file?

net/net-detect.pl
```diff
@@ -17,15 +21,128 @@ sub net_has_netplan_config
...
+			}
+		}
+	}
+close($fh);
+return 0;
+}
+
+# net_has_dhcpcd_config([dhcpcd-conf], [service-active])
+# Returns true if dhcpcd appears to own interface startup config
+sub net_has_dhcpcd_config
```
jcameron, line 75: Again, the actual parsing of the config file should be factored out, and any `has_` functions should just call that to evaluate the parsed config structure.

net/net-detect.pl
iliaross: This function checks whether `ifupdown` has real non-loopback interface config, not just whether the file exists.

And, it parses the file intentionally.

For ifupdown, `/etc/network/interfaces` often exists but contains only loopback or `source` lines, which should not make Webmin select the `ifupdown` backend...

net/net-detect.pl
iliaross: Same as here:
https://github.com/webmin/webmin/pull/2769#discussion_r3447255657

net/net-detect.pl
jcameron: What I mean is, wouldn't it better to have a separate function that parses that config file into an array, and then this function (and net_has_dhcpcd_config) could iterate over that array?

iliaross: > What I mean is, wouldn't it better to have a separate function that parses that config file into an array, and then this function (and net_has_dhcpcd_config) could iterate over that array?

That would be cleaner **if we needed to reuse parsed** ifupdown state, but here **detection only** needs a boolean answer.

A separate parser would add more API without being used elsewhere, I remember you didn't like it that much.

jcameron: Actually, we already have a parser for /etc/network/interfaces in debian-linux-lib.pl! So why not just use that?

net/net-detect.pl
jcameron: See my reply above

net/net-detect.pl
iliaross: `get_interface_defs()` lives inside `debian-linux-lib.pl`, which is the `ifupdown` backend, not a standalone parser.

Loading it from `net-detect.pl` would load backend functions before backend selection and mess up with later loading `dhcpcd` or `netplan` or `nm` libs.

Yet technically, we could clean things up and pull the `ifupdown` parser out into a standalone helper that both places use, though I don't think we shouldn't waste precious time on this!

jcameron: I would strongly recommend factoring this out into a separate library or maybe net-lib.pl , as the parsing of the interfaces file is pretty complex (for example include files), and currently we have two implementations that will fall out of sync with each other. But that can be done in a follow-up PR so we don't hold up this good work..

iliaross:
> I would strongly recommend factoring this out into a separate library or maybe net-lib.pl , as the parsing of the interfaces file is pretty complex (for example include files), and currently we have two implementations that will fall out of sync with each other. But that can be done in a follow-up PR so we don't hold up this good work..

Aright, fixed here https://github.com/webmin/webmin/commit/0dbb5d2f89ad38dc6ab69b8171d5e9ac8e0f8d5e. Thanks!

Issue #1827, opened 2023-01-21
Can Webmin configure key-based SSH authentication for existing user?


zinwelzl:
Hi.

I'm new Webmin user, so...

Can Webmin configure key-based SSH authentication for existing users or I need to do it in terminal?

Whatever I try, don't work.
Try all I can find online, still no public/private file generated.

I can use ssh-keygen and ssh-copy-id.
Can't find how to to it in Webmin.

jcameron:
No, there's no way in the UI to do this currently..

swelljoe:
This exists in Usermin, though, in the SSH Configuration module. I'm not sure where/how it would fit in Webmin, though? The Authorized Keys page allows you to add public keys for authentication.

<img width="1913" height="580" alt="Screenshot of Usermin SSH Configuration module" src="https://github.com/user-attachments/assets/d8066dbc-61a3-4e27-9969-5cab6ec50011" />

iliaross:
This sounds more like a feature request. We'll keep it open until it’s solved or closed.

iliaross:
Implemented here https://github.com/webmin/webmin/pull/2771.

PR #2771, opened 2026-06-21
Add editable SSH public keys for existing Unix users

iliaross:
Hey Jamie!

This PR adds SSH public key management to the Users and Groups edit flow for existing Unix users, as requested in this issue: https://github.com/webmin/webmin/issues/1827.

Webmin stores its managed key with a readable marker in `authorized_keys`, validates submitted public keys, preserves unrelated keys, supports rename/update/remove flows, and performs user `.ssh` file operations as the target Unix user.


useradmin/user-lib.pl
```diff
@@ -2853,4 +2853,224 @@ sub can_user_sudo_root
...
+sub validate_ssh_pubkey
+{
+my ($pubkey) = @_;
+
+# Callers pass empty string only when they mean "remove"; validation is for add.
+return $text{'sshkey_eempty'} if (!$pubkey);
+my $ssh_keygen = &has_command("ssh-keygen");
+if ($ssh_keygen) {
+	# Prefer OpenSSH's parser when available, because key formats evolve.
+	my $pubkeyfile = &transname("id.pub");
```
jcameron: Would it be safer to run ssh-keygen as the user who's key is being validated, in case there is a bug in ssh-keygen that could crash it when provided untrusted input and thus allow a root privilege escalation?

jcameron:
Looks really good, I just added one suggestion

useradmin/user-lib.pl
iliaross: > Would it be safer to run ssh-keygen as the user who's key is being validated, in case there is a bug in ssh-keygen that could crash it when provided untrusted input and thus allow a root privilege escalation?

Yes, absolutely! Fixed on the latest commit. Thanks!

PR #2772, opened 2026-06-22
Add support to proxy linked-server WebSockets

iliaross:
Howdy Jamie! 😉 

This PR adds general WebSocket proxying for linked Webmin servers, allowing modules such as `xterm` to work when opened through `servers/link.cgi`.

As requested in https://github.com/webmin/webmin/issues/1866.


iliaross:
Alright, it _must be_ fully production-ready!

I also fixed Authentic to support the Terminal module and dashboard live stats for proxied WebSockets.

Now everything is working correctly in both normal and `link.cgi` proxied modes!

iliaross:
I ran more tests and fixed one issue with cleaning up consumed linked websocket routes. That one was important.

Now it's all perfect! Jamie, please see!

jcameron:
Looks pretty good! Are you 100% sure about the security of this change though? Could it be abused in any way to open a websockets connection to a different destination by an untrusted xterm user?

iliaross:
> Looks pretty good!

Thanks! I'm glad you like it as it took a lot of effort.

> Are you 100% sure about the security of this change though? Could it be abused in any way to open a websockets connection to a different destination by an untrusted xterm user?

No, I don’t see a way for an untrusted `xterm` user to turn this into a WebSocket proxy to another destination. The destination is forced from the configured linked server entry and not from user-controlled page content.

PR #2775, opened 2026-06-24
Fix to recover from a stale PID file whose PID was reused

iliaross:
Hello Jamie!

This is the last PR needed to mark Webmin 2.650 as ready. It fixes a problem where, after an unclean exit, Webmin may leave `miniserv.pid` behind.

If the kernel later reuses that PID for an unrelated process, the startup guard only checked that the PID was alive and refused to start with “Webmin is already running”.

With systemd restart handling, this can leave Webmin permanently down until the PID file is manually removed.

This change verifies that the live PID actually belongs to `miniserv.pl` running the same config before treating it as an active Webmin instance.

On Linux, it reads `/proc/<pid>/cmdline`, checks the miniserv script, and compares the config file by inode rather than by string, so the same file is still recognised if reached through a different path (e.g. a symlinked config dir), while a genuinely different config such as Usermin's is told apart.

If the PID is confirmed unrelated, the stale PID file is removed and startup continues. If the process cannot be inspected, the previous conservative behavior is preserved.

Also hardens PID-file parsing with `chomp` and numeric validation, and adds tests for unrelated PID reuse, matching config, symlinked config, different miniserv config, and unreadable command-line fallback.


miniserv.pl
```diff
@@ -5182,6 +5191,56 @@ sub read_any_file
 return $rv;
 }
 
+# read_process_cmdline(pid)
+# Returns a process command line from /proc, or undef if unavailable.
+sub read_process_cmdline
```
swelljoe: I don't think I'd make a function for two simple lines of code, especially if it's got only one caller (other than test code).

Also, `/proc` is Linux-specific. Isn't there a general way to get current parent PID and just compare against that rather than reading files and comparing strings? That's extremely fragile, and probably has some kind of risks I'm not sure I fully understand...anybody can started a process that will match your string comparison.

getppid() gets the parent process PID (which we need in forking scenarios, but don't need in systemd systems...and I'm assuming all this stuff doesn't happen on a systemd system, since the Webmin process is managed directly with systemd and can't have this problem), so I think that would be the simple thing that works across all UNIXen? https://perldoc.perl.org/functions/getppid

iliaross: > I don't think I'd make a function for two simple lines of code, especially if it's got only one caller (other than test code).
> 
> Also, `/proc` is Linux-specific. Isn't there a general way to get current parent PID and just compare against that rather than reading files and comparing strings? That's extremely fragile, and probably has some kind of risks I'm not sure I fully understand...anybody can started a process that will match your string comparison.
> 
> getppid() gets the parent process PID (which we need in forking scenarios, but don't need in systemd systems...and I'm assuming all this stuff doesn't happen on a systemd system, since the Webmin process is managed directly with systemd and can't have this problem), so I think that would be the simple thing that works across all UNIXen? https://perldoc.perl.org/functions/getppid

`getppid()` unfortunately doesn’t help here. It only returns the parent PID of the currently running startup process, not the parent of the PID stored in `miniserv.pid`.

In the failing case, the PID file contains an old Webmin PID that now belongs to some unrelated process. A new `miniserv.pl --nofork ...` started by systemd has parent `systemd`, not the old daemon. In the forking case, the daemon child also gets reparented after the launcher exits. So comparing the pidfile PID to `getppid()` would either always fail or, worse, make us think a real already-running Webmin is stale and allow a double-start.

The `/proc/<pid>/cmdline` check is Linux-only, but intentionally so, as when `/proc` isn’t available, the helper returns `undef` and we keep the old conservative behavior. We only delete the pidfile when Linux gives us positive proof that the live PID is not miniserv for this config.

swelljoe:
OK, now that I understand the problem, I'm pretty sure we should just skip this check on a systemd system (when `--nofork` is used to start the process. There is no reason to check PIDs, no reason to compare process names, all that complexity is none of our business. systemd manages the process. If something was started on 10000 outside of systemd, that's the users fault, and systemd start will fail, because something is squatting on the port. But, there's no need for this check on a systemd system...so just skip it.

swelljoe:
The right thing is something like:

```
if (!$nofork_argv) {
  # Check vital config options
  &update_vital_config();

  # Check if already running via the PID file
  if (open(PIDFILE, $config{'pidfile'})) {
	  my $already = <PIDFILE>;
	  close(PIDFILE);
	  chop($already);
	  if ($already && $already != $$ && kill(0, $already)) {
		  die "Webmin is already running with PID $already\n";
		  }
	  }
}
```

On a systemd system, it manages the process. The only way to "break" this would be to manually start Webmin outside of systemd. Which is user error. If something is already on the port, it'll fail and log that. (Though it'll currently probably log it to the wrong place, since we redirect STDOUT/STDERR to the miniserv log, instead of letting it go to the journal.)

jcameron:
+1 to what Joe says, the simple fix would be to simply skip this check in `nofork` mode.

iliaross:
Alright, agreed! After looking into it more closely and running all the tests, I agree that simply skipping the PID-file guard in no-fork mode is a much better fix.

Thanks!

PR #2776, opened 2026-06-26
Fix SELinux labeling for Webmin runtime data

iliaross:
Hello Jamie,

This PR adds SELinux context handling for Webmin’s default runtime directory on EL systems.

Setup now persists and applies a `var_run_t` fcontext rule when SELinux is enabled, falls back safely when SELinux tooling is unavailable, and removes the local fcontext rule when `/var/webmin` is deleted during uninstall.

Discussed originally here:
https://forum.virtualmin.com/t/usermin-needs-selinux-fix/137474

### DIFF
--- a/makerpm.pl
+++ b/makerpm.pl
@@ -316,7 +316,7 @@ if [ "\$1" = 0 ]; then
 		# has taken it's place. Rename away the /etc/webmin directory
 		rm -rf /etc/webmin.rpmsave
 		mv /etc/webmin /etc/webmin.rpmsave
-		if command -v semanage >/dev/null 2>&1; then
+		if type semanage >/dev/null 2>&1; then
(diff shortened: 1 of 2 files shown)


makerpm.pl
```diff
@@ -316,6 +316,9 @@
 		# has taken it's place. Rename away the /etc/webmin directory
 		rm -rf /etc/webmin.rpmsave
 		mv /etc/webmin /etc/webmin.rpmsave
+		if command -v semanage >/dev/null 2>&1; then
```
jcameron: Same question as on the other PR ... I don't think `command` is available on non-bash systems

iliaross: Same answer—using `command -v` is best and cross-compatible across shells.

iliaross:
Is this one good too now?

jcameron:
Yes thanks!

PR #2777, opened 2026-06-26
net/linux-lib.pl: Fix bonding interface activation and use escape_shellarg

macmon:
## Problem

On systems where the `ip` command path is taken in `activate_interface()` (which is now the default since `ip` is prioritized over `ifconfig`), activating a bonding interface fails with:

    Cannot find device "bond0"

This happens because the code attempts `ip addr add ... dev bond0` without first creating the bond device.

Additionally, `quotemeta()` is used for shell argument quoting, which escapes dots in IP addresses unnecessarily. While this doesn't break functionality (the shell strips the backslashes), `escape_shellarg()` is the correct function and produces cleaner log output.

## Fix

1. Create the bonding device (`ip link add ... type bond`) with proper mode/miimon/slave configuration before assigning the IP address.
2. Replace `quotemeta()` with `escape_shellarg()` in the `ip` command path.

## Testing

Tested on Debian 12 (bookworm) with bonding interface creation via the Network Configuration module.

### DIFF
--- a/net/linux-lib.pl
+++ b/net/linux-lib.pl
@@ -276,14 +276,44 @@ if (!&has_command("ifconfig") && &has_command("ip")) {
 
 my $cmd;
 if (&has_command("ip")) {
+	# Create bonding device if it doesn't exist yet
+	if ($a->{'bond'} && !$old) {
+		my $out = &backquote_logged("ip link add ".&escape_shellarg($a->{'name'})." type bond 2>&1");
+		&error("Failed to create bond device : ".&html_escape($out)) if ($?);
+		if ($a->{'mode'}) {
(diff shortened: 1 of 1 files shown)


iliaross:
This PR targets the right bug, but it isn’t good as-is. It replaces `quotemeta()` with a non-existent `escape_shellarg()` helper, misses some bond options like primary, and doesn't address the legacy `load_module()` or `ip link` add double-creation risk. It also lacks regression coverage.

I have fixed it properly in https://github.com/webmin/webmin/commit/81d44f849180976d13692d48acda4af26c9ede91 by keeping Webmin's existing quoting style, creating the bond with `ip link` before assigning its address, mapping Webmin's numeric bond modes, attaching partner interfaces, applying primary, skipping legacy module creation when `ip` command is available, and adding a test for the live activation command sequence.

PR #2778, opened 2026-06-27
Fix to tighten privileged ACL checks

iliaross:
Hi Jamie!

This PR addresses privately reported issue and tries to restricts root user mutations and Custom Commands definition editing to full, non-safe Webmin admins.

This closes privilege-escalation paths for restricted operators while preserving execution of pre-approved Custom Commands under existing cmds ACLs.

### DIFF
--- a/custom/cgi_args.pl
+++ b/custom/cgi_args.pl
@@ -7,9 +7,9 @@ my ($cgi) = @_;
 my @cust = grep { &can_run_command($_) } &list_commands();
 if ($cgi eq 'edit_cmd.cgi') {
 	# Custom command editor
+	return 'none' if (!&custom_can_edit_commands());
 	my ($cmd) = grep { !$_->{'edit'} && !$_->{'sql'} } @cust;
-	return $cmd ? 'id='.&urlize($cmd->{'id'}) :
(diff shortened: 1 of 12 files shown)


jcameron:
What problem or vulnerability is being fixed here? The Custom Commands module already has an ACL setting that controls if commands can be edited or not, so I don't see why it needs to check permissions on other modules..

jcameron:
Oh I see the email now.

This is the wrong approach, and in my opinion the original security "vulnerability" report is invalid because Webmin module ACLs always grant full access by default. The only exception is when a module has a `safeacl` file, which we certainly could add for the Custom Commands module.

iliaross:
> What problem or vulnerability is being fixed here? The Custom Commands module already has an ACL setting that controls if commands can be edited or not, so I don't see why it needs to check permissions on other modules..
> > The only exception is when a module has a `safeacl` file, which we certainly could add for the Custom Commands module.

Ah, right. I totally forgot that neither `custom` nor `useradmin` had `safeacl`. I'll close it for now then.

jcameron:
FYI I added it for the Custom Commands module here : https://github.com/webmin/webmin/commit/1c9d57f5dd9b4b791e493bfd9aa1bad39f37603c

PR #2783, opened 2026-07-05
Fix to extract miniserv helpers into library

iliaross:
Howdy, Jamie!

This PR moves Miniserv helper functions into `miniserv-lib.pl`, removes the caller guard added in https://github.com/webmin/webmin/pull/2695 from `miniserv.pl`, and keep tests loading the helper library directly while preserving daemon startup behavior.


miniserv.pl
```diff
@@ -1,26 +1,13 @@
...
-@weekday = ( "Sun", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat" );
-@month = ( "Jan", "Feb", "Mar", "Apr", "May", "Jun",
-	   "Jul", "Aug", "Sep", "Oct", "Nov", "Dec" );
-
-# Only run the server logic when invoked as a script. When loaded as a
-# module (e.g. for unit tests) all subs below remain callable, but none
-# of the config-driven setup, socket binding, forking or accept loop
-# fires.
-unless (caller) {
+use FindBin;
```
jcameron, line 6: Why is FindBin needed here? Can we do whatever it does without this dependency?

miniserv.pl
```diff
@@ -1,26 +1,13 @@
...
+use lib $FindBin::Bin;
+BEGIN {
```
jcameron, line 8: Is there a reason for the BEGIN block?

miniserv.pl
iliaross: I used `FindBin` so that `miniserv.pl` can load `miniserv-lib.pl` relative to the location of `miniserv.pl`, not relative to whatever the current working directory happens to be.

That matters here because `miniserv-lib.pl` is required in a `BEGIN` block. We need it loaded before the rest of `miniserv.pl` is compiled, so constants imported by `Socket` are available correctly.

We could manually derive the script directory from `$0`, but that would basically reimplement what `FindBin` already does, and usually less reliably.

Important to notice that `FindBin` is also a core Perl module, so this is **not adding an external dependency**.

miniserv.pl
iliaross: Yes, see previous comment https://github.com/webmin/webmin/pull/2783#discussion_r3525306981.

miniserv.pl
jcameron: Personally I'd prefer use of $0 here, which is what we do elsewhere in Webmin and Virtualmin. That avoids the need to load and depend on a module that may not be available in all Perl versions.

iliaross: We should rely on Perl’s own docs here. `FindBin` is not an external dependency. It’s a *core* Perl module. The official perldoc says it is "supported as part of the core perl distribution", and `Module::CoreList` shows it has been present since Perl `5.00307` and not removed.

So I’d rather use the core library here instead of re-inventing the same path logic with `$0`. Existing `$0` usage can be cleaned up separately where it makes sense.

If `FindBin` is missing, that would mean a deliberately crippled custom Perl install, not a normal supported Perl distribution.

jcameron: Ok, sounds reasonable. Maybe in future we could use this elsewhere, like in Virtualmin API scripts?

iliaross: > Maybe in future we could use this elsewhere, like in Virtualmin API scripts?

Yes, it would be a good cleanup! Thanks!

PR #2785, opened 2026-07-09
net/linux-lib.pl: Fix VLAN interface activation with ip command

macmon:
## Problem

On systems where the `ip` command path is taken in `activate_interface()` (which is now the default since `ip` is prioritized over `ifconfig`), activating a VLAN interface fails with:

    Cannot find device "ens33.10"

This happens because the code attempts `ip addr add ... dev ens33.10` without first creating the VLAN device. The old `vconfig` path is skipped on Debian >= 5 (since it was expected to be handled by `ifup`), but `ifup` is never reached because the `ip` path comes first.

Additionally, the `ip link set ... up` block uses `$a->{'name'}` which is set to `"auto"` for VLAN interfaces, causing a second error.

## Fix

1. Create the VLAN device with `ip link add link <physical> name <physical>.<vlanid> type vlan id <vlanid>` when the `ip` command is available, before any address assignment. Falls back to `vconfig` on older systems without `ip`.
2. Resolve the correct device name (`physical.vlanid`) for VLAN interfaces in the link up/down block.

## Testing

- Added regression test for VLAN interface activation
- All 66 tests pass
- Tested on Debian 12 (bookworm) with VLAN interface creation via the Network Configuration module

Ref.: #2777

### DIFF
--- a/net/linux-lib.pl
+++ b/net/linux-lib.pl
@@ -242,9 +242,16 @@ sub activate_interface
 my ($a) = @_;
 my ($old) = grep { $_->{'fullname'} eq $a->{'fullname'} } &active_interfaces();
 
-# For Debian 5.0+ the "vconfig add" command is deprecated, this is handled
-# by ifup.
-if(($a->{'vlan'} == 1) && !(($gconfig{'os_type'} eq 'debian-linux') && ($gconfig{'os_version'} >= 5))) {
+# Create VLAN device with ip link if available (analogous to bond creation)
+if (&has_command("ip") && $a->{'vlan'} && $a->{'up'} && !$old) {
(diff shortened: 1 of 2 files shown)


jcameron:
@iliaross can you take a look at this one and merge it if it looks good? I believe you've been doing some work in this module recently..

iliaross:
The diagnosis and direction are right, but the implementation is incorrect. It works around the main bug. And, the patch as-is only works for the fresh-create flow, and the old guard can't work reliably.

I'm closing this PR, as this issue has been addressed in https://github.com/webmin/webmin/commit/77610666 commit.

@macmon Thanks for the report and the diagnosis! I'd appreciate it if you could apply the patch and give it a try to see if that also works for you:

```
webmin patch https://github.com/webmin/webmin/commit/77610666
```

PR #2789, opened 2026-07-12
Add ACL controls for URL download destinations

iliaross:
Hello Jamie!

This PR blocks non-public HTTP, HTTPS and FTP destinations by default in File Manager, Mailboxes and Upload and Download, with per-user IP/CIDR exceptions, redirect and proxy protection, cache isolation, and current-ACL enforcement for scheduled jobs.


filemin/acl_security.pl
```diff
@@ -31,6 +31,23 @@ sub acl_security_form
 		"style='width: 100%'"),
 	2);
 
+# Destinations allowed for URL downloads
+my $download_address_mode = $access->{'download_address_mode'} || 'public';
```
jcameron: Why not just set download_address_mode=public in defaultacl ?

iliaross: I actually do that too.

filemin/http_download.cgi
```diff
@@ -46,7 +46,10 @@ else {
 				$full, undef,
 				\&progress_callback,
 				$ssl, $in{'username'},
-				$in{'password'});
+				$in{'password'}, undef, undef,
+				undef, undef, undef,
+				$access{'download_address_mode'} || 'public',
+				$access{'download_allowed_addresses'});
```
jcameron: Checking allowed addresses in http_download is the wrong place... it's an internal API that should just fetch URLs, not perform higher-level access checks.

mailboxes/acl_security.pl
```diff
@@ -58,6 +58,23 @@ sub acl_security_form
 
 print &ui_table_row($text{'acl_candetach'},
 	&ui_yesno_radio("candetach", $o->{'candetach'}));
+
+# Destinations allowed for remote message content
+my $download_address_mode = $o->{'download_address_mode'} || 'public';
```
jcameron: Same here about defaultacl

filemin/acl_security.pl
jcameron: Oh so in that case, this line should be just : 

```
my $download_address_mode = $access->{'download_address_mode'};
```

filemin/http_download.cgi
iliaross: The ACL check itself stays in the module. `http_download` doesn’t read or interpret ACLs, and it only enforces an optional address policy supplied by the caller, so redirects, DNS resolution, and proxy connections are checked too. Existing callers that don’t pass a policy behave exactly as before.

Some checks can be done in the caller, but that would only validate the initial URL. `http_download` may then follow redirects, resolve the hostname again, or send the request through a proxy—none of which the caller can reliably validate.

filemin/acl_security.pl
iliaross: > Oh so in that case, this line should be just :

Agreed. Done here https://github.com/webmin/webmin/pull/2789/commits/d5b2a174c2dd849d6c05b7e8714899079373f43d.

jcameron: thanks!

filemin/http_download.cgi
jcameron: Couldn't we use the existing callback functionality in http_download for these checks? That seems cleaner than adding new special-purpose parameters..

iliaross: Nah, the existing callback is currently progress-only and runs too late for destination validation.

We need the check at the socket layer to prevent DNS rebinding. Yet, I agree the additional positional parameters are awkward, though.

...I've replaced it with  refactored download address checks to use callback bundles https://github.com/webmin/webmin/pull/2789/commits/1555224592a6ed4b9ed42adf40136b6261971e93.

jcameron: Couldn't you just check the original URL before calling `http_download`, and also if there is a redirect the callback function is called with the parameters `5` and the new URL here : https://github.com/webmin/webmin/blob/master/web-lib-funcs.pl#L3349

That should be enough to avoid the need to add callback bundles in https://github.com/webmin/webmin/commit/1555224592a6ed4b9ed42adf40136b6261971e93

iliaross: > Couldn't you just check the original URL before calling `http_download`

Callback event `5` is only a progress notification. Its return value is ignored, and the `Location` value may be relative.

More importantly, checking the hostname before the download is not enough because its IP address could change before connecting. The address callback checks the **actual IP** used for the connection and repeats the check after every redirect. That is why the check belongs inside the download code...

jcameron: The callback could call `error` though, right?

As for the IP, how could that change? Are you thinking that two DNS lookups for the same hostname in the same script could return different results?

iliaross: > Are you thinking that two DNS lookups for the same hostname in the same script could return different results?

I'm thinking of DNS rebinding, like the same hostname returns different IP addresses across two DNS lookups.

Simple example:

1. Attacker controls `evil.example`.
2. Webmin checks it:
   `evil.example ⇾ 8.8.8.8`
   This is public, so Webmin allows it.
3. A moment later, Webmin opens the connection and resolves it again:
   `evil.example ⇾ 169.254.169.254`
4. Webmin now connects to AWS metadata, even though it approved `8.8.8.8`.

So this is unsafe.

And, that is why I think the validation belongs closer to `open_socket()`, which checks the same resolved addresses that are actually used for the connection...

iliaross: > The callback could call `error` though, right?

...yes, but it would still check before the hostname is resolved again for the connection, leaving the DNS-rebinding gap.

jcameron: Check out https://github.com/webmin/webmin/commit/70011c11c0f5b96b887f43ec5e29cd360f39c0d6 , which makes the remote IP available to the callback function.

FYI while writing this I realized that for requests that go via a proxy, we are still vulnerable to DNS rebinding because we have no visibility into what IP the proxy selected!

iliaross: Thanks! I incorporated your `70011c1` and made it work with my callback bundle in this PR.

To be clear, for proxy requests, the destination is now resolved and validated locally first. Webmin sends the validated numeric IP in the HTTP URL or HTTPS `CONNECT` request while preserving the original hostname in the `Host` header and TLS SNI. And, FTP proxy downloads work the same way.

So the proxy no longer performs destination DNS resolution, closing the rebinding gap.

Mode 7 may still report an undefined IP through a proxy, but the security check does not depend on it. Redirects and direct proxy fallback are also revalidated.

Please check it again.

jcameron: Why is a separate address_callback still needed, when I made changes to support passing the existing callback function mode 7 and the destination IP? 

In the proxy case, I am not sure if we can safely resolve the hostname to an IP first and then pass that to the proxy in the URL, because the proxy may need to know the original hostname. 

Overall I think this PR is getting too complex to be a single change as it includes both some tricky core API changes and ACL UI additions. In fact, the changes to acl_security.pl probably shouldn't be specific to individual modules. Instead they could go into the top-level acl_security.pl file and be referenced by all modules that do downloads. That could do as a separate PR that I could approve quickly, and the actual implementation to use the list of allowed IPs could come later.

For FTP downloads, I realized I forgot to add the callback after the remote connection, so I added it here : https://github.com/webmin/webmin/commit/ed3cd51e40d689d1f000961a637b0696678650c8

iliaross: Alright, you know what? I agree, only the other way around. Let's just make more complex core API changes later and now simply add URL download destinations control. And, I also agree that making a global ACL is better than having each module handle it. All fixed in the latest https://github.com/webmin/webmin/pull/2789/commits/99f35fa38ac50b5af17b96dd7f9021a8a2960dac patch!

iliaross:
I think this is good to merge now?

filemin/http_download.cgi
jcameron: Sure sure, splitting it up that way also makes sense. I'll check out the latest commit..

jcameron:
Looks good now, thanks!!

jcameron:
Actually just noticed that this change would block root by default, so I committed https://github.com/webmin/webmin/commit/e805319233c5a6837e97b35a329c6bb2cad82657

iliaross:
> Actually just noticed that this change would block root by default, so I committed [e805319](https://github.com/webmin/webmin/commit/e805319233c5a6837e97b35a329c6bb2cad82657)

I don't think we should do it. The default should be equally safe for any user.

jcameron:
There's no point blocking what the root user can do, they have full permissions to run `curl` to fetch any URL anyway. 

But be we should limit untrusted users by default, hence https://github.com/virtualmin/virtualmin-gpl/commit/1640a43aa3e5458c44da7019a201351a7a5cd0ae

iliaross:
Alright, agreed. And this is a nice way to install new ACLs for all untrusted users on the system.

And by the way, I think `set_user_modules` needs a proper description.

Thanks!

jcameron:
> And by the way, I think set_user_modules needs a proper description.

Good idea, will add one..

Issue #2696, opened 2026-05-11
My webmin terminal broke after the update.


Bazsi-2006:
I had issue with my webmin terminal, that is behind a caddy reverse proxy, after the latest update. I was able to fix this, by adding the flowing lines to the miniserv.conf:
redirect_ssl=1
redirect_host=mydomain.example
redirect_port=443

I thought I would mention it, if someone else is experiencing this issue.

iliaross:
Hello,

Thanks for the heads up!

I'm glad to hear you managed to fix it! Though, what exact error did you see after the upgrade? How did it fail to work?

Bazsi-2006:
On Webmin’s terminal, I got a WebSocket connection error when I tried to access it from my domain behind a caddy reverse proxy. When I accessed it from my local network using the port directly, it worked fine.

I didn’t modify any configuration files on my proxy or in Webmin, I only updated the package. If the issue only happens in my setup, maybe I misconfigured something when I originally set it up.

Anyway, I just thought I’d leave a note in case someone else runs into the same issue as me.

iliaross:
This can be closed as resolved.

Issue #2329, opened 2024-12-01
nginx reverse proxy (ssl enabled) subdirectory installation terminal does not launch


truckerboy:
Operating system Ubuntu Linux 22.04.5
Webmin version 2.202 
Every time I try connecting to the terminal via the webmin GUI, I get the following websockets connection error in Chrome console:

`bundle.min.js?2120700009999999999:177 Refused to connect to 'ws://<my_domain>/webmin/xterm/ws-555' because it violates the following Content Security Policy directive: "default-src https: data: 'unsafe-inline' 'unsafe-eval'". Note that 'connect-src' was not explicitly set, so 'default-src' is used as a fallback.`

my nginx subdirectory config for webmin (ssl directives are in a separate file which are included in sites-enabled folder):

```
# Running webmin as a subfolder to an existing virtual host

location = /webmin {
    return 301 /webmin/;
}

location ~ /webmin/(?<ndpath>.*) {
    proxy_set_header Host $host;
    proxy_set_header X-Real-IP $remote_addr;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    proxy_http_version 1.1;
    proxy_set_header   Upgrade $http_upgrade;
    proxy_set_header   Connection "Upgrade";
    proxy_pass http://127.0.0.1:10000/$ndpath$is_args$args;
}
```
/etc/webmin/miniserv.conf

```
redirect_ssl=1
allow=127.0.0.1
cookiepath=/webmin
redirect_port=443
ssl=0

```
/etc/webmin/config

```
relative_redir=0
webprefix=/webmin
webprefixnoredir=1
referers=<my_domain>

```

/usr/share/webmin/miniserv.pl has the following additional code to avoid redirects to root folder when logging in.

```
# If the cookiepath exists presume we've got a prefix proxied config and change redirect base to include the cookiepath
$hostport = $config{'cookiepath'} ? $hostport.$config{'cookiepath'} : $hostport;

```

iliajie:
Hello,

Have you checked our FAQ on configuring Webmin behind a proxy?

https://webmin.com/faq/#can-i-run-webmin-or-usermin-behind-reverse-proxy

truckerboy:
yes I have. as I said, it seems to be working perfectly for me except when I try to launch the terminal where I am getting the websocket error. U have posted my nginx config file for your reference and as you can I see I have followed your instructions for enabling websockets but it refuses to work. Is there any additional configuration required in miniserv.conf?

[...]

Issue #2144, opened 2024-05-03
web terminal not working behind ha reverse proxy


kindlydodo:
webmin version: 2.111

i click on the left bar > Terminal icon, the popup is opening, and it displays "CONNECTING..." but i never get to the terminal.

This are the Browser (Chromium) logs:

```
Mixed Content: The page at 'https://webmin.site.com/sysinfo.cgi?xnavigation=1' was loaded over HTTPS, but attempted to connect to the insecure WebSocket endpoint 'ws://webmin.site.com/xterm/ws-563'. This request has been blocked; this endpoint must be available over WSS.
```

```
Uncaught DOMException: Failed to construct 'WebSocket': An insecure WebSocket connection may not be initiated from a page loaded over HTTPS.
    at https://webmin.site.com/unauthenticated/js/bundle.min.js?2110001019999999999:178:81906
```

tmp log file:
```
Running shell '/bin/bash -bash' with pid 8940
listening on port 564
timeout waiting for connection at /usr/share/webmin/xterm/shellserver.pl line 103.
```

iliajie:
Hello,

The default socket protocol in Webmin is `wss`. If you ran Webmin without SSL (which is not recommended, and what you don't do), then you should have changed the **Hostname for WebSocket connections** option to `ws://your-host:10000` on the **Terminal → Configuration** page. The same true visa versa or in case of using revers proxy.

Simply saying make connection to Webmin using `https` and use `wss` for socket connection, or configure it otherwise in module config.

Nevertheless, you should always avoid mixed content, and ensure that your HA Proxy is configured to proxy connections to `wss` instead.

kindlydodo:
Thank you!

As far as i change to manual and set the value : `wss://webmin.site.com` it works

shoulders:
@kindlydodo if all is working, can you close the issue 😄

karwrkinga1:
Where do I go to update this WS://webmin.site.com?

harifhaqim:
Hi, I'm also having the same issue with Cloudflare Tunnel...I can't find "Hostname for WebSocket connections" in Terminal Configuration...how could I solve it?

iliajie:
> Hi, I'm also having the same issue with Cloudflare Tunnel...I can't find "Hostname for WebSocket connections" in Terminal Configuration...how could I solve it?

Thanks for the heads up! It was moved to "Webmin ⇾ Webmin Configuration: Ports and Addresses" page.

iliaross:
[...]

PR #2794, opened 2026-07-15
Fix to honor redirect settings for WebSocket URLs

LynYugiri:
WebSocket URLs currently ignore `redirect_ssl`, `redirect_host`, and
`redirect_port`, even though these settings define Webmin's canonical
browser-visible URL and are already used for redirects and WebSocket
origin validation.

This change makes automatic WebSocket URLs inherit the configured
redirect scheme, host, and port. An explicit `websocket_host` or
caller-provided host continues to take precedence, and existing proxy
header and direct-request fallbacks remain unchanged when no redirect
settings are configured.

This prevents insecure `ws://` URLs from being generated when Webmin is
served over HTTPS through a reverse proxy using an HTTP upstream.

Related: #2696, #2329, #2144

### DIFF
--- a/web-lib-funcs.pl
+++ b/web-lib-funcs.pl
@@ -14765,9 +14765,13 @@ my $webprefix = &get_webprefix();
 my $trust_proxy = $miniserv{'trust_real_ip'};
 my $default_ws_proto = lc($ENV{'HTTPS'}) eq 'on' ? 'wss' : 'ws';
 my $ws_proto;
+# Match the canonical redirect scheme when one has been configured.
+if ($miniserv{'redirect_ssl'} ne '') {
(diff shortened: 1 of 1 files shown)


iliaross:
Thanks!

I can see now that we'll also need to handle the `redirect_port` case without `redirect_host`. I'll fix it in the follow-up commit.

iliaross:
Quick update.

While working on the `redirect_port` follow-up, I found two more issues with linked-server WebSockets.

The child was applying its public redirect settings before returning the URL, which could prevent the parent from recognizing it. The parent could then use `redirect_host` for its local WebSocket URL instead of the hostname the browser was actually using.

I’ve fixed both cases along with `redirect_port` handling in this https://github.com/webmin/webmin/commit/0eaf0ef80d3eddcd3cf05d480942d3dcad4b9392 patch.

PR #2795, opened 2026-07-16
MM-12032 Fix bond/VLAN/bridge not removed from active list on delete

macmon:
This PR fixes multiple related issues found while working on bonding
interfaces:

**1. Virtual devices not removed on delete**

When deleting a bond, VLAN or bridge interface via "Delete and Apply",
the interface disappears from the boot configuration but is still
showing in the "Active Now" list. Only a reboot makes it go away.

The reason is that delete_bifcs.cgi and save_bifc.cgi only called
deactivate_interface() which does ip addr del + ip link set down.
That brings the link down but the kernel keeps virtual devices until
they are removed with ip link delete.

Fix: Added destroy_interface_device() in linux-lib.pl that calls
ip link delete. It is called from both CGIs after deactivation.
Uses defined(&destroy_interface_device) as guard so other OS
backends are not affected.

**2. Bridge interfaces not created during activation**

Creating a bridge interface and clicking "Save and Apply" fails with:
"Failed to bring up link: Cannot find device br0". The bridge device
does not exist in the kernel because nobody creates it.

Unlike bond (81d44f8) and VLAN (7761066) interfaces, bridges had no
ip link add type bridge in activate_interface().

Fix: Added bridge creation block in activate_interface() following
the same pattern as bond and VLAN.

**3. Bonding options causing trailing whitespace**

After creating a bonding interface with miimon, updelay or downdelay
values, every save adds more spaces to these values in
/etc/network/interfaces. After a few saves the values are broken and
"Save and Apply" does not work anymore.

The cause is that bonding options were pushed as single-element arrays:
  push(@options, ['bond-miimon 100']);

The write loop uses "\t$param $value\n" which adds a space after the
value because $value is empty.

Fix: Changed to key-value pairs:
  push(@options, ['bond-miimon', '100']);

**4. Interfaces getting IP but staying DOWN**

After deactivating a normal ethernet interface and then activating it
again, the interface gets its IP address back but stays DOWN. It is
not reachable on the network.

The ip link set up/down block in activate_interface() was guarded by:
  if (($a->{'bond'} || $a->{'vlan'} || !&has_command("ifconfig")) && ...)

But ip addr add runs always when ip is available - the elsif path
with ifconfig/ifup is never used. So the address was set but the link
was never brought up.

Fix: Changed condition to if (&has_command("ip")) since the ip command
is used for address assignment anyway.

**Testing:**

Tested on Debian 12 with bonding (mode 0, 1), VLAN (802.1Q) and
bridge interfaces. Create + Save and Apply, Deactivate, Activate,
Delete and Apply all work. Unit tests added in net/t/run-tests.t.

### DIFF
--- a/net/debian-linux-lib.pl
+++ b/net/debian-linux-lib.pl
@@ -230,12 +230,12 @@ if ($cfg->{'bridge'}) {
 
 # Set bonding parameters
 if(($cfg->{'bond'} == 1) && ($gconfig{'os_version'} >= 5)) {
-	push(@options, [&bonding_option('mode').' '.$cfg->{'mode'}]);
-	push(@options, [&bonding_option('miimon').' '.$cfg->{'miimon'}]) if ($cfg->{'miimon'});
-	push(@options, [&bonding_option('updelay').' '.$cfg->{'updelay'}]) if ($cfg->{'updelay'});
-	push(@options, [&bonding_option('downdelay').' '.$cfg->{'downdelay'}]) if ($cfg->{'downdelay'});
-	push(@options, [&bonding_option('primary').' '.$cfg->{'primary'}]) if ($cfg->{'primary'});
-	push(@options, ['slaves '.$cfg->{'partner'}]);
+	push(@options, [&bonding_option('mode'), $cfg->{'mode'}]);
+	push(@options, [&bonding_option('miimon'), $cfg->{'miimon'}]) if ($cfg->{'miimon'});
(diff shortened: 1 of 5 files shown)


iliaross:
Hello,

Thanks for this PR.

I've checked it, and this mostly addresses the reported issues correctly.

I’ll merge it and follow up with a small adjustment to narrow live device deletion to recognized bond, VLAN, and bridge interfaces.

PR #2804, opened 2026-08-02
Add Btrfs qgroup management APIs

iliaross:
Hey Jamie,

This PR adds Linux quota helpers as requested in https://github.com/virtualmin/virtualmin-gpl/issues/1259 for detecting Btrfs filesystems, reporting full and simple quota status, inspecting subvolume qgroups, and managing limits and qgroup, with support older btrfs-progs releases through a compatibility fallback.

### DIFF
--- a/quota/linux-lib.pl
+++ b/quota/linux-lib.pl
@@ -1084,6 +1084,392 @@ my $out = &backquote_logged(
 &error($out) if ($?);
 }
 
+=head2 is_btrfs_fs(path)
+
+Returns 1 if a path is on a mounted Btrfs filesystem, 0 otherwise.
+
+=cut
+sub is_btrfs_fs
+{
+my ($path) = @_;
+return 0 if (!&valid_btrfs_path($path));
+
+# Btrfs subvolumes can have a different st_dev from the filesystem mount, so
+# select the longest containing mount path instead of comparing device numbers.
+my $best;
+foreach my $m (&mount::list_mounted()) {
+	next if (!defined($m->[0]) || !defined($m->[2]));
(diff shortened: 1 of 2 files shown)


quota/linux-lib.pl
```diff
@@ -1084,6 +1084,392 @@ sub set_group_quota
...
+	$best = $m if (!$best || length($m->[0]) > length($best->[0]));
+	}
+return $best && $best->[2] eq "btrfs" ? 1 : 0;
+}
+
+# valid_btrfs_path(path)
+# Returns 1 for an absolute path that is safe to pass to Btrfs tools.
+sub valid_btrfs_path
+{
+return defined($_[0]) && $_[0] =~ /^\// && $_[0] !~ /[\r\n\0]/ ? 1 : 0;
```
jcameron: Best to use a proper function name here instead of `$_[0]`

jcameron:
Is this the entire implementation of BTRFS support, or will you also need other changes to call these new functions from elsewhere in the `quota` module?

quota/linux-lib.pl
iliaross: Agreed! Fixed https://github.com/webmin/webmin/pull/2804/commits/99ab9159c322407c7172a94d706eea99f1af1bde.

iliaross:
> Is this the entire implementation of BTRFS support, or will you also need other changes to call these new functions from elsewhere in the `quota` module?

This only adds the backend support. We’ll still need a follow-up to connect it to the Webmin Quotas module, and Virtualmin will be able to call these functions directly in a follow-up too.

PR #2805, opened 2026-08-03
Fix openSUSE logrotate vendor overlay handling

### LINKED ISSUE #2682: Log rotation changes cause errors Opensuse Leap 16.0
In older Opensuse versions, there was /etc/logrotate.conf and /etc/logrotate.d. 
Now there's also the system defined /usr/etc/logrotate.conf and /usr/etc/logrotate.d which you're not supposed to edit.
A few other packages are now working like this too, supposedly for 'immutable' (read-only) systems 

An upgrade causes duplicate config errors, and requires you to remove `include /etc/logrotate.d `  in your config, because of the new `logrotate-all` script below that combines them - complains of dup's otherwise

Removing that line means webmin appears unaware of the files in /etc/logrotate.d - they can't be edited. 
Leaving it in causes lograte process to fail when it runs (because of the dup's)

logrotate-all  - pasted below.

How best to work around this change?

```
#!/bin/sh
set -eu

configs=

# Only read /usr/etc/logrotate.conf if /etc/logrotate.conf does not exist
if ! [ -e /etc/logrotate.conf ]; then
	configs="$configs /usr/etc/logrotate.conf"
else
	configs="$configs /etc/logrotate.conf"
fi

# Then read in all of {/usr,}/etc/logrotate.d/*, with /etc/ overriding /usr/etc/.
dirs=
[ -d /usr/etc/logrotate.d ] && dirs="/usr/etc/logrotate.d"
[ -d /etc/logrotate.d ] && dirs="$dirs /etc/logrotate.d"

if [ -n "$dirs" ]; then
	for confname in $(find $dirs -type f -printf "%P\n" | sort -u); do
		if [ -e "/etc/logrotate.d/$confname" ]; then
			configs="$configs /etc/logrotate.d/$confname"
		else
			configs="$configs /usr/etc/logrotate.d/$confname"
		fi
	done
fi

exec /usr/sbin/logrotate $configs

```

iliaross:
hello Jamie!

This PR updates the Log File Rotation module for openSUSE Leap 16's split `/usr/etc` and `/etc` configuration model.

- Select the local main configuration when present, otherwise use the vendor configuration.

- Recursively merge vendor and local drop-ins using the same precedence as `logrotate-all`.

- Display vendor configurations while keeping `/usr/etc` read-only.

- Copy vendor files into matching `/etc` overrides before editing or deleting them.

- Preserve empty overrides so deleted vendor rotations remain disabled.

- Use the openSUSE wrapper for scheduled rotation and the effective configuration set for forced rotation.

- Keep behavior unchanged on distributions that do not enable the vendor overlay.

Includes regression coverage for selection, precedence, copy-on-write, deletion, scheduling, backups, and protection against direct vendor writes.

Closes [#2682](https://github.com/webmin/webmin/issues/2682).

### DIFF
--- a/logrotate/logrotate-lib.pl
+++ b/logrotate/logrotate-lib.pl
@@ -435,8 +435,7 @@ return undef;
 }
 
 # save_directive(&parent, &old|name, &new, [indent])
-# Updates one entry identified by either its name or parsed object.  Vendor
-# entries are transparently copied and rebound to their writable overrides.
+# Updates one entry identified by either its name or parsed object
 sub save_directive
 {
 my ($parent, $oldv, $newv, $indent) = @_;
@@ -446,10 +445,22 @@ my $new = !defined($newv) ? undef : ref($newv) ? $newv :
 			{ 'name' => $old ? $old->{'name'} : $oldv,
 		     	  'value' => $newv };
 
-# Find the vendor file behind this write.  Existing directives use their own
-# file, while additions use the parent section or the effective main config.
+# Deleting an entry that is already absent is a true no-op. In particular,
+# do not put a missing local main config into the writable line cache.
+return if (!$old && !$new);
+
+# Find the file behind this write. Existing directives use their own file,
+# new sections may name a separate file, and other additions use the parent
+# section or effective main config.
 my $vendor_file;
-my $write_file = $old ? $old->{'file'} : $parent->{'file'};
+my $write_file = $parent->{'file'};
+if ($old) {
+	$write_file = $old->{'file'};
+	}
+elsif ($new && $new->{'file'} &&
+       !($parent->{'global'} && !$new->{'members'})) {
+	$write_file = $new->{'file'};
+	}
 if ($write_file) {
 	my $shadowed_vendor = &get_vendor_config_file($write_file);
 	if (&is_vendor_main_config($write_file) ||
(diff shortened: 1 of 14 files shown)


logrotate/save_log.cgi
```diff
@@ -4,8 +4,22 @@
...
+# On systems with vendor configuration below /usr, create the writable local
+# main config before changing global options.  The parent object intentionally
+# keeps this local path as its write destination.
+&ensure_local_main_config() if ($in{'global'});
 $parent = &get_config_parent();
 $conf = $parent->{'members'};
+
+# A local drop-in shadows the whole vendor file, so copy it intact before
+# editing or deleting one section.  Rotate Now is read-only and needs no copy.
+if (!$in{'global'} && !$in{'new'} && !$in{'now'} &&
```
jcameron, line 43: Wouldn't it be better to handle this case at the API level inside logrotate-lib.pl ? Otherwise, would other callers of the API (like Virtualmin) still work?

iliaross: You're right, it would be better for other callers to work. Thanks for pointing this out! Fixed here  https://github.com/webmin/webmin/pull/2805/commits/b7aed037587dde5d06865f17ad8a382c4b6e46f8.

logrotate/t/run-tests.t
```diff
@@ -24,24 +32,57 @@ print $fh $text;
 close($fh) or die "close $file: $!";
 }
 
+# read_text(file)
+# Returns the complete contents of a text fixture
+sub read_text
```
jcameron: Don't we have `read_file_contents` for this already?

logrotate/logrotate-lib.pl
```diff
@@ -28,28 +33,235 @@ sub get_config_parent
...
+# even an incomplete local file would hide the valid vendor configuration.
+if (!$ok || !&set_ownership_permissions(undef, undef, 0644, $dest)) {
+	$err ||= $!;
+	&unlink_file($dest) if (-e $dest || -l $dest);
+	&error(&text('save_ecopy', "<tt>".&html_escape($dest)."</tt>",
+		&html_escape($err)));
+	}
+
+# Force the next read to select and parse the newly-created local file.
+&flush_logrotate_config_cache();
```
jcameron, line 147: Wouldn't it be better to just fix the cache (if there is one) by updating the `file` field, rather than doing a full flush?

logrotate/t/run-tests.t
iliaross: It's for isolated tests and intentional, so it does not load `web-lib-funcs.pl`.

logrotate/logrotate-lib.pl
iliaross: We could, but every cached entry from that file would need updating. Flushing and reparsing is simpler and safer, and only happens once.

jcameron: Yes but if any other code is holding onto a copy of the config structure, it will be invalid now..

logrotate/t/run-tests.t
jcameron: Oh ok ... maybe in future this should change so that tests aren't repeating a lot of code?

logrotate/logrotate-lib.pl
iliaross: Yes, but all current callers reload the config after copying, so it’s safe.

logrotate/t/run-tests.t
iliaross: Yeah, in the future we could add something like `t/test-lib.pl` for common test helpers if more tests need them.

jcameron: good idea!

logrotate/logrotate-lib.pl
jcameron: So it looks like this function can be called from `save_directive` via `ensure_local_main_config` . It really doesn't seem safe to me to be flushing the config cache inside a function that is updating the config structure!

iliaross: Jamie, you're right in general. However, here, the cache is cleared before anything changes.

`save_directive()` then reloads and updates the objects it received before writing. No current caller uses any other old references afterward.

To be clear, this happens only the first time a vendor file is edited. We copy it from `/usr/etc` to `/etc`, clear the old cache, reload, and save. Later edits use the `/etc` copy normally.

I don't see any risk in this case.

logrotate/t/run-tests.t
iliaross: Alright, we can add it now https://github.com/webmin/webmin/pull/2805/commits/07f0ddfb6dcf4a7bd2cf010d9dff0f2e612eca9b.

jcameron: Nice!

logrotate/logrotate-lib.pl
jcameron: I know I said it would make more sense to do this inside save_directive, but now I regret that! Actually I think that any copying of vendor configs should be done in in the logrotate module .cgi scripts, so that it can happen before the code starts manipulating config objects..

iliaross: > I know I said it would make more sense to do this inside save_directive, but now I regret that! Actually I think that any copying of vendor configs should be done in in the logrotate module .cgi scripts, so that it can happen before the code starts manipulating config objects..

Ah, I actually agree that it would probably be better design-wise to do it outside `save_directive`. I will look into reworking this when I finish testing CF patch...

iliaross: Alright, I've fixed it in this https://github.com/webmin/webmin/pull/2805/commits/5969c33e060c042adbcd7c722aba6e8d2eb76116 patch.  And, actually, it's a better approach and risk-free because there are no `logrotate:save_directive` calls outside the Webmin logrotate module.

iliaross:
Jamie, are we all good here, or is there anything else left?

PR #2807, opened 2026-08-03
Add Btrfs subvolume quota management

iliaross:
Hey Jamie!

This PR adda Btrfs subvolume quota support to the Disk Quotas module, including full and simple accounting modes, quota lifecycle actions, and qgroup usage and limit management.

Includes ACL enforcement, compatibility with older btrfs-progs versions, destructive-action confirmation, top-level qgroup protection, help, and tests.

https://github.com/virtualmin/virtualmin-gpl/issues/1259


quota/config_info.pl
```diff
@@ -0,0 +1,36 @@
...
+return if (@btrfs && &has_command("btrfs"));
+
+# Remove the field from both the configuration metadata and display order.
+delete($info->{'btrfs_mode'});
+@$order = grep { $_ ne "btrfs_mode" } @$order if ($order);
+}
+
+# show_btrfs_mode(mode)
+# Display the accounting mode selector, defaulting unknown values to full mode.
+sub show_btrfs_mode
```
jcameron: If the options are fixed here, why a custom function instead of just putting these options into config.info ?

quota/linux-lib.pl
```diff
@@ -1398,6 +1501,43 @@ sub set_btrfs_qgroup_limit
 push(@args, $qgroup) if (defined($qgroup));
 push(@args, $path);
 my ($out, $err) = &run_btrfs_command(1, @args);
+
+# An exceeded qgroup can block the metadata write needed to raise or remove its
```
jcameron: Wow, really? Is this the case for all qgroups ?

quota/index.cgi
```diff
@@ -97,13 +107,84 @@ if (@list) {
...
-else {
+# Report no support only when neither traditional nor Btrfs filesystems exist.
+elsif (!@btrfs) {
 	print "<b>$text{'index_nosupport'}</b><p>\n";
 	if (&foreign_available("mount")) {
 		print &text('index_mountmod', "../mount/"),"<p>\n";
 		}
 	}
 
+# Btrfs subvolume quotas use qgroups instead of Unix users and groups, so they
```
jcameron, line 118: Out of interest, what are qgroups exactly?

Also, can regular Unix user and group quotas still be used on BTRFS ?

quota/config_info.pl
iliaross: No need. It's a leftover from previous tries and tests. Fixed now https://github.com/webmin/webmin/pull/2807/commits/65e171a92ff46cc752617d39fbbc467acd918f65.

quota/linux-lib.pl
iliaross: Not all `qgroups`. This can happen when the qgroup associated with the command path, or one of its parents, is over its limit. The retry only runs when Btrfs actually returns `EDQUOT`.

I’ve updated the comment and tests to make this clearer https://github.com/webmin/webmin/pull/2807/commits/b433cdd810d7065b1f25eb3366b6b219ebcbdf90.

quota/index.cgi
iliaross: > Out of interest, what are qgroups exactly?

`qgroups` are Btrfs quota groups, where each subvolume gets a level-0 `qgroup`, such as `0/256`, which tracks its referenced and exclusive space. Higher-level `qgroups` can combine multiple `subvolumes` and snapshots under one limit. They are unrelated to Unix users and groups. Btrfs does not support traditional UID/GID quotas such as `usrquota` and `grpquota`. Per-user limits require separate `subvolumes` with `qgroup` limits. 

> Also, can regular Unix user and group quotas still be used on BTRFS 

Traditional quotas can still be used at the same time on other filesystems such as ext4 or XFS, and this is one of the reasons why it was hard to implement support for it.

jcameron: Oh .. so in that case, can we still offer the same API for user and group quotas that Virtualmin can already call for other filesystem types?

iliaross: No, not at all.

Webmin's existing quota API is based on Unix UID and GID quotas and supports block and inode soft and hard limits.

Btrfs `qgroups` are attached to subvolumes, not users or groups, and provide referenced and exclusive byte limits. 

That said, I'd rather keep Btrfs support in separate `qgroup`-specific APIs instead of making the generic user and group quota APIs imply UID and GID semantics that Btrfs does not provide.

And traditional Unix user and group quotas cannot be enabled on Btrfs itself, but they can still be used on other ext4 or XFS filesystems on the same system alongside Btrfs.

iliaross:
Jamie, this one should also be good to merge. Do you have any other questions?

quota/index.cgi
jcameron: Ok got it, thanks

quota/config_info.pl
jcameron: Thanks!

quota/linux-lib.pl
jcameron: The command path meaning the `btrfs` command? Do you mean that it could fail to **execute** because the qgroup is out of quota?

iliaross: No, no... the `btrfs` command still executes. By "command path" I meant the filesystem path passed to it, such as `/`.

If the `qgroup` for the `subvolume` containing that path, or one of its parents, is over quota, the kernel can reject the limit transaction with `EDQUOT`.

I’ve reworded the comments in https://github.com/webmin/webmin/pull/2807/commits/40f023922467c56128fa950f91afc83ac6d15bf4 to make this clear.

jcameron: So does this mean that if the parent volume is over quota, it may be impossible to update quotas on a sub-volume? That's pretty surprising behavior! Is this documented anywhere?

quota/linux-lib.pl
```diff
@@ -1398,6 +1501,43 @@ sub set_btrfs_qgroup_limit
...
+# own limit. Retry once with the kernel's administrative override, preserving
+# the previous state and restoring enforcement immediately after the command.
+if ($err && $err =~ /disk quota exceeded/i) {
```
jcameron: Can we be certain this error will always be in english?

quota/linux-lib.pl
jcameron: Also, changing kernel parameters silently inside the API seems rather scary to me! Especially if it crashes half way through and leaves the system in an unusual state..

quota/linux-lib.pl
iliaross: Yes, because we call `run_btrfs_command` with `clean_language`.

quota/linux-lib.pl
iliaross: > So does this mean that if the parent volume is over quota, it may be impossible to update quotas on a sub-volume? That's pretty surprising behavior! Is this documented anywhere?

The "parent" means a parent `qgroup`, not a parent volume. Btrfs documents that limits apply through the [`qgroup` hierarchy](https://btrfs.readthedocs.io/en/latest/btrfs-qgroup.html), but not this specific limit-update failure clearly.

> Also, changing kernel parameters silently inside the API seems rather scary to me! Especially if it crashes half way through and leaves the system in an unusual state..

Okay, I removed it. The API now no longer changes this setting and would just return the Btrfs error.

iliaross:
Jamie, I think we're good here?

quota/linux-lib.pl
jcameron: Yeah I think that's safer, thanks!

PR #2811, opened 2026-08-05
Fix #2810

### LINKED ISSUE #2810: XML-RPC endpoint loses type information
We use the XML-RPC API to automate kea-dhcp configuration by calling functions in kea-dhcp-lib.pl which in general works fine.

One call uses a complex object with nested keys and values which are later translated into JSON.

Some properties are numbers, like `'rebind-timer': 1800`. In the resulting JSON/dhcp.config they are represented as `"rebind-timer": "1800"`, which kea-dhcp rightfully rejects.

The request is correct and transmits the parameters/properties as
`
<int>1800</int>
`

This seems to be ignored in `sub parse_xml_value` in xmlrpc-lib.pl, line 20 ff.

miq:
XML-RPC now retains type information on scalars.

**Note: I am not a perl coder and the actual implementation is implemented using AI.**

However, it fixes our use cases and looks ok to me, but I cannot judge on idiomaticness and edge cases.

### DIFF
--- a/xmlrpc-lib.pl
+++ b/xmlrpc-lib.pl
@@ -18,7 +18,13 @@ my ($base64) = &find_xmls("base64", $value, 1);
 my ($struct) = &find_xmls("struct", $value, 1);
 my ($array) = &find_xmls("array", $value, 1);
 if ($scalar) {
-	return $scalar->[1]->[2] // "";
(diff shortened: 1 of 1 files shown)


iliaross:
Thanks for the patch!

PR #2812, opened 2026-08-09
Remove bundled Miniserv certificate and safely fall back to HTTP

iliaross:
Hey Jamie,

This PR drops the fallback to a pre-created SSL certificate, as we all discussed.


pam_login.cgi
```diff
@@ -74,15 +74,21 @@ print &ui_form_start("@{[&get_webprefix()]}/pam_login.cgi", "post");
...
-if ($ENV{'HTTPS'} ne 'ON' && $miniserv{'ssl'}) {
-	my $link = ui_tag('a', "&#9888; $text{'login_notsecure'}",
-		{ 'href' => "javascript:void(0);",
-		  'class' => 'inherit-color',
-		  'onclick' => "window.location.href = ".
-		    "window.location.href.replace(/^http:/, 'https:'); return false;",
-		});
-	$not_secure = ui_tag('span', $link,
-		{ class => 'not-secure', title => $text{'login_notsecure_desc'} });
+if ($ENV{'HTTPS'} ne 'ON' &&
```
jcameron: Instead of repeating this on pam_login.cgi and session_login.cgi , move it to a common function in web-lib-funcs.pl

iliaross: Done https://github.com/webmin/webmin/pull/2812/commits/6afba8bc12e9307b8f92b286cd8e4d50107d50c3.

login-lib.pl
```diff
@@ -0,0 +1,27 @@
+# login-lib.pl
```
jcameron, line 1: Why not put these in web-lib-funcs.pl , instead of a new lib file?

iliaross: Because we don't need this to be in WebminCore and exported.

jcameron: Ok ... long term we should think about maybe moving these into a module, rather than having a bunch of top-level libraries.

iliaross: Like which module? `webmin/webmin-lib.pl` for example?

jcameron: Yeah maybe the login CGI scripts should be moved into a module? Not a big deal though, as migration could be tricky..

iliaross: Yeah, it’s not a big deal to worry about and make such drastic changes. I think we are fine, and later today we should make a new Webmin and Usermin release after wrapping up this last https://github.com/webmin/webmin/pull/2805 PR.

PR #2813, opened 2026-08-13
Add ability to use `vfsv1` for newly created Linux quota files

iliaross:
Hello Jamie!

This PR adds ability to use the `vfsv1` quota format for newly created user and group quota files when supported by quota-tools 4 or newer. This removes the `vfsv0` range limitation while preserving existing quota files and configured journal formats.

And, quota creation remains scoped to the requested user or group type, safely normalizes configured `quotacheck` flags, and retains `vfsv0` and legacy fallbacks for compatibility.

Addresses issue report in:
https://forum.virtualmin.com/t/quota-out-of-range/137331?u=ilia

### DIFF
--- a/mount/linux-lib.pl
+++ b/mount/linux-lib.pl
@@ -1924,7 +1924,30 @@ elsif ($_[0] =~ /^ext\d+$/) {
 		($u, $g) = ("usrjquota", "grpjquota");
 		$jufile ||= "aquota.user";
 		$jgfile ||= "aquota.group";
-		$options{"jqfmt"} = "vfsv0";
+		if (!$options{"jqfmt"}) {
+			# Keep the format of existing external quota files when
+			# switching them to journaled quotas.
+			my $jqfmt;
+			foreach my $qfile ($jufile, $jgfile) {
+				next if (!-s "$_[2]/$qfile");
+				if (open(my $qfh, "<", "$_[2]/$qfile")) {
+					my $header;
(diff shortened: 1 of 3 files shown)


mount/linux-lib.pl
```diff
@@ -1924,7 +1924,30 @@ sub check_options
 		($u, $g) = ("usrjquota", "grpjquota");
 		$jufile ||= "aquota.user";
 		$jgfile ||= "aquota.group";
-		$options{"jqfmt"} = "vfsv0";
+		if (!$options{"jqfmt"}) {
```
jcameron, line 1927: So in summary, does this code check the existing quota files to see if they are in the old format, and if not (or if they don't exist) does it add the flag to swich to vfsv1? Or is it the other way around?

iliaross: The code checks for existence first. It inspects the format only where the mount configuration needs an explicit `jqfmt`.

It _never_ sees an existing `vfsv0` file and "switches" it. The explicit `-F vfsv1` flag is used only when there is no quota database yet (`aquota.user` and `aquota.group`), to be sure that a newly created database uses the larger format without risking existing limits.

That said, a system administrator can switch manually to `vfsv1` by turning quotas off, deleting the `aquota.user` and `aquota.group` files, turning quotas back on with the patched Webmin, and having Virtualmin set the quotas again.

jcameron: Ok makes sense thanks!

PR #2827, opened 2026-08-28
Add MySQL binary log management

iliaross:
Hi, Jamie!

This PR adds a "Binary Logging" page to configure log format, file size and retention, rotate or purge logs, and support restore sessions that do not write back to binary logs.


mysql/save_binlogs.cgi
```diff
@@ -0,0 +1,162 @@
...
+&ReadParse();
+
+# Get the mysqld section
+foreach my $l (&get_all_mysqld_files()) {
+	&lock_file($l);
+	}
+$conf = &get_mysql_config();
+# Prefer the main server section over generic MariaDB sections that can
+# belong to a plugin-specific include file
+($mysqld) = grep { $_->{'name'} eq 'mysqld' } @$conf;
```
jcameron: This code has been repeated in a few places, maybe it should be factored out into a separate function

iliaross: Good suggestion, thanks! Fixed now in this patch https://github.com/webmin/webmin/pull/2827/commits/7f237871ba784fde14e260c7ce215eac485eba30.

PR #2829, opened 2026-08-29
Fix errors in Turkish (tr) language files

tbsagdic:
This PR fixes verifiable defects in the shipped Turkish translations across 25 modules.

Only `lang/tr` files are touched — no `.auto` (machine-translated) file is modified, and no key is added or removed. Every file keeps its original key order and line count.

### What was fixed

| Category | Keys |
|---|---|
| Placeholder (`$1`/`$2`/`$3`) mismatches vs `lang/en` | 33 |
| Dropped or swapped HTML tags | 19 |
| Double-encoded UTF-8 (mojibake) | 10 |
| Strings left in English | 96 |

### Notable defects

* **`fdisk/mkfs_desc1`** used `$1` twice, so the *"All data on this partition will be permanently erased"* confirmation displayed the filesystem type where the target device name belongs.
* **`bind8/mass_desc`** translated the literal `named.conf` keywords `master`, `slave`, `stub` and `forward` into Turkish. A user following the instructions would enter invalid configuration values. These are now left in English.
* **`useradmin/batch_desc6`** and **`batch_desc3`** had the literal strings `$batch_desc1` and `$batch_desc0` as their values and rendered as plain text.
* **`useradmin/export_eopen`** contained the Perl special variable `$!` instead of `$1`.
* **`apache/reconfig_ever`** linked to `<a href='$1'>` instead of `$2`, so the *module configuration* link pointed at the version string.
* **`mailboxes/lang/tr`** contained 10 double-encoded lines, e.g. `klasÃ¶rÃ¼nden` instead of `klasöründen`.
* **`htaccess-htpasswd/lang/tr`** was a verbatim copy of `lang/en` — the module had no Turkish at all. It is now fully translated.

### Deliberately not changed

* 5 keys that no longer exist in `lang/en` (dead keys) — removing them belongs in a separate cleanup.
* `proc` button labels padded with non-breaking spaces — `lang/en` and `lang/de` use the same padding convention.
* Technical identifiers such as `LVM VG $1, LV $2` and `IPsec (no AH)`, which should stay in English.

### Verification

Each modified file was checked to be valid UTF-8, free of trailing whitespace, and to have placeholders matching `lang/en` exactly. Line endings are LF.


tbsagdic:
Just a note that this one is also finished and ready for review.

It only touches strings that were already in the shipped Turkish files and were verifiably wrong against `lang/en`: 33 placeholder mismatches, 19 markup mismatches, 10 double-encoded UTF-8 lines and 96 lines left in English. 25 files, 178 lines, no new keys.

It is independent of #2830 and the two branches do not overlap, so they can be merged in either order.

PR #2830, opened 2026-08-29
Complete the Turkish (tr) translation

tbsagdic:
This completes the Turkish translation of Webmin.

**lang/tr — 31,522 of 31,522 strings (100%), across all 124 modules with a `lang` directory.**

Before this branch, 63 modules had a partial or missing `lang/tr` and the gaps were filled by `lang/tr.auto`. Every reviewed string has now been moved into `lang/tr`, and all `lang/tr.auto` files are removed — the same way `de.auto` was retired in #2274.

**module.info — 245 of 245 descriptions (100%), across all 125 modules.**

37 were previously reviewed, 194 came from `module.info.tr.auto` and 14 modules had no Turkish description at all (grub2, hardware-info, kea-dhcp, logviewer, nftables, systemd, xterm), so they showed English titles in the menu. `module.info.tr.auto` now holds only `name_tr`, matching how `de` and `ca` are organised; the 36 files that became empty are deleted.

### Defects fixed in the pre-existing Turkish

The work was driven by a script that compares every `lang/tr` line against its `lang/en` counterpart. It found 58 lines that were wrong independently of the missing translations, and all of them are fixed here:

- **25 missing or wrong `$N` placeholders.** The affected messages silently dropped a filename, hostname or count at runtime — or printed the wrong one. `adsl-client/index_stopdesc` and `inittab/log_rename` each repeated `$1` where `$2` was meant, so both values printed the same string; `useradmin/export_eopen` contained a literal `$!`; `cron/env_ename` carried a `$1` that `lang/en` does not have.
- **13 mismatched HTML markup.** Several of these were also truncated mid-sentence, dropping a whole sentence of the original; those are now translated in full.
- **4 encoding defects.** `mailboxes/confirm_warnf`, `cron/desc_tr` and `mailboxes/desc_tr`/`longdesc_tr` were stored as double-encoded UTF-8 and displayed as mojibake.
- **4 lines with trailing whitespace** in `proc`.
- **Untranslated strings** left in English in `fdisk`, `mount` and `shorewall`.

Some of the replaced machine text was wrong rather than merely clumsy — `sarg`'s description had "Squid" translated as *kalamar* (the fish), `proc` had "renice" as "rename", and `firewall`/`firewall6` had lost the `(iptables)` / `(ip6tables)` qualifiers that tell the two modules apart.

### Upstream `lang/en` issues noticed along the way

Not changed here, only flagged:

- `postgresql/acl_login` reads "Login to **MySQL** as" — a copy-paste slip from the mysql module.
- `rbac/auth_err` reads "Failed to authorization" (missing "save"); `rbac/policy_eclash` spells "Algorithsm".
- `firewall/move_chain` reads "Current change" where "Current chain" is clearly meant.
- `bsdfdisk` and `format` both spell "paritions".

These are translated by intent rather than literally.

### Conventions followed

- Key order, comments, blank-line grouping and line endings of `lang/en` are preserved exactly; the files are generated from `lang/en` so no key can be dropped or added.
- Protocol keywords, directive names, command names and literal configuration values stay in English, since they are text the user has to read in or type into a config file — `systemctl`, `journalctl`, `[Install]`, `.rhosts`, `known_hosts`, `ssh-keygen`, `TCP_NODELAY`, `PASV`, `RFC2228`, `master`/`slave`/`stub`/`forward`, `mdadm --grow`, `endscript`, `Makefile.PL`, `qgroup`, `noatime`, `INBOX`, and so on.
- Product and specification names with no established Turkish form are left as they are: Novell Directory Services, Active Directory, System Security Services, Boot Loader Specification, Trusted Platform Module, Alpine Package Keeper, Service Management Facility.
- Strings shared between related modules (`cluster-software`/`software`, the four `iscsi-*` modules, `pptp-*`/`pap`) are reused only where the English source text is byte-identical.


jcameron:
Nice, let us know when this is done!

tbsagdic:
@jcameron It's done — this is ready for review.

**`lang/tr`: 31,522 / 31,522 strings (100%), all 124 modules.** Every reviewed string is now in `lang/tr` and all `lang/tr.auto` files are removed, the same way `de.auto` was retired in #2274.

**`module.info.tr`: 245 / 245 descriptions (100%), all 125 modules.** Seven modules had no Turkish description at all and showed English titles in the menu (grub2, hardware-info, kea-dhcp, logviewer, nftables, systemd, xterm). `module.info.tr.auto` now holds only `name_tr`, matching how `de` and `ca` are organised; the 36 files that became empty are deleted.

I also fixed 58 lines that were already wrong in the shipped Turkish, independently of the missing translations — 25 missing or wrong `$N` placeholders, 13 markup mismatches (several also truncated mid-sentence), 4 double-encoded UTF-8 strings, 4 lines with trailing whitespace, plus a few strings left in English. The PR description has the full breakdown.

Every file is generated from `lang/en`, so key order, comments, blank-line grouping and line endings are preserved and no key can be dropped or added. I re-ran the comparison against `lang/en` after the last commit: no placeholder, markup, encoding or whitespace mismatches remain. The only 14 intentional matches with English are product and specification names that have no established Turkish form (Novell Directory Services, Boot Loader Specification, Trusted Platform Module, Alpine Package Keeper, `LVM VG $1, LV $2`) and two `$key` references in `useradmin`.

A few things I noticed in `lang/en` and deliberately did **not** touch, in case you want them fixed separately:

- `postgresql/acl_login` reads "Login to **MySQL** as" — looks like a copy-paste from the mysql module.
- `rbac/auth_err` reads "Failed to authorization" (missing "save"), and `rbac/policy_eclash` spells "Algorithsm".
- `firewall/move_chain` reads "Current change" where "Current chain" seems to be meant.
- `bsdfdisk` and `format` both spell "paritions".

I translated those by intent rather than literally.

Happy to split this into per-module commits or squash it, whichever is easier to review. #2829 fixes a separate set of errors and does not overlap with this branch.

jcameron:
@iliaross what do you think of this PR?

tbsagdic:
Thanks for merging #2829. That put 173 of the same keys on master, so this branch went into conflict — I've merged master back in and it's mergeable again.

Resolved by rule rather than by hand: for any key #2829 already fixed and merged, master's value wins (54 keys); everything else keeps this branch's version. That keeps the diff limited to what is genuinely new here instead of re-editing work that already landed.

It also avoided a regression worth mentioning. This branch was cut before #2829, so it still carried the old double-encoded UTF-8 in nine `mailboxes` strings (`confirm_warnallf`, `left_autoreply`, `left_folders`, `left_forward`, `left_pass`, `left_sig`, `mail_fromfilter`, `mail_selread`, `mail_subfilter`). Taking master's value for those keeps them fixed — a plain "keep ours" would have reintroduced the mojibake. A scan over the whole tree after the merge reports 0 remaining broken lines.

`master` also added `acl/forgot_sudouser`, which is now translated, so `lang/tr` is still complete: **31,523 of 31,523 keys, all 124 modules**, plus 245 of 245 `module.info.tr` descriptions across 125 modules.

@iliaross happy to answer anything on the Turkish side, or to reshape the branch if that helps review — per-module commits, or squashed into one. It's currently 44 commits grouped by batch of modules.

webmin/lang/tr
```diff
@@ -249,11 +492,15 @@ startpage_deftab=Öntanımlı kategori
 startpage_nohost=Sürüm, makine adı ve işletim sistemi gösterilsin mi?
 startpage_gotonone=&lt;Hiçbiri&gt;
 startpage_gotomodule=Girişten sonra her zaman bu modüle git
+startpage_webminup=Webmin güncellemeleri Sistem Bilgileri sayfasında gösterilsin mi?
+startpage_moduleup=Modül güncellemeleri Sistem Bilgileri sayfasında gösterilsin mi?
 
 upgrade_title=Webmin Güncelleştir
 upgrade_desc=Bu sayfa varolan Webmin'i daha yeni bir sürümüne güncellemek için kullanılır. Yerel dosyadan, yüklenebilecek bir dosyadan veya <a href="webmin.com">webmin.com</a> adresinden en son sürümünü kurabilirsiniz. Eğer elle yapacaksanız bütün yapılandırma ayarları ve üçüncü parti modüller saklanacaktır.
```
iliaross: `href="webmin.com"` will be treated as a relative URL and resolved against the current Webmin server.

proc/lang/tr
```diff
@@ -113,24 +138,31 @@ sysv_rgroup=Gerçek grup
 sysv_pgid=İşlem grup numarası
 sysv_tty=TTY
 sysv_stime=$stime
```
iliaross: `stime=Başlama zamanı` is already defined, but `__norefs=1` prevents `$stime` from being expanded, so these four values display `$stime` literally!

For it to work, you would want either remove `__norefs=1` from both `proc/lang/en` and `proc/lang/tr`, or keep it and replace each `$stime` with `Başlama zamanı`. And, I think replacing the four values is the safer, translation-only fix.

Though, I would personally like to hear Jamie's thoughts on that! Let's wait until he comes back around.

tbsagdic:
@iliaross Thanks — both are real, and both are fixed in the latest commit.

**`href="webmin.com"`** — you're right, that resolves against the Webmin server. `lang/en` has `<a href=https://webmin.com>`, so the Turkish now matches it exactly. I checked every Turkish file for the same pattern and this was the only relative `href` in the tree.

**`$stime` under `__norefs=1`** — agreed, and I went with the translation-only fix you preferred. What settles it for me is that `lang/en` doesn't use a reference there either: `linux_stime`, `freebsd_stime`, `hpux_stime` and `sysv_stime` are all spelled out as `Started`, so the Turkish was deviating from English rather than following it. Removing `__norefs=1` would change behaviour for every language in the module, which feels out of scope for a translation PR — but if Jamie would rather go that way, this commit is trivial to revert.

I then scanned every module for the same class of defect — a `$key` or `${key}` reference in a file whose module sets `__norefs=1` — and it turned up two more things:

- **`useradmin/uedit_dothers` was `$udel_dothers`**, and `useradmin` also sets `__norefs=1`, so it had the same symptom. `lang/en` spells it out, so it is fixed the same way. Thanks for putting me onto this — I would not have found it otherwise.

- **`heartbeat/conf_auto_on` and `conf_auto_off` are `${conf_enabled}` and `${conf_disabled}` in `lang/en` itself**, and `heartbeat` sets `__norefs=1` too. So English is affected as well, not just Turkish — and `conf_auto_` references `${default}`, which is not a key in that file at all. That is an upstream bug rather than a translation one, so I have deliberately left it alone here rather than making Turkish silently diverge from English. The `de` translation copied the same reference. Happy to send a separate PR for it if you want.

Current state: `lang/tr` complete at 31,523 of 31,523 keys across 124 modules, `module.info.tr` at 245 of 245 across 125 modules, no remaining placeholder, markup, encoding, whitespace or broken-reference mismatches against `lang/en`.

tbsagdic:
The failing `prove` job on this PR isn't caused by these changes.

The failure is `./hardware-info/t/perlcritic.t` — 26 `return undef` and `return sort` violations in `hardware-info/hardware-info-lib.pl`. That is Perl source; this branch touches only `lang/*` and `module.info*` files and modifies no `.pl` or `.cgi` at all:

```
$ git diff --name-only origin/master...HEAD | grep -vE '(lang/|module\.info)'
(empty)
```

The same test already fails on `master`:

- [33343226175](https://github.com/webmin/webmin/actions/runs/33343226175) — the merge commit of #2829
- [33289923115](https://github.com/webmin/webmin/actions/runs/33289923115) — "Fix sending of forgot password link to non-root users"
- [33221011212](https://github.com/webmin/webmin/actions/runs/33221011212) — "Fix IPsec host key generation with modern Libreswan"

All three show the identical `Failed test './hardware-info-lib.pl perlcritic' at ./hardware-info/t/perlcritic.t line 56`. It looks like it has been failing since the module was added in 7bb24044f2, so this PR just inherits it by merging `master` in.

@iliaross happy to send a separate PR for it if useful, though `return undef` → `return` changes what the function yields in list context, so it needs a closer look than a blanket search-and-replace.

acl/lang/tr
```diff
@@ -249,3 +377,161 @@ hide_clone=(Klon $1)
...
+forgot_adminmsg=Bu e-postayı $3 adresindeki Webmin sisteminin yöneticisinden, $1 girişi için alıyorsunuz.\n\nParolayı sıfırlamaya devam etmek isterseniz şu bağlantıyı izleyin :\n$2
+forgot_sending=$2 için parola sıfırlama e-postası $1 adresine gönderiliyor ..
+forgot_sent=.. gönderildi
+forgot_link=Aşağıdaki bağlantı, $1 için Webmin parolasını önümüzdeki $2 dakika boyunca sıfırlamak üzere kullanılabilir :
+forgot_enosudo=Bu sistemde sudo kullanılamıyor!
+forgot_ecansudo=Girilen kullanıcının sudo yetkisi yok
+forgot_eunix=Girilen sudo yetkili kullanıcı mevcut değil!
+forgot_sudouser=sudo yetkisi olan kullanıcının parolasını sıfırla
+
+__norefs=1
```
iliaross: This file has no final newline. Fix it, and I think we're good.

proc/lang/tr
jcameron: I would recommend not using references to $stime , as expanding these causes Webmin to spend longer parsing lang files and prevents them from being cached.

tbsagdic:
@iliaross Fixed — `acl/lang/tr` ends with a newline again.

It was my own regression: `master` added `acl/forgot_sudouser` while this branch was open, and the script I used to slot that key into the right position dropped the trailing newline. I checked all 481 files in this PR and it was the only one affected; every one of them now ends with a newline.

@jcameron Good to know about the parsing and caching cost — that settles it, and it's already done. The four `proc` references were replaced with literal text in [0c4941e8](https://github.com/webmin/webmin/pull/2830/commits/0c4941e8aa), before you commented, since `lang/en` spells `Started` out in all four places rather than referencing `stime`.

Scanning every module for the same pattern also turned up `useradmin/uedit_dothers`, which was `$udel_dothers` under `__norefs=1`; that one is replaced with literal text too.

One left, and it isn't a translation issue: `heartbeat/conf_auto_on` and `conf_auto_off` are `${conf_enabled}` and `${conf_disabled}` in `lang/en` itself, with `__norefs=1` set in that module, so English pays the same cost. `conf_auto_` also references `${default}`, which is not a key in the file at all. I left those alone rather than making Turkish diverge from English — happy to send a separate PR if you want them spelled out across `lang/en` and the translations that copied them.

No other `$key` references remain in any Turkish file.

iliaross:
Thanks for the patch!

PR #2835, opened 2026-09-05
Cleanup meta-confg layer

swelljoe:
A bug introduced by Copilot Code Review way back (where it decided the module should edit a custom `nftables.conf` inside the module config dir instead of editing the system rules), that I missed, has led to subsequent layers of crud being piled up to accommodate that bad judgment.

This removes that misbehavior, and reverts to reading/parsing/modifying the system `nftables.conf` (either in `/etc/nftables.conf` or `/etc/sysconfig/nftables.conf`). It also removes all the metadata management junk that was needed to make that behavior look normal and kinda work. (But, not really: https://forum.virtualmin.com/t/setup-nftables-questions/137928 )

It adds a postinstall to cleanup that misfeature and remove the extra webmin-nftables service.

The import rules feature remains, though it probably ought to get better docs discouraging wanton importing, since it could easily lead to duplicate rules or turning dynamic rules from fail2ban or whatever into permanent rules, which would be a problem. But, the ability to import Firewalld generated rules or other rules generated by scripty-firewall generators is probably useful, so I left it.

It also manages files that are included in the primary config file via `include`, both in the usual view and in the edit config files view.


iliaross:
> Don't merge yet, I'm still refactoring, I just like being able to scroll through the diffs.

Alright, let me know when it’s fully ready.

swelljoe:
OK, I think this sorts out all the issues. Implementation is a bit simpler, though there's more code because it handles converting from the dumb config file layout to using the normal system files in a postinstall. I'll delete that code after a few versions, as it seems unlikely to impact many people; just early adopters.

iliaross:
Thanks, Joe! I'll need to take a deeper look into it and get back to you.

iliaross:
Alright, I have finally taken a close look and am going to merge it after a few follow-up commits. Thank you, Joe!

PR #2836, opened 2026-09-07
Add return code redirects to Nginx URL rewriting pages

iliaross:
Hi Jamie,

This PR adds a "Redirect all requests" option to the server block and location URL Re-Writing pages in the Nginx Webserver module, backed by the Nginx return directive with a choice of 301, 302, 303, 307 or 308.

The existing rewrite table only offers the redirect and permanent flags, which map to 302 and 301. The other codes [requested on the forum](https://forum.virtualmin.com/t/new-nginx-module-server-block-settings-url-re-writing-return-code-303-308/137940) are not supported by rewrite at all, so they can only be provided through return.

The new option only manages return directives that are redirects. A bare URL form such as return `https://example.com/;` is shown as a 302, while non-redirect uses like return 404; are hidden from the page and never modified or removed. The target can be a path or a full URL, and the field shows an example of each as a placeholder.

Example of the resulting config for a server block:

```
return 308 "https://example.com$request_uri";
```


jcameron:
Looks good, thanks!!

iliaross:
> Looks good, thanks!!

You're very welcome!

PR #2839, opened 2026-09-10
Fix admin detection when RPC is disabled

### LINKED ISSUE #2837: Users created on 2.650+ get `rpc=0` by default and are treated as safe users → Authentic theme: "The Unix user X does not exist"
**Environment:** Webmin `2.660` with Authentic theme, Debian `13`. Webmin user `anyusername` (no Unix account of that name), all modules, all privileges, File Manager ACL *Access files as root* = Yes.

**Cause:** since d41377983 (2.650) a new user gets *Can accept RPC calls? = No* by default, i.e. `rpc=0` in `/etc/webmin/<user>.acl`. `webmin_user_login_mode()` in `web-lib-funcs.pl` (26fa48623, 1.995) treats `rpc=0` as `safe-user` for any login not named `root`/`admin`/`sysadm`, so `webmin_user_is_admin()` returns false for a fully privileged user. Users created before 2.650 have `rpc=2` and are unaffected, which is why identical setups behave differently.

**Effect:** the theme's `switch_to_remote_user_safe()` calls `switch_to_remote_user()` for non-admins, so `xhr.cgi` with `xhr-get_size` (also `xhr-get_list`, `xhr-encoding_convert`, `xhr-get_autocompletes`, `xhr-search-in-file`) returns a **"The Unix user anyusername does not exist"** error page. In the File Manager it shows up in the size column after creating or editing a file. The same user also loses everything else gated on `webmin_user_is_admin()`: e.g. live dashboard stats and some other features.

**Reproduce:**
1. Webmin Users → create `anyusername` (no Unix account of that name) as mentioned in **Environment** section above
2. Leave RPC at default *No*
3. Log in → File Manager → create or edit a file → size cell shows the error page.

**Workaround:** set RPC to *Yes*, or put `rpc=2` in `/etc/webmin/<user>.acl` back by hand (the "Only for root or admin" radio option isn't offered in Webmin UI if not set manually in `<user>.acl` file).

**Suggested fix:** stop using `rpc == 0` as a restricted-user signal in `webmin_user_login_mode()` now that it's the default for admins (let `_safe` drive it). That restores all affected features at once.

iliaross:
Aloha Jamie,

This PR stops treating `rpc=0` as a safe-user restriction, restoring File Manager ops and admin features for fully privileged accounts created with RPC disabled, while preserving explicit `_safe` restrictions intact.

Regression tests and Perl syntax checks pass.

Fixes https://github.com/webmin/webmin/issues/2837

### DIFF
--- a/filemin/filemin-lib.pl
+++ b/filemin/filemin-lib.pl
@@ -71,24 +71,9 @@ return get_selinux_command_type() ?
 sub can_write
 {
 my ($file) = @_;
-# No restrictions for root
-if (&webmin_user_is_admin()) {
-	return 1;
-	}
-# If strict check is enabled or if safe user check for write
-# access explicitly
+# Enforce strict and safe-user ACLs regardless of the Webmin login role
 if ($access{'work_as_user_strict'} || $access{'_safe'}) {
-	# Check if the file is a symbolic link
-	if (-l $file) {
-		# Resolve symbolic link
-		my $resolved_file = readlink($file);
-		# If the link is broken, allow writing
-		# to the link itself
-		return -w $file if (!$resolved_file);
-		# Otherwise, check the resolved file
-		$file = $resolved_file;
-		}
-	# Check if the file itself is writable
+	# The file test follows symlinks relative to the link's directory
 	return -w $file;
 	}
 # Otherwise, allow writing depending on Unix permissions
--- a/t/web-lib-funcs-login-mode.t
+++ b/t/web-lib-funcs-login-mode.t
@@ -1,0 +1,92 @@
+#!/usr/bin/perl
+# Tests for login privileges and RPC authorization in web-lib-funcs.pl.
+
+use strict;
+use warnings;
+use Test::More;
+use File::Basename qw(dirname);
+use File::Spec;
+
+my $script = File::Spec->rel2abs(
+	File::Spec->catfile(dirname(__FILE__), '..', 'web-lib-funcs.pl'));
+require $script;
+
+our ($remote_user, $base_remote_user);
+my (%test_acls, %test_modules);
+my $test_product = 'webmin';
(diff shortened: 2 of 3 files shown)


jcameron:
The code looks reasonable, but what is `webmin_user_login_mode` being used for? In general shouldn't we be using more fine-grained module ACLs than a single user mode?

iliaross:
> The code looks reasonable, but what is `webmin_user_login_mode` being used for? In general shouldn't we be using more fine-grained module ACLs than a single user mode?

Authentic uses it to choose what users see (ie nav menu type), or what settings they can change or which Unix user runs their requests.

Yet, the concern now as I see it is that this broad role may override module permissions, especially for file access and Unix-user switching...

...I'll need to take a more close look into it again.

jcameron:
> Yet, the concern now as I see it is that this broad role may override module permissions, especially for file access and Unix-user switching...

Ok let me know when this is fixed , and/or this PR is ready for review.

iliaross:
I'm working on theme changes. For Webmin side I think it's good to review.

iliaross:
Alright, all theme patches are checked-in now.

jcameron:
Ok this looks fine for Webmin, merging it now. However, I would strongly recommend against using this function for anything other than displaying what kind of user is logged in - it should never be used for ACL checks on user actions.

iliaross:
> However, I would strongly recommend against using this function for anything other than displaying what kind of user is logged in - it should never be used for ACL checks on user actions.

Alright, I have reviewed it once more and fixed remaining issues in this patch https://github.com/webmin/authentic-theme/commit/d055e361ab09b50069fd82aaee45897011f3f39d — thanks for the heads-up, Jamie!

jcameron:
Excellent thanks!

PR #2840, opened 2026-09-12
Add WebminCron pause during module post-install

iliaross:
Hello Jamie,

This PR stops Miniserv from starting scheduled jobs while `run-postinstalls.pl` is running, so jobs never execute against half-updated module code or config.

Miniserv defers all jobs, including startup jobs, while any marker belongs to a live process, and discards markers left by runners that died or that are older than an hour.

The pause is best-effort, as a failure to create the marker is reported but never blocks post-installs.

By the way, this issue was not reported by anyone. I discovered it while running many other extensive tests.

### DIFF
--- a/miniserv-lib.pl
+++ b/miniserv-lib.pl
@@ -10,6 +10,10 @@ eval "use Time::HiRes;";
 @itoa64 = split(//, "./0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz");
 @miniserv_argv = @ARGV;
 
+# Limit scheduler pauses so an interrupted update or reused PID cannot block
+# WebminCron jobs indefinitely.
+$webmincron_pause_max_age = 60*60;
+
 # init days and months for http_date
 @weekday = ( "Sun", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat" );
 @month = ( "Jan", "Feb", "Mar", "Apr", "May", "Jun",
@@ -3975,6 +3979,7 @@ if (!$config{'webmincron_wrapper'}) {
 	$config{'webmincron_wrapper'} = $config{'root'}.
 					"/webmincron/webmincron.pl";
 	}
+$config{'webmincron_pause'} ||= $var_dir."/webmincron-pause";
 if (!$config{'twofactor_wrapper'}) {
 	$config{'twofactor_wrapper'} = $config{'root'}."/acl/twofactor.pl";
 	}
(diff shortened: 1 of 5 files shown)


jcameron:
I'm not sure I like this, it seems like a work-around for a deeper problem if we cannot run cron jobs during the install. What was going wrong exactly during upgrades? Can we instead fix it with locking?

iliaross:
> I'm not sure I like this, it seems like a work-around for a deeper problem if we cannot run cron jobs during the install. What was going wrong exactly during upgrades? Can we instead fix it with locking?

I did observe the failure though, during an RPM update, the scheduled Virtualmin sysinfo collection was running WHOIS checks for hundreds of domains without DNS records. As each lookup can take up to 15 seconds, while `collect-lib.pl` holds that domain's lock.

Perhaps, a better and a more concrete fix, is to run the network lookups outside the domain lock?

But also pausing jobs during `run-postinstalls.pl` is not a bad idea in my view.

jcameron:
I'm not even sure that pausing would work in this case, because what if the collection script was already started?

Instead we should make sure that collection doesn't run for a long time .. for example in this whois cache, we could add a negative cache of the failed lookups so they aren't run every time. Or better still, not do `whois` lookups on domains that are not first-level under known TLDs.

iliaross:
> I'm not even sure that pausing would work in this case, because what if the collection script was already started?

Yeah, if it's already running, then it all gets to the running process to do things right.

Yet not letting a potentially heavy cron job start during a package update should also be a good thing, right?
 
> Instead we should make sure that collection doesn't run for a long time .. for example in this whois cache, we could add a negative cache of the failed lookups so they aren't run every time. Or better still, not do `whois` lookups on domains that are not first-level under known TLDs.

Sure we could, and I think you [already did](https://github.com/virtualmin/virtualmin-gpl/commit/b3050f748960566eb91d059617912c2956b5b808) though it's a different type of problem.

jcameron:
> Yet not letting a potentially heavy cron job start during a package update should also be a good thing, right?

Why though? Unless there's a clear reason to do this, we shouldn't be adding complexity by stopping and starting cron jobs. Instead let's make sure that locking is done correctly so that there is no chance of data corruption, and that scheduled collection doesn't run for too long.

For example, collectinfo will only do a limited number of LE renewals per run for this reason.

iliaross:
> Why though? Unless there's a clear reason to do this, we shouldn't be adding complexity by stopping and starting cron jobs. Instead let's make sure that locking is done correctly so that there is no chance of data corruption, and that scheduled collection doesn't run for too long.

Alright, that's fair. Closing now.

PR #2841, opened 2026-09-13
Fix IPv4 delays caused by IPv6 lookups

iliaross:
Hello Jamie,

This PR fixes delays when IPv6 DNS lookups stall otherwise working IPv4 connections.

The earlier changes ([04c74877](https://github.com/webmin/webmin/commit/04c74877c39c33283052ee2a8e7ccfe09448fd4e) and [8a6f7784](https://github.com/webmin/webmin/commit/8a6f778410aee972f87736d408175b4628311a22)) added fallback across available IPv4 and IPv6 addresses, which this PR preserves.

Resolving both families upfront appears to have been an implementation choice, yet now IPv6 is resolved only when no IPv4 connection succeeds.

Unit tests and real socket compatibility tests passed on Alma 10 and Debian 13.


jcameron:
Nice patch, and the code looks reasonable .. however I am curious as to what kind of DNS failure could cause an IPv4 address lookup to succeeded by a v6 lookup to time out?!

iliaross:
> Nice patch, and the code looks reasonable .. however I am curious as to what kind of DNS failure could cause an IPv4 address lookup to succeeded by a v6 lookup to time out?!

I reproduced it on Alma 10. The IPv4 lookup finished in about 50 ms. The IPv6 lookup also got a quick answer, but no IPv6 address.

The system resolver then tried the hostname with the VM’s DNS search suffix added:

```text
v4.download.virtualmin.com.alma10-gpl.virtualmin.dev
```

That extra query timed out, causing the 10-second delay. Removing that suffix for just the test process brought the lookup down to about 45 ms.

And, previously, IPv4 had to wait for all of this before connecting. So my patch lets IPv4 connect immediately and still tries IPv6 if IPv4 fails.

jcameron:
Ok I see it's the suffixes that are the issue.

What's up with the failing test above?

iliaross:
Test should be fixed now.

PR #2842, opened 2026-09-14
Fix concurrent Nginx configuration updates

iliaross:
Hi Jamie!

This PR fixes concurrent Nginx updates using stale config data.

It reloads the config after locking and safely handles nested locks and included files.


nginx/nginx-lib.pl
```diff
@@ -534,21 +535,59 @@ sub flush_config_file_lines
...
+my $main = &resolve_links($config{'nginx_config'}) || $config{'nginx_config'};
+my $ok = eval {
+	my @files = ($main);
+	my %locked;
+	while (@files) {
+		foreach my $f (@files) {
+			if (&lock_file($f)) {
+				push(@lock_all_config_files_cache, $f);
+				}
+			elsif (!defined($main::locked_file_list{&translate_filename($f)})) {
```
jcameron: What is this checking for? It seems like a bad idea to use variables that are part of the internal implementation of lock_file

nginx/nginx-lib.pl
```diff
@@ -534,21 +535,59 @@ sub flush_config_file_lines
...
+			$parent ? &get_all_config_files($parent) : ());
+		}
+	1;
+	};
+my $err = $@;
+if (!$ok) {
+	# Do not leave partially acquired locks behind if parsing or locking fails.
+	&unlock_file($_) foreach reverse(@lock_all_config_files_cache);
+	@lock_all_config_files_cache = ();
+	die $err;
```
jcameron: Won't lock files be cleaned up by Webmin's END function?

nginx/nginx-lib.pl
```diff
@@ -534,21 +535,59 @@ sub flush_config_file_lines
...
+	while (@files) {
+		foreach my $f (@files) {
+			if (&lock_file($f)) {
+				push(@lock_all_config_files_cache, $f);
+				}
+			elsif (!defined($main::locked_file_list{&translate_filename($f)})) {
+				&error("Failed to lock Nginx config file $f");
+				}
+			$locked{$f} = 1;
+			&unflush_file_lines($f);
```
jcameron: Why is this needed when the code above doesn't call read_file_lines?

nginx/nginx-lib.pl
```diff
@@ -534,21 +535,59 @@ sub flush_config_file_lines
 }
 
 # lock_all_config_files([&parent])
-# Locks all files used in the current config
+# Locks all config files and refreshes the config on the outermost call.
+# Fetch directive objects after locking; earlier objects may have stale lines.
 sub lock_all_config_files
```
jcameron, line 540: Since this can flush the cache, be extra careful that it is never called in a case where the caller has a copy of the config

nginx/nginx-lib.pl
iliaross: That was checking whether `lock_file` returned `0`. Yet I agree that using its internal implementation is wrong. Fixed https://github.com/webmin/webmin/commit/0c7fc411aa818c9b52c029df8f29d3a7fa125bcf.

nginx/nginx-lib.pl
iliaross: Yes, `END` removes lock files when the process exits. But the process may not exit here, and be caugt by a higher-level caller.

nginx/nginx-lib.pl
iliaross: Because, `read_file_lines` may have been called earlier in the same process.

nginx/nginx-lib.pl
iliaross: Yeah, I know and agree, and all callers already lock first, then fetch the config.

nginx/nginx-lib.pl
jcameron: I don't get that patch ... if the file is already locked by the current process, then lock_file will immediately succeed. And lock_file has an internal reference count, so if the file is already locked and you lock again, then unlock_file needs to be called twice to fully release the lock.

nginx/nginx-lib.pl
jcameron: This seems like a pretty rare case that we don't handle anywhere else in Virtualmin that takes locks. Unless there is a clear existing case where this function is called inside an eval, you could simplify it by removing this whole eval/die block.

nginx/nginx-lib.pl
jcameron: And so? What is the purpose of flushing it here? Can you give an example of where this happens?

nginx/nginx-lib.pl
iliaross: Alright, this one was overly defensive. Dropped now https://github.com/webmin/webmin/pull/2842/commits/9766789a94521bab7a4a810027594a3a0c4f4927.

nginx/nginx-lib.pl
iliaross: Flushing? We are unflushing, right? Only to clear the cached lines. For example, if this process read `nginx.conf`, another process changed it, and then this process acquired the lock, the old cached copy must be cleared before editing. Otherwise, it could (and would) overwrite the newer changes.

jcameron: Oh I see ... in that case, maybe this unflushing should go into the existing function that clears the Nginx cache?

nginx/nginx-lib.pl
jcameron: Thanks!

nginx/nginx-lib.pl
iliaross: > Oh I see ... in that case, maybe this unflushing should go into the existing function that clears the Nginx cache?

No, as `unflush_file_lines()` can drop edits that have not been saved yet. I think it should stay here, as all the tests I have are already done.

nginx/nginx-lib.pl
```diff
@@ -534,21 +535,53 @@ sub flush_config_file_lines
...
+		else {
+			&error("Failed to lock Nginx config file $f");
+			}
+		$locked{$f} = 1;
+		&unflush_file_lines($f);
+		}
+	# An included file may also be edited directly. Re-read after
+	# waiting for its lock and pick up any newly included files.
+	&flush_config_cache();
+	@files = grep { !$locked{$_} } &unique(&get_all_config_files(),
```
jcameron: I don't get what is happening here - why would get_all_config_files change inside this loop? Wouldn't it make more sense to populate @files once and then just iterate over and lock them?

nginx/nginx-lib.pl
```diff
@@ -534,21 +535,53 @@ sub flush_config_file_lines
...
+		elsif (&lock_file($f)) {
+			push(@lock_all_config_files_cache, $f);
+			}
+		else {
+			&error("Failed to lock Nginx config file $f");
+			}
+		$locked{$f} = 1;
+		&unflush_file_lines($f);
+		}
+	# An included file may also be edited directly. Re-read after
```
jcameron: What does "edited directly" mean?

nginx/nginx-lib.pl
iliaross: Jamie, `get_all_config_files()` follows Nginx `include` directives and globs, so its result can change while we wait for locks.

For instance, `nginx.conf` may include `sites-enabled/*`. While this process waits for the main config lock, another process can create a domain and add a new file under `sites-enabled`. After acquiring the lock, rereading the config discovers that new domain file and adds it to the next batch of files to lock.

Populating `@files` only once before locking would miss the new domain file.

So, the loop continues until every file found by the latest parse is locked.

nginx/nginx-lib.pl
iliaross: By "edited directly," I mean changed through Webmin's raw Nginx configuration editor. That editor locks only the selected included file, not the entire Nginx configuration.

If it adds another `include` while this process waits for the file's lock, `get_all_config_files()` will return a different list after the lock is acquired!!

nginx/nginx-lib.pl
jcameron: Ok I see. But this whole loop is still very hard to understand .. It would be a lot clearer if the logic was like : 

```
while(1) {
   # Flush config cache

   # Get list of files using get_all_config_files

   # Try to lock them all, and keep track of which new locks were created

   # If all locks were already owned by this process, exit the loop
   }
```



Also in the case where `$parent` is given presumably there is only a single file, so this whole loop can be skipped and replaced with a single lock_file call.

iliaross: > Ok I see. But this whole loop is still very hard to understand .. It would be a lot clearer if the logic was like :

Alright, I rewrote it to make it easier to read and reran all tests in full to ensure it works correctly.

> Also in the case where $parent is given presumably there is only a single file, so this whole loop can be skipped and replaced with a single lock_file call.

Yet again, I kept the normal discovery path for `$parent` because a parent can contain `include` directives.

This time though, I added a regression test covering that case.

nginx/nginx-lib.pl
```diff
@@ -534,21 +535,59 @@ sub flush_config_file_lines
...
+if ($lock_all_config_files_depth) {
+	# Nested edits share the caller's config tree and pending changes.
+	$lock_all_config_files_depth++;
+	return;
+	}
+
+# Lock the main file before parsing, so another writer cannot change the
+# include list or directive line numbers while we acquire the remaining locks.
+my $main = &resolve_links($config{'nginx_config'}) || $config{'nginx_config'};
+my @files = ($main);
```
jcameron, line 551: Is $main the right file to use here if $parent is given?

nginx/nginx-lib.pl
jcameron: Does this module use read_file_lines_cached ? If not, there's not much value in calling unflush_file_lines to clear that cache, and it's potentially dangerous to other code which is reading these configs directly. Not that this happens in practice anywhere as far as I know. So I would recommend just drop this unless there is a clearly demonstrated case where it is needed.

nginx/nginx-lib.pl
jcameron: Thanks!

nginx/nginx-lib.pl
iliaross: Yes, why not? `$main` is the main Nginx config file, so we lock it first because it controls which other files are included.

And, `$parent` may reference additional included files, which are locked latet.

nginx/nginx-lib.pl
iliaross: > And it's potentially dangerous to other code which is reading these configs directly. Not that this happens in practice anywhere as far as I know.

I understand what you mean now. We shouldn't completely drop it though! See this change https://github.com/webmin/webmin/pull/2842/commits/a8a1b700150f60539f2c9e65992ef5e98e86f7ae. All good now?

nginx/nginx-lib.pl
iliaross: Jamie, sorry I totally missed this comment.

> And lock_file has an internal reference count, so if the file is already locked and you lock again, then unlock_file needs to be called twice to fully release the lock.

Um, surely it does _not_! Before this PR, there was only one place where you had lock tracking, which is `lock_all_config_files` in fail2ban module.

jcameron: Sorry I described that wrong, it just tracks if the file is already locked, so the second call on the same file does nothing. Anyway I think this is moot now..

nginx/nginx-lib.pl
jcameron: I don't thing that will actually do much in practice, but I guess its OK

iliaross: Yeah, it's fine to me!

nginx/nginx-lib.pl
jcameron: Is lock_all_config_files ever even called with a parent object? I don't see it anywhere in the nginx or virtualmin-nginx modules. Maybe we can just drop this use case?

iliaross: Alright, after checking again, I agree that `$parent` adds nothing useful and not reduces lock scope. Dropped in this https://github.com/webmin/webmin/pull/2842/commits/c134426e2ae317f70cfaaacbde1bbc9ba9b626ed patch.

iliaross:
I think this is good to be merged?

nginx/nginx-lib.pl
jcameron: Nice thanks!

PR #2844, opened 2026-09-15
bind8: allow the DMARC pct field to be left empty

### LINKED ISSUE #2843: bind8: DMARC form forces the pct= tag, which RFC 9989 removed
**Affected versions:** 2.000 up to and including 2.660 (the relevant code is unchanged between them).

**Behaviour:** In Servers → BIND DNS Server → zone → DMARC, the field *Percentage of messages to apply policy* is mandatory. Leaving it empty fails validation, so every DMARC record written through Webmin contains a `pct=` tag. There is no way to produce a record without it.

**Why this matters:** RFC 9989 (DMARCbis, May 2026), which obsoletes RFC 7489, removes the `pct` tag from the specification. Records generated by Webmin therefore always contain a tag that the current standard no longer defines, and DMARC analysers flag it as deprecated.

**Where it happens:** `bind8/save_record.cgi`, around line 397 on master:

```perl
$in{'dmarcpct'} =~ /^\d+$/ && $in{'dmarcpct'} >= 0 &&
  $in{'dmarcpct'} <= 100 || &error($text{'edit_edmarcpct'});
$dmarc->{'pct'} = $in{'dmarcpct'};
```

The value is validated as mandatory and then always assigned. Compare this with the `sp` field immediately below, which deletes the tag when the input is empty:

```perl
if ($in{'dmarcsp'}) {
	$dmarc->{'sp'} = $in{'dmarcsp'};
	}
else {
	delete($dmarc->{'sp'});
	}
```

`join_dmarc()` in `bind8/records-lib.pl` already skips tags with an empty value, so no change is needed there.

**Suggested fix:** treat `pct` the same way as `sp`. Accept an empty value and delete the tag when it is empty, while still accepting 0-100 for anyone who deliberately uses it during a rollout.

Deltaman-MWI:
Fixes #2843.

RFC 9989 (DMARCbis, May 2026) removes the `pct` tag, but `bind8/save_record.cgi` validated the percentage as mandatory and then always assigned it, so every DMARC record written through Webmin contained `pct=`. There was no way to produce a record without it.

This treats the field the same way as the `sp` field directly below it: when the input is empty, the tag is deleted. Values 0-100 are still accepted and validated for anyone who deliberately uses `pct` during a rollout. `join_dmarc()` in `bind8/records-lib.pl` already skips tags with an empty value and `ui_textbox` renders an empty box when the tag is absent, so no other changes were needed.

Behaviour of the changed block:

| input | result |
|---|---|
| `100` | `pct=100` (unchanged) |
| `50` | `pct=50` (unchanged) |
| empty | tag omitted (new) |
| `abc` | validation error (unchanged) |
| `150` | validation error (unchanged) |

`perl -c bind8/save_record.cgi` passes.


jcameron:
Thanks for the fix!

PR #2847, opened 2026-09-16
Add Apache comment preservation

iliaross:
Hello Jamie,

This PR fixes full-line comments being removed when Webmin rewrites Apache configuration blocks.


jcameron:
I don't really like this approach - those 'dummy' directives are just used as a marker for the start of blocks that might not have any real directives in them. Using them for comments is the wrong approach.

If we really do want to preserve comments, it should be done by attaching them to the following real directive as a new field.

iliaross:
> If we really do want to preserve comments, it should be done by attaching them to the following real directive as a new field.

Understood and agreed! Please see the latest changes.

jcameron:
Nice ... but please run the full Virtualmin functional-test suite (at least the parts that touch the apache config) with this PR to make sure it doesn't cause unexpected corruption.

I like this change, but maybe let's not include it in the upcoming release without very thorough testing, as Apache config parsing is quite complex.

iliaross:
> but please run the full Virtualmin functional-test suite (at least the parts that touch the apache config) with this PR to make sure it doesn't cause unexpected corruption.

I have run multiple _live_ tests on local virtual machines that assumed callers in Virtualmin. But I can add tests to `functional-test.pl` too.

 > I like this change, but maybe let's not include it in the upcoming release without very thorough testing, as Apache config parsing is quite complex.
 
Yeah, I don't mind delaying it. It can wait, and it's not important enough to be a blocker.

jcameron:
Yeah, a functional test that adds a directive with a comment and removes it would be handy.

Also, running all the other tests would ensure that this hasn't somehow introduced a bug.

iliaross:
> Yeah, a functional test that adds a directive with a comment and removes it would be handy.

Done https://github.com/virtualmin/virtualmin-gpl/commit/4daa012ca915b00c8eba78b58e6f275c3ff61943.

jcameron:
> Done https://github.com/virtualmin/virtualmin-gpl/commit/4daa012ca915b00c8eba78b58e6f275c3ff61943.

Ah no .. please see my comment on that commit.

iliaross:
> Ah no .. please see my comment on that commit.

Yes, I did see it and will fix it.

iliaross:
> > Done [virtualmin/virtualmin-gpl@4daa012](https://github.com/virtualmin/virtualmin-gpl/commit/4daa012ca915b00c8eba78b58e6f275c3ff61943).
> 
> Ah no .. please see my comment on that commit.

Fixed in https://github.com/virtualmin/virtualmin-gpl/pull/1290 and https://github.com/virtualmin/virtualmin-gpl/pull/1291.

iliaross:
Should be good too?

jcameron:
Yes sorry I missed this one!

iliaross:
> Yes sorry I missed this one!

Thanks!

